Issue #7, opened 2019-08-06
Old Kubernetes labels are used


rimusz:
It should use https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

Latest versions of `helm` cli   generates the correct ones as well: `helm create some_chart`

jasonodonnell:
Will fix, nice catch @rimusz !

jasonodonnell:
Fixed via #8, closing.  Thanks!

PR #8, opened 2019-08-06
Update helm labels

jasonodonnell:
Addresses #7 by updating the label metadata.

### DIFF
--- a/templates/server-config-configmap.yaml
+++ b/templates/server-config-configmap.yaml
@@ -6,10 +6,11 @@ kind: ConfigMap
 metadata:
   name: {{ template "vault.fullname" . }}-config
   labels:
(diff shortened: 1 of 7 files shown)


jasonodonnell:
> Can we verify these labels work for our earliest supported version of Kube? Or README says that's ~1.9 but mostly 1.10/1.11. Otherwise, maybe we should consider setting a fixed min-version supported

@catsby These labels are just metadata and really have no impact to runtime - they're used for filtering via selectors.  I went ahead and provisioned a 1.11 cluster (oldest I could get on GKE) and they work as expected.

I agree that we might need to increase this to 1.10/11 for integration testing coverage.  Bootstrapping those old versions only gets harder over time.

PR #10, opened 2019-08-07
Add clusterrolebinding, fix service, update Vault

jasonodonnell:
- Added cluster role binding to make bootstrapping the Kube auth method easier
- Added configurable for services to enable/disable headless services.  Headless services make sense when Consul is being used but not when using file storage.
- Upgraded to Vault 1.2.1

### DIFF
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -1,8 +1,4 @@
-# Headless service for Vault server DNS entries. This service should only
-# point to Vault servers. For access to an agent, one should assume that
-# the agent is installed locally on the node and the NODE_IP should be used.
-# If the node can't run a Vault agent, then this service can be used to
-# communicate directly to a server agent.
+# Service for Vault cluster
 {{- if and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
 apiVersion: v1
 kind: Service
@@ -20,7 +16,7 @@ metadata:
     # https://github.com/kubernetes/kubernetes/issues/58662
     service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
 spec:
-  clusterIP: None
+  clusterIP: {{ .Values.server.service.clusterIP }}
   # We want the servers to become available even if they're not ready
   # since this DNS is also used for join operations.
   publishNotReadyAddresses: true
@@ -32,7 +28,7 @@ spec:
       port: 8201
       targetPort: 8201
   selector:
-    app: {{ template "vault.name" . }}
-    release: "{{ .Release.Name }}"
+    app.kubernetes.io/name: {{ include "vault.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
     component: server
 {{- end }}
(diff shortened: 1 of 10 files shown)


values.yaml
```diff
@@ -21,6 +21,12 @@ server:
   #     memory: 256Mi
   #     cpu: 250m
 
+  # authDelegator enables a cluster role binding to be attached to the service 
+  # account.  This cluster role binding can be used to setup Kubernetes auth 
+  # method.  https://www.vaultproject.io/docs/auth/kubernetes.html
+  authDelegator:
+    enabled: true
```
malnick: Are we sure we want to default this to true? I think this should be an opt-in feature.

jasonodonnell: Agreed.  Changed in https://github.com/hashicorp/vault-helm/pull/10/commits/b82db908dd2bebbbcb158824e05ea3ca1e3b7ef4

values.yaml
```diff
@@ -69,6 +75,9 @@ server:
   # Enables a headless service to be used by the Vault Statefulset
   service:
     enabled: true
+    # clusterIP can be used to disable a cluster IP address from being 
+    # attached to the Vault service if set to "None".
```
catsby: Can we re-word this to be clear that the default is to in fact have a cluster IP address attached? The default scenario has to be derived from this sentence, I would prefer something like "by default vault has a cluster ip , set to `None` to disable" and maybe a small bit on why one would disable that

jasonodonnell: Done in https://github.com/hashicorp/vault-helm/pull/10/commits/e8460cd120ae0c5017adfcc80cc4aeb689e99946, thanks!

jasonodonnell:
Indeed, follow up PR for the website coming.

templates/server-service.yaml
```diff
@@ -20,7 +16,7 @@ metadata:
     # https://github.com/kubernetes/kubernetes/issues/58662
     service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
 spec:
-  clusterIP: None
+  clusterIP: {{ .Values.server.service.clusterIP }}
```
kuuji: This doesn't work if clusterIP is not set in values.yaml (and it's not by default)

I think you need an if statement before that to only include clusterIP if clusterIP is defined.

kuuji: fyi, In my case I don't want to specify a ClusterIP and I don't want it set to None.

jasonodonnell: Interesting that you're receiving an error because it works fine on GKE when the value is empty.

Which version of Kube are you using?

jasonodonnell: Ah, I see the error doesn't appear until you try to upgrade from a service from an earlier version.

Thanks for pointing this out, will fix this @kuuji

kuuji: Actually, after double checking, I'm only getting it on upgrade, because I guess you can't remove that field once you set it.

kuuji: I have a branch with a fix if you want @jasonodonnell . I can open a PR

PR #11, opened 2019-08-08
Add AWS credential environment variable

anhdat:
This PR allows user to add AWS credential via environment variable with a Kubernetes Secret object:

```yaml
# values
server:
  aws:
    secretName: vault
    secretKey: AWS_SECRET_ACCESS_KEY
  extraEnvironmentVars:
    AWS_REGION: eu-west-1
    AWS_ACCESS_KEY_ID: xxx
```

Secret object:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: vault
type: Opaque
stringData:
  AWS_SECRET_ACCESS_KEY: xxx
```

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -143,6 +143,13 @@ Set's additional environment variables based on the mode.
             - name: VAULT_DEV_ROOT_TOKEN_ID
               value: "root"
   {{ end }}
+  {{ if .Values.server.aws.secretName }}
+            - name: AWS_SECRET_ACCESS_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.server.aws.secretName }}
+                  key: {{ .Values.server.aws.secretKey }}
+  {{ end }}
 {{- end -}}
 
 {{/*
--- a/test/unit/server-ha-statefulset.bats
+++ b/test/unit/server-ha-statefulset.bats
@@ -258,6 +258,28 @@ load _helpers
 }
 
 #--------------------------------------------------------------------
+# aws
+
+@test "server/ha-StatefulSet: set aws" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/server-statefulset.yaml  \
+      --set 'server.aws.secretName=foo-name' \
+      --set 'server.aws.secretKey=bar-key' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].env' | tee /dev/stderr)
+
+
+  local actual=$(echo $object |
+      yq -r '.[4].valueFrom.secretKeyRef.name' | tee /dev/stderr)
+  [ "${actual}" = "foo-name" ]
+
+  local actual=$(echo $object |
+      yq -r '.[4].valueFrom.secretKeyRef.key' | tee /dev/stderr)
+  [ "${actual}" = "bar-key" ]
+}
+
+#--------------------------------------------------------------------
 # storage class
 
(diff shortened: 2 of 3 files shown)


jasonodonnell:
Hi @anhdat, thanks for the contribution!

This is already possible using the Chart as is (but in a more generic way that isn't specific to AWS):

```yaml
server:
   extraVolumes:
     - type: secret
       name:  vault
   extraEnvironmentVars:
     AWS_REGION: eu-west-1
     AWS_ACCESS_KEY_ID: xxx
```

Can you provide some additional context on why we should add an AWS specific option?

anhdat:
hi @jasonodonnell, thanks for fast response. 

Not sure how `extraVolumes` will work since it would be mounted as a file at `mountPath: /vault/userconfig/{{ .name }}` while we want an environment variable or able to mount a file at `~/.aws/credentials`. 

I don't know what should be the standard way to do this in HashiCorp's charts, in many charts we usually allow users to freely pass the env vars configuration (e.g. https://github.com/helm/charts/blob/master/stable/cert-manager/templates/deployment.yaml#L58). I guess that HashiCorp charts want to restrict which mount point should be used and limit on names in env var. So I followed the approach in `consul-helm` https://github.com/hashicorp/consul-helm/blob/master/templates/enterprise-license.yaml#L40

Let me know if I misunderstood this. Thanks.

anhdat:
hi @jasonodonnell do we have any update on this? Thanks.

jasonodonnell:
@anhdat,

I'd still like this to be not AWS specific.  I think this could be done with the following:

Update helper (add `mountPath`)
```
{{/*
Set's which additional volumes should be mounted to the container
based on the mode configured.
*/}}
{{- define "vault.mounts" -}}
  {{ if eq .mode "standalone" }}
    {{ if eq (.Values.server.auditStorage.enabled | toString) "true" }}
            - name: audit
              mountPath: /vault/audit
    {{ end }}
    {{ if eq (.Values.server.dataStorage.enabled | toString) "true" }}
            - name: data
              mountPath: /vault/data
    {{ end }}
  {{ end }}
  {{ if and (ne .mode "dev") (or (ne .Values.server.standalone.config "")  (ne .Values.server.ha.config "")) }}
            - name: config
              mountPath: /vault/config
  {{ end }}
  {{- range .Values.server.extraVolumes }}
            - name: userconfig-{{ .name }}
              readOnly: true
              mountPath: {{ .path | default "/vault/userconfig" }}/{{ .name }}
  {{- end }}
{{- end -}}
```

New helper: _extraSecretEnvironmentVars_
```
{{/*
Inject extra environment populated by secrets, if populated
*/}}
{{- define "vault.extraSecretEnvironmentVars" -}}
{{- if .extraSecretEnvironmentVars -}}
{{- range .extraSecretEnvironmentVars }}
- name: {{ .envName }}
  valueFrom:
   secretKeyRef:
     name: {{ .secretName }}
     key: {{ .secretKey }}
{{- end -}}
{{- end -}}
{{- end -}}
```

Thoughts?

anhdat:
I like the generic approach more. :+1: 
Your comment is almost a complete PR, we still need tests for that.
Will you open a new PR yourself or should I update this PR?

jasonodonnell:
I can add this in a new PR later this week, however, if you want this sooner (and the contribution) you can update the PR. :)

anhdat:
I'm using this chart for my company deployment, managing many branches at the same time is not so nice, so I will submit another PR today. Thanks for the guidance.

anhdat:
Close in favor of generic approach in #16

PR #12, opened 2019-08-08
Only include clusterIp on vault service if set in values

kuuji:
Prevents error if the field is set to empty and then remove on an upgrade.

reference : https://github.com/hashicorp/vault-helm/pull/10

### DIFF
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -16,7 +16,9 @@ metadata:
     # https://github.com/kubernetes/kubernetes/issues/58662
     service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
 spec:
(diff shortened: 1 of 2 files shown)


kuuji:
I don't think that solves the issue actually @jasonodonnell. The field is immutable, the only way to change seems to be to recreate the resource.

jasonodonnell:
@kuuji Yeah, that's what I'm seeing as well.  It worked when I did:
`helm upgrade --force` but that recreates the objects.

kuuji:
heh, I was going to mention that as well. 

Do what you want with this PR then.
I still think there's value in having the if statement rather than the empty `ClusterIp` field, but it's your call ;).

PR #18, opened 2019-08-13
UI Service annotations

amoskyler:
Allows customization of UI service's annotations.

I followed a similar configuration to the other annotation template annotations.


amoskyler:
I totally botched this PR. Re-creating

PR #19, opened 2019-08-13
Support UI service annotations

amoskyler:
Allows customization of UI service's annotations.

I followed a similar configuration to the other annotation template annotations.


templates/ui-service.yaml
```diff
@@ -25,3 +28,5 @@ spec:
       targetPort: 8200
   type: {{ .Values.ui.serviceType | default "ClusterIP" }}
 {{- end -}}
+
+{{ end }}
```
anhdat: New line at the end of file 
```suggestion
{{ end }}

```

jasonodonnell:
Hi @amoskyler, thank you for the contribution!

I like this feature and can see its usefulness.  Would you mind adding unit tests for this to ensure its functioning properly?

amoskyler:
Great idea @jasonodonnell. I added some tests, and in the process found a bug!

PR #21, opened 2019-08-15
Add TLS support 

anhdat:
- Add TLS support for health check: with TLS enabled, health check also need to use `https` scheme
- Add custom NodePort: in the situation of using NodePort type for Service, user should be able to define selected NodePort

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -103,7 +103,7 @@ extra volumes the user may have specified (such as a secret with TLS).
          {{- if (eq .type "configMap") }}
            name: {{ .name }}
          {{- else if (eq .type "secret") }}
-          secretName: {{ .name }}
+           secretName: {{ .name }}
          {{- end }}
   {{- end }}
 {{- end -}}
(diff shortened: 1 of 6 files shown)


jasonodonnell:
Hey @anhdat, thanks for another contribution!

This was something that I was working on myself, appreciate you doing this.  This did cause a failure in the acceptance tests.  Since the default Vault config disables TLS (because there is no TLS bootstrapping as part of the helm install), I think we will need to make `tls_disable: true` by default.

jasonodonnell:
@anhdat #23 reduced the complexity of the UI service by using 8200.  The Vault server will handle either http/https connections depending on the configuration.

Given this change, I would suggest making just a single variable for specifying `nodePort`.

anhdat:
@jasonodonnell got it. I will update for both remarks :+1:

anhdat:
@jasonodonnell the acceptance tests are passing now, please proceed the review.

jasonodonnell:
@anhdat Thanks!  I have one minor request, can we turn `tls_disable` into `tlsDisable` just to be consistent with the rest of the options (we've been using camel case).

Besides that minor feedback this LGTM!

anhdat:
>can we turn tls_disable into tlsDisable just to be consistent with the rest of the options (we've been using camel case)?

Absolutely, I missed that. Updated now :ok_hand:

PR #24, opened 2019-08-20
Address K8s Deprecation of Kubelet security controls. Remove Whitespaces.

Xtigyro:
Two changes:

1. Address K8s Deprecation of Kubelet security controls - https://github.com/kubernetes/kubernetes/pull/77820
It's done here: https://github.com/hashicorp/vault-helm/pull/24/files#diff-60ca1594dfcfe4f0d22db67a9583d9fe

2. Remove all whitespaces.

PR #35, opened 2019-08-22
Add privileged for mlock requirements

jasonodonnell:
#24 caused an error in the Vault pod because `mlock` was failing due to reduced privileges.  Reverted for now until we can investigate other security capabilities.


jasonodonnell:
Done and done, thanks @christopherbloghett-disney !

PR #38, opened 2019-08-27
add namespace in vault manifest definition

smehboub:
Hello,

Thank you very much for your work. it's amazing

I submit this pull request to make a small addition if it's possible.

When I use the helm template command by specifying the namespace, the namespace isn't replace in some k8s manifest.
By adding the statement ( namespace: {{ .Release.Namespace }} ) like you did for the service account k8s manifest for example, it works well.

Thanks in advance.
Rgs,

Sophian Mehboub


smehboub:
Hello Jason,

Thank you very much for your prompt response and for your help.
Could you create a GitHub release, please ?

Thanks in advance.
Rgs,

Sophian Mehboub

PR #41, opened 2019-08-29
Simplify TLS Certs Implementation. Add Ingress.

Xtigyro:


jasonodonnell:
Hi @Xtigyro, thanks for the contribution!

I really like this feature, it adds functionality that I think is needed.  That being said I'm concerned with the container image that is being used here for the job.  This image is out of our control and we'd rather not have external dependencies.

I need to do some research and talk to colleagues about this feature.  Leaving this open for now :)

Thanks!

Xtigyro:
> 
> 
> Hi @Xtigyro, thanks for the contribution!
> 
> I really like this feature, it adds functionality that I think is needed. That being said I'm concerned with the container image that is being used here for the job. This image is out of our control and we'd rather not have external dependencies.
> 
> I need to do some research and talk to colleagues about this feature. Leaving this open for now :)
> 
> Thanks!

@jasonodonnell I'm really glad that we think the very same!

Can we create such image for HashiCorp - that HashiCorp controls it? It won't need any changes for very long time, for sure.

Xtigyro:
@jasonodonnell @kawsark 
I'm gonna add tomorrow `bats` tests for the Ingress and resolve the new stuff that was merged after my initial PR.

billimek:
The ingress seems to be hard coded to use nginx which may be too opinionated. 

Is it possible to define the ingress in a way that allows for any ingress controller?

Xtigyro:
@billimek Yes, it should be. I wanted nice and useful options to be possible with super easy config in `values.yaml`. Will dial it back a little bit to allow for other Ingress Controllers.

Thank you!

Xtigyro:
@billimek Done.

jasonodonnell:
@Xtigyro, thanks for the work on ingress!  It would be helpful to split this into a separate PR as we're not ready to move forward with this TLS work yet.

Xtigyro:
> 
> 
> @Xtigyro, thanks for the work on ingress! It would be helpful to split this into a separate PR as we're not ready to move forward with this TLS work yet.

@jasonodonnell @kawsark Okay - will try though the Ingress code builds on the work for the TLS but will try to separate it somehow later today. Thank you for being so supportive!

Xtigyro:
@jasonodonnell @kawsark 
Added `bats` Unit tests for the Ingress resources. The initially planned work has been completed.

I'll separate it into an independent PR if the implementation is considered advantageous.

Xtigyro:
@jasonodonnell @kawsark 
I would like to reflect on a certain deployment scenario.

If you must keep a very strict security rules applied to your environments and no Internet access can be allowed - you won't be able to utilize `Let's Encrypt` with `Cert Manager`. In that case you either need to rely on signing all your certificates with a manually created CA signing keys or you actually can utilize `Vault's PKI Engine` that can come to the rescue for you. If you use Vault's PKI Engine, you can actually rely on manual cert creation only for Vault's CA cert and all the other certificates you might need, will be generated by Cert Manager gunned with Vault's PKI Engine. So it can come handy to have an independent way for generating Vault's TLS certs.

Having said that - we can dial this PR down a bit and give the user two ways of enhancing the current situation: 

1. Either create only CA signing keys and save them in a K8s secret; 
2. Or provide them with an option to have all the TLS certs created and saved in a K8s secret. 

In that way if you choose the option (1) - you'll be able to provide the `Cert Manager CA Issuer` with the required CA keys to create and save for you a TLS key and a TLS cert signed by that CA pair that can be utilized by Vault later.

Xtigyro:
Too much genuine optimism was put into this PR. Thank you for the lessons learnt.

PR #43, opened 2019-08-30
Changed vault service port and targetPort to values file

arunbsar:


values.yaml
```diff
@@ -11,6 +11,11 @@ global:
   tlsDisable: true
 
 server:
+  # Port on which Vault server is listining
```
jasonodonnell: Typo in comment, should be `listening`.

jasonodonnell:
Hi @arunbsar, thanks for the contribution!

I see that these have been added to the `values.yaml` file, but nothing has changed in the templates.  How do these values get rendered into say the Vault service?

There will also need to be unit tests covering these configurables.

Thanks!

arunbsar:
Hi Jason,

Fixed the typo, and added missing changes and tests

PR #44, opened 2019-09-02
feat: allow setting loadBalancerIP and externalPort for service-ui

Miradorn:
One way to allow external access to the vault cluster is via a LoadBalancer service (another would be an ingress).
With this PR it is possible to set the external port of the ui service and if wanted to specify a pre-registered ip for the service.

### DIFF
--- a/templates/ui-service.yaml
+++ b/templates/ui-service.yaml
@@ -24,12 +24,15 @@ spec:
     component: server
   ports:
     - name: http
-      port: 8200
+      port: {{ .Values.ui.externalPort }}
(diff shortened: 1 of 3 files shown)


jasonodonnell:
Hi @Miradorn, thanks for the contribution!

I would like to add this feature but need test coverage on the new items.  Can you add unit tests (in the `test/unit` directory)?  Thanks!

Miradorn:
Hi,
I added a test and also removed one that was wrong (see commit message why the test was wrong/duplicate)

PR #48, opened 2019-09-08
Support for ingress

mogaal:
Hello, 

Within my actual implementation I have to expose the Vault UI. For that, I had to add ingress support. Sharing new feature here. 

Have a nice weekend

### DIFF
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -1,0 +1,41 @@
+{{- if .Values.server.ingress.enabled -}}
+{{- $serviceName := include "vault.fullname" . -}}
+{{- $servicePort := .Values.server.service.port -}}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{ template "vault.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    helm.sh/chart: {{ include "vault.chart" . }}
+    app.kubernetes.io/name: {{ include "vault.name" . }}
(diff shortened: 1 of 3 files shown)


jasonodonnell:
Be aware this came in @Xtigyro before you go splitting your PRs.

jasonodonnell:
Hi @mogaal, thanks for the contribution!

This is a great feature which we would like to add.  Can you add tests for this?  You can find all the tests in `test/unit`.

Xtigyro:
@jasonodonnell This implementation differs quite a bit in some ways. We should consider the pros and cons of the two.

mogaal:
@jasonodonnell Just added units tests. To be honest I followed the most common approach in almost all helm charts you find in [stable/*](https://github.com/helm/charts/tree/master/stable) for ingress set up. 

Having a look to @Xtigyro's implementation I personally would take away SSL management out of the helm chart. You already have custom charts implementations like [certmanager/lets-encrypt](https://github.com/jetstack/cert-manager) which are specifically for that and they work great.

billimek:
Agree with @mogaal on both points here. IMO the chart ingress should follow the pattern employed by other helm charts.

Xtigyro:
@mogaal @billimek The Ingress implementation is independent from the TLS stuff. Re the Ingress - I combined the general approach with some simplification if you want to use `NGINX` - which most people do.

catsby:
Thank you @mogaal for the contribution! And thank you everyone who participated in civil discussion here.

PR #51, opened 2019-09-11
Affinity should be always a available

mogaal:
Even when I am in dev mode I should be able to set affinity, it is actually my case: At work, my team is only available to deploy in certain nodes. I am playing with the cluster (no HA) in dev mode and I should be able to et up affinity rules to decide where I want my cluster

The same applies to annotations, node selectors, etc.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -212,7 +212,7 @@ storage might be desired by the user.
 Set's the affinity for pod placement when running in standalone and HA modes.
 */}}
 {{- define "vault.affinity" -}}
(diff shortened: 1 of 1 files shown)


malnick:
@mogaal this looks great! Can you rebase this on master to pick up the circleci integration, and also add those unit tests so we can merge this with confidence, thanks again!

jasonodonnell:
Bumping for unit tests.  If you're unavailable to add tests we can help!

jasonodonnell:
One last bump before closing due to inactivity.

PR #52, opened 2019-09-11
Service annotations support

mogaal:
Added support to define custom annotations inside the service definitions.


malnick:
@mogaal This was accidentally prematurely merged, can you open up a follow on PR with the relevant test? Thanks again for the contribution!

PR #58, opened 2019-09-23
WIP: Raft support

jasonodonnell:
This adds tentative support for Raft storage in Vault Helm.  This required allowing HA mode to create PVCs during installation for persistence needs (we were relying on external HA storage before, so creating PVCs as disabled in HA mode).

Additionally there was a problem with the StatefulSet.  Long story short, the StatefulSet needs to know how to lookup the other Vault pods in the set.  To do this we require a headless service to be used for DNS purposes (regular services do not suffice here because they round robin load balance).  This means we will now have two services (configurable): one for our internal DNS uses (raft) and the other for ingress/load balance/cluster traffic.

### DIFF
--- a/test/acceptance/_helpers.bash
+++ b/test/acceptance/_helpers.bash
@@ -117,7 +117,7 @@ wait_for_ready() {
     for i in $(seq 60); do
         if [ -n "$(check ${POD_NAME})" ]; then
             echo "${POD_NAME} is ready."
-            sleep 2
+            sleep 10
             return
         fi
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -181,7 +181,7 @@ storage might be desired by the user.
 {{- define "vault.volumeclaims" -}}
   {{- if and (ne .mode "dev") (or .Values.server.dataStorage.enabled .Values.server.auditStorage.enabled) }}
   volumeClaimTemplates:
-      {{- if and (eq (.Values.server.dataStorage.enabled | toString) "true") (or (eq .mode "standalone") (and ((eq .mode "ha") (eq (.Values.server.ha.raft.enabled | toString) "true")))) }}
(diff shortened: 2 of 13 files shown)


test/acceptance/_helpers.bash
```diff
@@ -117,7 +117,7 @@ wait_for_ready() {
     for i in $(seq 60); do
         if [ -n "$(check ${POD_NAME})" ]; then
             echo "${POD_NAME} is ready."
-            sleep 2
+            sleep 10
```
malnick: This could make the tests really slow as our test cases grow in number and complexity. Wondering if there's a better way to manage this other than `sleep`?

jasonodonnell: I can be smarter here with `kubectl wait`.

billimek: A possible implementation could be something like,

```shell
  VAULT_READY=1
  while [ $VAULT_READY != 0 ]; do
    kubectl -n <some namespace> wait --for condition=available statefulset/vault > /dev/null 2>&1
    VAULT_READY="$?"
  done
```

jasonodonnell: That's what I was thinking @billimek! 👍

hollow442020: Very good

msidd111:
Any Idea when raft will be available in Vault GA release?

msidd111:
Any Idea when raft will be available in Vault GA release?

jasonodonnell:
Will reopen when this is ready to be released.  Don't have the capacity to work on this at the moment.

hollow442020:
You bet!

keir-rex:
@jasonodonnell this branch has been deleted - can I get a copy for testing Raft?

ngarafol:
> @jasonodonnell this branch has been deleted - can I get a copy for testing Raft?

You can check this pull request and merge locally, like I did. Here are instructions https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally

PR #63, opened 2019-09-27
build: add circleci config

malnick:
Adds a circleci configuration for executing bats unit tests. Acceptance tests will require more care in CI and will come in a later PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,13 @@
+version: 2
+jobs:
+  bats-unit-test:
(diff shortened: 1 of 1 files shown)


malnick:
All tests pass but the UI one which a fix was pushed for in https://github.com/hashicorp/vault-helm/pull/57. @jasonodonnell it looks like the test is back to returning `null`.

I'm going to merge this in so our other PRs pick up the status check and we can push a fix for that test specifically in a follow up PR.

PR #76, opened 2019-10-08
Add statefulset security context

dclark:
- Set runAsGroup to 1000
- Set runAsUser to 100
- Remove chown of /tmp/storageconfig.hcl


dclark:
By default the vault container is run as the the root user, this PR updates the securityContext to run as 
a non root user and sets the vault user and group.

jasonodonnell:
Hi @dclark, thanks for the contribution!

More was needed to make this feature work without subtle bugs so I went ahead and did that here: https://github.com/hashicorp/vault-helm/pull/76.  It's currently in review.

Thanks!

PR #80, opened 2019-10-10
Require vault to run as non root

jasonodonnell:
This removes the need to run as root user by adding extra security contexts and setting an emptyDir memory volume in the Vault users home directory (so Vault can cache tokens there if needed).

I've removed the `mlock` configurable: with this change if mlock is not disabled in Vault config, it will error due to not having root privileges.  Since this is running in Kubernetes already, we know that swap has been disabled.


templates/server-statefulset.yaml
```diff
@@ -41,17 +41,21 @@ spec:
       terminationGracePeriodSeconds: 10
       serviceAccountName: {{ template "vault.fullname" . }}
       securityContext:
+        readOnlyRootFilesystem: true
+        runAsNonRoot: true
+        runAsGroup: 1000
+        runAsUser: 100
```
malnick: One thing I just realized here is if someone overrides our docker image with their own the UID and GID may differ, so we may want to either document that in the docs or comment inline here, or consider making this a template variable.

jasonodonnell: I can make this configurable, good point @malnick !

ryno696:
Kk

ryno696:
Yesbu can

ryno696:
Yesbu can

Issue #81, opened 2019-10-11
Add StatefulSet update strategy


dclark:
The updatestrategy is currently set to OnDelete, allow for the values to define the strategy to be the alternative RollingUpdate.

jasonodonnell:
Hi @dclark ,

The `OnDelete` strategy is on purpose to prevent the active node from being upgraded prior to standby nodes when using HA deployments.  RollingUpdates do not provide the coordination required to orchestrate this upgrade pattern.

dclark:
@jasonodonnell understood. A feature to reviewed if there is a pod priority which can be applied which would affect the order of the RollingUpdate. So the standby nodes would be a higher priority allowing for the active node to be updated last.

j-q-in-berlin:
I see this was resolved here: https://github.com/hashicorp/vault-helm/pull/172

PR #82, opened 2019-10-11
Add StatefulSet update strategy configuration

dclark:
Fixes issue #81 

Default is still set to OnDelete as per the original chart. Changing to the default to RollingUpdate may be preferred.


dclark:
Closing due to an issue if the standby nodes are upgraded before the master.

PR #83, opened 2019-10-13
Add load balancer source range for UI service

savagete2860:
Configure the UI service so that it is able to be locked down to specified source IPs if specified. This same logic or similar should be added into all the services.


jasonodonnell:
Hi @savagete2860, thanks for the contribution!

Can you add unit tests for this configurable?  They can be found in `test/unit`.

savagete2860:
Updated. The output for that test:
✓ ui/Service: set loadBalancerSourceRanges when LoadBalancer is configured as serviceType

PR #89, opened 2019-10-19
Remove readOnlyRootFilesystem from securityContext

charlieegan3:
This is a psp restriction rather than something we set here.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -41,7 +41,6 @@ spec:
       terminationGracePeriodSeconds: 10
       serviceAccountName: {{ template "vault.fullname" . }}
       securityContext:
-        readOnlyRootFilesystem: true
         runAsNonRoot: true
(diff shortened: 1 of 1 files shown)


jasonodonnell:
@charlieegan3 Thanks for the contribution!

I think we should make this configurable with a default of `true`.  Thoughts?

charlieegan3:
Hmm, I got an error from this when applying which I've now lost to my terminal history. I was confused as it wasn't in kubectl explain.

```sh
$ kubectl explain pod.spec.securityContext.readOnlyRootFilesystem
error: field "readOnlyRootFilesystem" does not exist
```

I'll close this for now and reopen once I've had a look into it. Sorry for confusion.

jemc:
This should have been merged instead of closed. See my comment in #93.

malnick:
@jemc See my response on #91 and lets keep the convo in that thread, thanks again for the pointers here.

PR #98, opened 2019-10-22
Update README.md

nithasha-samad:


jasonodonnell:
Hi @nithasha-samad, this isn't an improvement over the existing README as it just removes all the formatting.

Issue #68, opened 2019-10-03
Remove image from global variables


tongpu:
Because of the way global variables are implemented in this and the consul chart it is currently impossible to configure the consul chart as a dependency/subchart of the vault chart, because `global.image` with the value `vault:1.2.2` from the vault chart will also be rendered in the consul templates, because it too uses `global.image` for the configuration of the image version.

It is not recommended to use global variables in helm charts in this way and I would propose to switch to the normal way this is implemented in helm/charts, e.g.

```yaml
image:
  repository: vault
  tag: 1.2.2
```

The issue is tracked is also tracked in consul-helm: hashicorp/consul-helm#238

jasonodonnell:
This is something we can look into, thanks for the report @tongpu !

PR #87, opened 2019-10-17
Add possibility to run sidecars with vault

shahbazn:
Problem: This chart does not allow running sidecar containers with vault, that might me needed for some scenarios e.g. running prometheus exporter.

Solution: Add configurable sidecar containers to the chart

PR #103, opened 2019-11-01
Use a standard way to define the container image

jbialy:
Fixes https://github.com/hashicorp/vault-helm/issues/68

Refactor the way that the image repository and tag are defined avoiding globals.

Example:

```
server:
  image:
    repository: vault
    tag: 1.2.2
    pullPolicy: Always
```

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault
-version: 0.2.2
+version: 0.2.1
 description: Install and configure Vault on Kubernetes.
 home: https://www.vaultproject.io
 icon: https://github.com/hashicorp/vault/raw/f22d202cde2018f9455dec755118a9b84586e082/Vault_PrimaryLogo_Black.png
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault
-version: 0.2.1
+version: 0.2.2
 description: Install and configure Vault on Kubernetes.
 home: https://www.vaultproject.io
 icon: https://github.com/hashicorp/vault/raw/f22d202cde2018f9455dec755118a9b84586e082/Vault_PrimaryLogo_Black.png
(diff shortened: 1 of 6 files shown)


jbialy:
Hi @jasonodonnell, any chance of getting this merged?

jasonodonnell:
@jbialy This will not get merged until after v0.2.0 releases this week.  

I've been thinking about how this will conflict with additional images we're going to need in the near future (adding additional sidecars and other services).  The helm chart is going to need more than one image, similar to consul-helm.

jbialy:
> @jbialy This will not get merged until after v0.2.0 releases this week.
> 
> I've been thinking about how this will conflict with additional images we're going to need in the near future (adding additional sidecars and other services). The helm chart is going to need more than one image, similar to consul-helm.

Adding images (for things like sidecar containers) shouldn't be a problem. There's actually a [PR](https://github.com/hashicorp/vault-helm/pull/87) for allowing `extraContainers` to be added to the pod. With this change, any additional containers and their images can be defined directly without any affect on the base image used for Vault.

If there's a specific component (in addition to the Vault image) that needs to be added to the chart it can have it's own scope in the values files as so:

```
my-component:
  image: 
  tag:
```

Does this help?

jasonodonnell:
@jbialy Yes, that does help.  I think the `image` block you've defined needs to be moved into `server`.  We'll be rolling out more components soon and I think it makes sense to do this now.

So this should be:
```yaml
server:
  image:
    repository: vault
    tag: 1.2.2
    pullPolicy: Always
```

Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault
-version: 0.2.1
+version: 0.2.2
```
jasonodonnell: Please remove this version bump, we'll bump this when we're ready to release a new version of the chart.

jbialy:
@jasonodonnell I've gone ahead and made the modification. The `image` block is now inside the `server` block!

PR #104, opened 2019-11-05
Configure pod probes

kalinon:
While trying to use GKE ingresses with vault i ran into issues with the generated load balancers.

Current issue is that the created Google Backend is creating health checks on `/`. This is because the readinessProbe is set to use exec. See: https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#health_checks 

image

This is a problem because `/` redirects to `/ui/` causing the health check to fail.
```
curl 127.0.0.1:8200 -v
* Rebuilt URL to: 127.0.0.1:8200/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8200 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:8200
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 307 Temporary Redirect
< Cache-Control: no-store
< Content-Type: text/html; charset=utf-8
< Location: /ui/
< Date: Tue, 05 Nov 2019 13:48:30 GMT
< Content-Length: 40
<
<a href="/ui/">Temporary Redirect</a>.

* Connection #0 to host 127.0.0.1 left intact
```

This can be fixed by using `/v1/sys/health?standbyok` instead. However when setting a new Health Check to a new one, it is eventually reverted.

This PR adds optional values that can be defined to configure the readiness and liveness probes based on user needs. If nothing is defined it will operate as normal with the readiness exec command.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -102,6 +102,16 @@ spec:
             periodSeconds: 3
             successThreshold: 1
             timeoutSeconds: 5
+          {{- if .Values.server.livenessProbe.enabled }}
+          livenessProbe:
+            httpGet:
+              path: {{ .Values.server.livenessProbe.path | quote }}
+              port: 8200
+            initialDelaySeconds: 5
+            periodSeconds: 3
+            successThreshold: 1
+            timeoutSeconds: 5
+          {{- end }}
       {{- if .Values.global.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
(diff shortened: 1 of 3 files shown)


jasonodonnell:
@kalinon This contribution is awesome and would love to merge it.  This will likely not merge until after the `v0.2.0` release this week.

kalinon:
> Can you add unit tests for this configurable? You can find the tests in `test/unit`.

Yes, i will try to do this soon!

kalinon:
Tests added and merge conflicts resolved

jasonodonnell:
@kalinon Another conflict, this looks good though!

kalinon:
> @kalinon Another conflict, this looks good though!

think that should do it

PR #120, opened 2019-11-22
Return support for `readOnlyRootFilesystem`

mbrancato:
This adds the configurable `readOnlyRootFilesystem` back that was removed in v0.2.1. It has been moved from `.spec.template.spec.securityContext` (v1/PodSecurityContext) to `.spec.template.spec.container[0].securityContext` ([v1/SecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#securitycontext-v1-core))


jasonodonnell:
Thinking on this more, I think a memory volume could be added to `/home/vault` to give us a place to render the template and other caching things (such as login tokens temporarily).

mbrancato:
@jasonodonnell yes. I can look at doing `spec.volumes.emptyDir.medium: Memory`.

mbrancato:
I added a tmpfs mount for `/tmp`, if you know of other volumes needed I could add those. I do see folders under `/vault/`, but they are configMaps or using PVC. After making this change, I was able to launch and unseal a cluster using the chart.

mbrancato:
I noticed that `/home/vault` is used in some cases (I think dev mode). I've added this and fixed some tests needed.

jasonodonnell:
Hi @mbrancato, thanks for doing this!

I think it's probably best to have the `/home/vault` directory have the mounted volume and change the arg to render the configuration file there: https://github.com/hashicorp/vault-helm/blob/master/templates/_helpers.tpl#L117-L124.

This would remove the memory volume at `/tmp`.  `/home/vault` is definitely required for things like token caching and isn't configurable.

mbrancato:
@jasonodonnell I've updated this to remove the use of `/tmp` in preference for `/home/vault` as ephemeral storage and tested locally.

jasonodonnell:
@mbrancato This has conflicts, can you resolve them?

mbrancato:
@jasonodonnell - I've rebased this and fixed the conflict

jasonodonnell:
Looks good but needs conflicts to be resolved.

samueltorres:
Hi, 
Our internal container scanning tools is reporting this issue, and we really wanted this fixed. Is there any ETA for this PR to be merged ? :)

abhilashbs1981:
Getting below error after making [readOnlyRootFilesystem] as true as below in values.yaml
       container:
         allowPrivilegeEscalation: false
         readOnlyRootFilesystem: true

controlplane $ k logs -f vault-0
cp: can't create '/tmp/storageconfig.hcl': Read-only file system

mchoy:
I am getting the same error after I set readOnlyRootFilesystem: true.
Is there a plan to merge this PR eventually?

PR #124, opened 2019-11-27
fix service account annotations

clowdi:
there is a issue with the service account when running this chart, the PR fixes this

### DIFF
--- a/templates/server-serviceaccount.yaml
+++ b/templates/server-serviceaccount.yaml
@@ -10,5 +10,6 @@ metadata:
     app.kubernetes.io/name: {{ include "vault.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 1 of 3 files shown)


templates/server-serviceaccount.yaml
```diff
@@ -10,5 +10,6 @@ metadata:
     app.kubernetes.io/name: {{ include "vault.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
-  {{ template "vault.serviceaccount.annotations" . }}
+  annotations:
+  {{ template "vault.serviceAccount.annotations" . }}
```
jasonodonnell: This is incorrect, there's already an [annotations in entry in the template](https://github.com/hashicorp/vault-helm/pull/124/files#diff-93c03d74bec56c58046cc102d9474f25R255).  This would duplicate annotations.

PR #128, opened 2019-11-29
Remove chart from labels

jasonodonnell:
The label `{{ include "vault.chart" . }}` has the version included in the name, which prevented upgrades on StatefulSets.  Removed this from the StatefulSet label to prevent this from happening.


johncblandii:
Tested this on our cluster and it worked after we ran it with `--force` one time to get through the error.

sysadmind:
For future google searcher, I wasn't able to get past the upgrade with the `--force` command. I was able to workaround the issue by deleting the statefulset with the `--cascade=false` flag to leave the pods behind and then apply the helm upgrade.

Issue #137, opened 2019-12-06
Statefulset Liveness Probe failing on standby nodes due to SSL, initial delay and 429 response


dclark:
- Liveness probe failing when tls is enabled
 - Liveness and readiness probe on standby's receiving a 429 response rather than a 200
 - Liveness probe initialDelaySeconds is set to 5 seconds, value should be configurable and set to a higher default to allow for vault unsealing.

jasonodonnell:
Merged, closing.

PR #138, opened 2019-12-06
Statefulset Liveness Probe failing on standby nodes due to SSL, initi…

dclark:
…al delay and 429 response #137

livenessProbe

 * Set the scheme to the vault.scheme to ensure that the check is works tls enabled or not.

 * Allow a configurable value initialDelaySeconds rather than the set 5 seconds.

 * Set the default initialDelaySeconds to 60 seconds before the probe starts to allow for vault unsealing.

 * Set the path to /v1/sys/health?standbyok=true to ensure a 200 response on standbys.

readinessProbe

 *  Set the path comment to /v1/sys/health?standbyok=true to ensure a 200 response on standbys.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -90,6 +90,7 @@ spec:
             httpGet:
               path: {{ .Values.server.readinessProbe.path | quote }}
               port: 8200
+              scheme: {{ include "vault.scheme" . | upper }}
             {{- else }}
             # Check status; unsealed vault servers return 0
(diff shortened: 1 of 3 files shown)


dclark:
Unit Tests have been updated, both unit and acceptance tests are passing.

jasonodonnell:
@dclark One more request, can you add `scheme` to the HTTP readiness probe?

dclark:
@jasonodonnell The pull request has been updated.

PR #146, opened 2019-12-10
Update chart and tests to Helm 3

mpiekunk:
I have updated the chart for Helm v3 along with all unit and acceptance tests.  This is a breaking change in that Helm v3 is required to install the chart, and any existing deployments would need to be migrated from v2 to v3 via instructions here: https://helm.sh/docs/topics/v2_v3_migration/.  I opted to make a clean switch instead of trying to support/verify both Helm v2 and v3.  I could probably rework it to support both versions, but it would require some decisions on if separate unit/acceptance tests will be maintained for both.  Let me know your thoughts and if this is something you want to move forward with.


tvoran:
Thanks again for the contribution! We've decided to continue this in another PR.

Superseded by #195

PR #149, opened 2019-12-16
Use correct default value for liveness probe path

fischerman:
The query parameter `standbyok` must be set to `true` to have any effect. With the current version the standby pod is restarted due to 429 HTTP responses.


fischerman:
Duplicate of #138

PR #150, opened 2019-12-18
Add vault agent injector

jasonodonnell:
This adds support for the upcoming Vault Agent Injector.  This expanded the values to include a new stanza, `injector`, which wraps all the configurable for deploying the admission controller.  The injector requires Vault 1.3.1.

This also changes `server.authDelegator.enabled=true` by default to improve the installation experience for users who are most likely going to use the Kube auth method.  By setting this too true, the necessary role and clusterrolebinding is automatically created for the installation of Vault.

### DIFF
--- a/test/acceptance/_helpers.bash
+++ b/test/acceptance/_helpers.bash
@@ -128,3 +128,32 @@ wait_for_ready() {
     echo "${POD_NAME} never became ready."
     exit 1
 }
+
+wait_for_complete_job() {
+	POD_NAME=$1
+
+    check() {
+        # This requests the pod and checks whether the status is running
+        # and the ready state is true. If so, it outputs the name. Otherwise
+        # it outputs empty. Therefore, to check for success, check for nonzero
+        # string length.
+        kubectl get job $1 -o json | \
+            jq -r 'select(
+                .status.succeeded == 1 
+            ) | .metadata.namespace + "/" + .metadata.name'
+    }
+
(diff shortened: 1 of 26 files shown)


test/acceptance/_helpers.bash
```diff
@@ -128,3 +128,32 @@ wait_for_ready() {
...
+        kubectl get job $1 -o json | \
+            jq -r 'select(
+                .status.succeeded == 1 
+            ) | .metadata.namespace + "/" + .metadata.name'
+    }
+
+    for i in $(seq 60); do
+        if [ -n "$(check ${POD_NAME})" ]; then
+            echo "${POD_NAME} is complete."
+            sleep 10
```
malnick: Not a blocking comment, but this test could run for a long time. I'm wondering if 3-5s is more appropriate?

values.yaml
```diff
@@ -11,6 +11,56 @@ global:
   # TLS for end-to-end encrypted transport
   tlsDisable: true
 
+injector:
+  # True if you want to enable vault agent injection.
+  enabled: true
+
+  # image for vault-k8s that contains the injector
+  image: "hashicorp/vault-k8s:0.1.0"
```
malnick: Let's make sure this is published before merging this.

test/acceptance/_helpers.bash
jasonodonnell: Lowered!

PR #172, opened 2020-01-13
Add configuration for StatefulSet updateStrategy

lawliet89:
If Auto Unseal is configured, it should be safe to allow a rolling update strategy. This should be an option for a less involved upgrade process.

I have been using the "incubator/vault" chart which deploys Vault as a Deployment and I've not had issues performing upgrades.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -15,7 +15,7 @@ spec:
   podManagementPolicy: Parallel
   replicas: {{ template "vault.replicas" . }}
   updateStrategy:
-    type: OnDelete
+    type: {{ .Values.server.updateStrategyType }}
   selector:
     matchLabels:
       app.kubernetes.io/name: {{ template "vault.name" . }}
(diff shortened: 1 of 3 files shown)


jasonodonnell:
Hi @lawliet89, thanks for the contribution.

The update strategy is this way to avoid upgrading the primary Vault server before replicas in HA clusters.  I think would probably make sense if you aren't running in HA mode, but this isn't helpful for HA clusters.  Incubator can get away with rolling updates because it's a single deployment, not a statefulset.

lawliet89:
Could you explain to me how a rolling update for a StatefulSet would differ from a Deployment please? I think it would be nice to have an option (that is not default) to have a "hands-off" upgrade approach. Otherwise, it's quite laborious to upgrade Vault.

jasonodonnell:
@lawliet89 Actually, I was wrong, there wouldn't be a difference between the two.  I'm going to merge this as there are use cases for non-HA deployments.

fischerman:
@lawliet89 Beware that this does not free you from recreating the pods if the config file changes

Issue #142, opened 2019-12-09
doc: add log rotation and log shipping sidecar examples


jasonodonnell:
Add an example to the documentation that sets up Vault Helm to include a sidecar container that rotates Vault audit logs to satisfy this issue: https://github.com/hashicorp/vault-helm/issues/109.

This would use `server.extraContainers` to add in any other containers we may need.

References: 
- Doc: https://www.vaultproject.io/docs/platform/k8s/helm.html
- Source: https://github.com/hashicorp/vault/tree/master/website/source/docs/platform/k8s

sidewinder12s:
Are there any examples of how to setup these sidecars?

I'm running into a lot of trouble trying to get this working correctly with the number of security bits enabled that make most off the shelf sidecar containers just break because they all assume they can run as root.

sandeep0701:
Have you found any workaround for this ..? 
> Are there any examples of how to setup these sidecars?
> 
> 
> 
> I'm running into a lot of trouble trying to get this working correctly with the number of security bits enabled that make most off the shelf sidecar containers just break because they all assume they can run as root.

sidewinder12s:
I ended up writing a custom container with logrotate & a go based cron scheduler. Due to the security settings on Vault most OSS log rotate containers just don't work and I wish I had just started with something custom.

lhw:
I know reviving an old issue is kind of annoying. But this is the best fitting place right now as its the first search hit for the exact issue.

> I ended up writing a custom container with logrotate & a go based cron scheduler. Due to the security settings on Vault most OSS log rotate containers just didn't work and I wish I had just started with something custom.

I took this as inspiration to update this old issue with a solution for the problem at hand:
Here is the go binary and container which is specifically setup for this helm chart deployment (uid/gid): https://github.com/HanseMerkur/vault-logrotate just build it according to the dockerfile and it should work.

The values for the most recent helm charts (v0.20.1) look something like this:
```yaml
vault:
  server:
    # HUP signal for logrotate
    shareProcessNamespace: true
    # Add the lograte config from a config map
    volumes:
      - name: logrotate-config
        configMap:
          name: logrotate-config
    # And finally the container
    extraContainers:
     - name: auditlog-rotator
       image: vault-logrotate:latest
       imagePullPolicy: Always
       env:
         - name: CRONTAB
           value: "*/5 * * * *"
       volumeMounts:
       - mountPath: /etc/logrotate.conf
         name: logrotate-config
[...]

PR #174, opened 2020-01-13
make shareProcessNamespace configurable

fischerman:
I've noticed that some unit tests are replicated for standalone and HA. What's the way to go here? For now, I've only added standalone tests.

related to #142


fischerman:
rebased

PR #178, opened 2020-01-14
Update to v0.3.3

jasonodonnell:


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+## 0.3.3 (January 14th, 2020)
+
```
kalafut: Any Security section is listed first in the Vault changelog, and I think we should copy that pattern here.

jasonodonnell: Done!

PR #181, opened 2020-01-17
Label selector and failure policy for injector

fischerman:
This is related to https://github.com/hashicorp/vault-k8s/issues/40#issuecomment-572445785

This is just the change in the `values.yaml`. I've tried to keep it flexible but excluded options which don't make sense (e.g. `defaultFailure`). This change would be non-breaking and as long as the label selector is disabled Kubernetes 1.15 is not required.

If I get positive feedback, I can implement the rest.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -43,15 +43,15 @@ injector:
   # Label selectors are only available in Kubernetes 1.15 or higher.
   # Possible options:
   # *disabled*: The webhook is applied to all pods.
-  # *explicit*: The webhook is only applied to pods with the label present. Set the label to *failure* to enfore the injection (else the pod is rejected).
-  # *defaultIgnore*: If the label is not present, the webhook is applied but any errors are ignored. Set the label to *failure* to enfore the injection (else the pod is rejected).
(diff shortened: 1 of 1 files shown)


jasonodonnell:
@fischerman I can see the value of this contribution, I think it would be safe to proceed!

jasonodonnell:
Bump on this, you still considering adding this @fischerman ?  No worries if not.

sfynx:
Yeah, would be appreciated. Right now I'm patching the mutating webhook configuration myself so I can set the failure policy to prevent any Pods from starting without their init/sidecar container when the admission controller happens to be down... but would love to have this configurable in the chart values.

fischerman:
I'm no longer working on the project for which I intended this feature. Maybe I find some time to implement it in the future but I can't make any promises. I still find it a useful addition. In the meantime, if someone wants to continue working on it, feel free!

jasonodonnell:
Configuring failure policy is supported. Label selector isn't supported but we're adding it in the near future. Closing.

PR #195, opened 2020-02-03
Helm 3 support

tvoran:
- [x] unit tests passing
- [x] acceptance tests passing


test/acceptance/server-ha.bats
```diff
@@ -95,17 +94,17 @@ setup() {
   kubectl create namespace acceptance
   kubectl config set-context --current --namespace=acceptance
 
-  helm install https://github.com/hashicorp/consul-helm/archive/v0.8.1.tar.gz \
-    --name consul \
+  helm install consul \
+    https://github.com/hashicorp/consul-helm/archive/v0.8.1.tar.gz \
```
jasonodonnell: Let's bump this version?

tvoran: Sure, bumped to 0.16.2.

jasonodonnell:
@tvoran I think you need to bump the Chart API version to v2 as well

Chart.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 name: vault
```
JulienBreux:
```suggestion
name: vault
type: application
```

tvoran: Thanks for the review! I think we'll leave this out for now since it looks like `application` is the default.

JulienBreux: 🙏

PR #196, opened 2020-02-04
adding custom Vault Address value for injector deployment

mikhailponomaryov:
Addresses the issue with running Vault and custom service discovery like Consul.
Agent injector defaults to Kubernetes native service discovery so in order to use consul it's required to put `vault.hashicorp.com/service` for each pod. This PR makes it configurable on the servers side.

Unit/Acceptance tests are passing.


reynaldi-wijaya-dragon:
+1 for this feature!

hayderimran7:
👍 need this

tvoran:
Thanks for the patch! This was on the right track, but has been superseded by #207.

PR #198, opened 2020-02-10
Remove IPC_LOCK capability

pabrahamsson:
Since we know we're deploying to k8s and are already [disabling `mlock`](https://github.com/hashicorp/vault-helm/blob/master/templates/server-config-configmap.yaml#L16) (also see #80), is there a need for `IPC_LOCK`?
This also helps with deploying on for example Openshift as we now only need `--set server.uid` and `--set server.gid` for the ranges allowed by the `restricted` SCC. No more mucking around with SCCs and capabilities.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -51,9 +51,6 @@ spec:
       containers:
         - name: vault
           {{ template "vault.resources" . }}
-          securityContext:
-            capabilities:
-              add: ["IPC_LOCK"]
           image: {{ .Values.server.image.repository }}:{{ .Values.server.image.tag | default "latest" }}
           imagePullPolicy: {{ .Values.server.image.pullPolicy }}
           command: {{ template "vault.command" . }}
--- a/test/acceptance/server-ha.bats
+++ b/test/acceptance/server-ha.bats
@@ -18,11 +18,6 @@ load _helpers
     jq -r '.initialized')
   [ "${init_status}" == "false" ]
 
-  # Security
-  local ipc=$(kubectl get statefulset "$(name_prefix)" --output json |
-    jq -r '.spec.template.spec.containers[0].securityContext.capabilities.add[0]')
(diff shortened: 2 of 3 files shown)


tongpu:
Always great to see that someone already created the PR I just wanted to start working on. :+1:

@jasonodonnell The only way that I can envision that this could break existing deployments is if someone would have explicitly defined `disable_mlock = false` in either `server.standalone.config` or `server.ha.config`.

tongpu:
@pabrahamsson You probably should also remove the two test cases at https://github.com/hashicorp/vault-helm/blob/master/test/acceptance/server-ha.bats#L21-L24 and https://github.com/hashicorp/vault-helm/blob/master/test/acceptance/server.bats#L24-L24

pabrahamsson:
Thank you @tongpu, updated.

karabijavad:
this would also be nice for allowing deployment on eks + fargate

pabrahamsson:
@jasonodonnell is there anything I can do to help getting this PR merged?

karabijavad:
@pabrahamsson you would probably want to conditionally have the securityContext capability exist based on some .Values parameter

some people may want that security context field there

perhaps a .Values.mlock_enabled bool, defaulting to true, which adds the securityContext.capabilities element?

pabrahamsson:
@karabijavad while I can do what you suggest I'd like to understand the reasoning. The conditionals you're mentioning were already removed as part of #80. `mlock` is  disabled by default (and can't be enabled through values.yaml) hence what's the purpose of `IPC_LOCK`? It sounds to me like either both `mlock` and `IPC_LOCK` are supported  (through values.yaml) or both should be removed. Am I missing something?

karabijavad:
@pabrahamsson my apologies. i had no idea about that. youre right.

lucaspouzac:
Hello, could you release new version with this fix please ?

Thanks!

PR #213, opened 2020-02-24
Change config specification

davidmogar:
As it is right now, the specification of the config is done through a string. When using storage backends like PostgreSQL, the password for the database has to be included in the config variable of the values file.

This change allows to specify the configuration through a map, making the chart GitOps friendly (eg. using it with [Flux](https://github.com/fluxcd/flux)). Now, sensitive values can be stored in a different values file or passed on deployment time with `--set`.

To have a very generic specification:
- I've assumed that the combination stanza (eg. storage) name (eg. file) is unique.
- Quoted values for all stanza parameters. I tested a generated configuration in a vault docker image and it seems to work just fine.

As an example, your values file could contain something like this:
```yaml
server:
  standalone:
    config:
      ui: true
      listener:
        tcp:
          address: "[::]:8200"
          cluster_address: "[::]:8201"
          tls_disable: 1
      storage:
        file:
          path: "/vault/data"
      telemetry:
        statsite_address: "127.0.0.1:8125"
        disable_hostname: true
```
**Edit**:

As stated by @jasonodonnell, Vault support json as a format for the configuration. The new change will generate the following ConfigMap when providing the values shown above:
```yaml
---
# Source: vault/templates/server-config-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: RELEASE-NAME-vault-config
  namespace: default
  labels:
    helm.sh/chart: vault-0.4.0
    app.kubernetes.io/name: vault
data:
  extraconfig-from-values.hcl: |-
    {
      "disable_mlock": true,
      "listener": {
        "tcp": {
          "address": "[::]:8200",
          "cluster_address": "[::]:8201",
          "tls_disable": 1
        }
      },
      "storage": {
        "file": {
          "path": "/vault/data"
        }
      },
      "telemetry": {
        "disable_hostname": true,
        "statsite_address": "127.0.0.1:8125"
      },
      "ui": true
    }
```

### DIFF
--- a/templates/server-config-configmap.yaml
+++ b/templates/server-config-configmap.yaml
@@ -14,7 +14,7 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
   extraconfig-from-values.hcl: |-
-  {{- if or (eq .mode "ha") (eq .mode "standalone")}}
+  {{- if or (eq .mode "ha") (eq .mode "standalone") }}
   {{- $type := typeOf (index .Values.server .mode).config }}
   {{- if eq $type "string" }}
     disable_mlock = true
@@ -26,8 +26,11 @@ data:
     {{ tpl .Values.server.ha.raft.config . | nindent 4 | trim }}
   {{ end }}
   {{- else }}
-  {{- $config := merge (dict "disable_mlock" true) (index .Values.server .mode).config }}
-{{ $config | toPrettyJson | indent 4 }}
+  {{- if and (eq .mode "ha") (eq (.Values.server.ha.raft.enabled | toString) "true") }}
+{{ merge (dict "disable_mlock" true) (index .Values.server .mode).raft.config | toPrettyJson | indent 4 }}
+  {{- else }}
+{{ merge (dict "disable_mlock" true) (index .Values.server .mode).config | toPrettyJson | indent 4 }}
+  {{- end }}
   {{- end }}
   {{- end }}
 {{- end }}
--- a/templates/server-config-configmap.yaml
+++ b/templates/server-config-configmap.yaml
@@ -1,7 +1,7 @@
 {{ template "vault.mode" . }}
 {{- if ne .mode "external" }}
 {{- if and (eq (.Values.global.enabled | toString) "true") (ne .mode "dev") -}}
-{{ if or (ne .Values.server.standalone.config "")  (ne .Values.server.ha.config "") -}}
+{{ if or (.Values.server.standalone.config) (.Values.server.ha.config) -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/templates/server-config-configmap.yaml
+++ b/templates/server-config-configmap.yaml
@@ -15,13 +15,10 @@ metadata:
 data:
   extraconfig-from-values.hcl: |-
     disable_mlock = true
-  {{- if eq .mode "standalone" }}
-    {{ tpl .Values.server.standalone.config . | nindent 4 | trim }}
-  {{- else if and (eq .mode "ha") (eq (.Values.server.ha.raft.enabled | toString) "false") }}
-    {{ tpl .Values.server.ha.config . | nindent 4 | trim }}
-  {{- else if and (eq .mode "ha") (eq (.Values.server.ha.raft.enabled | toString) "true") }}
-    {{ tpl .Values.server.ha.raft.config . | nindent 4 | trim }}
-  {{ end }}
+  {{- if or (eq .mode "ha") (eq .mode "standalone")}}
+  {{- $config := merge (dict "disable_mlock" true) (index .Values.server .mode).config }}
+{{ $config | toPrettyJson | indent 4 }}
+  {{- end }}
 {{- end }}
 {{- end }}
 {{- end }}
--- a/templates/server-config-configmap.yaml
+++ b/templates/server-config-configmap.yaml
@@ -14,10 +14,21 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
   extraconfig-from-values.hcl: |-
-    disable_mlock = true
   {{- if or (eq .mode "ha") (eq .mode "standalone")}}
+  {{- $type := typeOf (index .Values.server .mode).config }}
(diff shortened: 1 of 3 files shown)


jasonodonnell:
Hi @davidmogar, thanks for the contribution!

This feature has been on our list to do for some time, so I'm happy to see some progress here.  I'm wondering if this feature could simply be reduced to using the `toJson` since JSON is a valid HCL alternative.

For example, the config could be defined such:

```yaml
server:
  standalone:
    config:
      ui: true
      listener:
        tcp:
          address: "[::]:8200"
          cluster_address: "[::]:8201"
          tls_disable: 1
      storage:
        file:
          path: "/vault/data"
      telemetry:
        statsite_address: "127.0.0.1:8125"
        disable_hostname: true
```

Then the configmap could simply be:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: "{{ template "vault.fullname" . }}-config"
data:
  config.json: |
    {{ .Values.server.standalone.config | toJson }}
```

Thoughts on your implementation verse this approach?

davidmogar:
I wasn't aware of the possibility of specifying the config through json. In this case the code is way more simple and we can use `toPrettyJson` to have a much friendly output. As shown in the new commit.

Not sure about the need for `disable_mlock` as not the docs mention is not suit for prod. If needed, could be merged into the generated json.

jasonodonnell:
@davidmogar `disable_mlock` is set for a few reasons (and will need to stay):

* swap is disabled in Kubernetes, so `disable_mlock` doesn't really do anything
* the container isn't running as root, so it doesn't have privileges to mlock (which causes errors in the startup script).

Hope that helps!

templates/server-config-configmap.yaml
```diff
@@ -14,43 +14,7 @@ metadata:
...
-      {{ $parameter_key }} = {{ $parameter_value | quote }}
-          {{- end }}
-        {{- end }}
-    }
-      {{- else }}
-    {{ $key }} = {{ $value }}
-      {{- end }}
-    {{- end }}
-  {{ end }}
+{{ .Values.server.standalone.config | toPrettyJson | indent 4 }}
```
jasonodonnell: We'll need to support all modes and their respective configurations here.

davidmogar: Yeah, sure. Was just thinking in a very loudly way ;)

Last commit has the actual code, taking the config from standalone or from dev and merging the default parameter.

In the future the way the config and the mode is defined could be improved to have a more straightforward values file (modify a single key to change the mode for example).

davidmogar:
Documentation for the change: https://github.com/hashicorp/vault/pull/8423

davidmogar:
Any news on this?

jasonodonnell:
Sorry for the delay @davidmogar, going to review this today.  My only hesitation is this breaks existing functionality.  On the other hand this is a way better UX.  If you get a chance please resolve conflicts.

davidmogar:
No worries @jasonodonnell. Was merely a heads-up. I've fixed the conflicts. Let me know if you need anything else.

nesl247:
I wouldn't worry about breaking BC. This still isn't even a 1.0 version of the chart.

davidmogar:
@jasonodonnell Any news on this?

jasonodonnell:
Hi @davidmogar, sorry for the long delay.  We talked about this internally across other teams that use Helm and we're definitely hesitant on changing the UX across our products.  That being said, we've experimented with a way of supporting the old style of config (string) vs this new style.  An example of this can be found here: https://github.com/hashicorp/vault-helm/pull/272

I'm wondering if we can modify this PR to include similar logic?  Users who have been using Vault Helm for some time won't need to upgrade their configuration to the new UX and we can gave them fair warning about deprecating the string style in the future.

Thoughts?

davidmogar:
Hi @jasonodonnell,

Makes sense to have both definitions working at the same time. I strongly suggest you to keep the yaml one. It's way more friendly and standard. In the meantime I left the original values file unchanged though. Your call on what to do with it although, again, I would use the yaml definition there too.

Take a look and let me know if the last change goes more on the line of what you want. If so, would be nice if you could put this on the fast track as this PR is already 2 months old and forcing us to use a fork.

templates/server-config-configmap.yaml
```diff
@@ -14,6 +14,9 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
   extraconfig-from-values.hcl: |-
+  {{- if or (eq .mode "ha") (eq .mode "standalone")}}
+  {{- $type := typeOf (index .Values.server .mode).config }}
```
jasonodonnell, line 18: There's actually a new `ha` submode, `raft`, which has it's own configuration.  I think we'll need additional logic for that config file.

In the future we hope to consolidate these modes but we'll need to change this for now.

davidmogar: Sorry, I didn't realize about the new submode. My last commit should address this problem.

In a side note, I think that you are over complicating the values file addressing problems where shouldn't be addressed.

It would be extremely more easy to define a single `config` variable with the specifics. If you want to provide examples, do so by creating multiple values files that the users can choose from or simply add comments. That would simplify the code a lot.

The only difference between modes, apart from the config, is the `disruptionBudget` in the `ha`, which could go outside the mode and be used if defined.

jasonodonnell: Hi @davidmogar, thanks for adding this.

As I mentioned in my last comment:

> In the future we hope to consolidate these modes but we'll need to change this for now.

Testing this now and looking to get it merged today.

davidmogar:
Thank you for merging this, @jasonodonnell.

gw0:
> `disable_mlock` is set for a few reasons (and will need to stay):
>     * swap is disabled in Kubernetes, so `mlock` doesn't really do anything
>     * the container isn't running as root, so it doesn't have privileges to mlock (which causes errors in the startup script).

@jasonodonnell Just noticed these arguments and would like to leave a comment... I find it confusing that the `config` block is not directly used as provided by the user. Overriding values in templates is bad UX and complicates the code. As does having 2 HA modes each with its own `config` block (that are pretty similar).

PR #185, opened 2020-01-29
feat: add AGENT_INJECT_VAULT_AUTH_PATH option to the injector

rasta-rocket:

PR #216, opened 2020-02-28
Add ability to use custom auth path

fbongiovanni29:
It's desirable to use a different auth path when using multiple k8s clusters. I noticed [most of the work](https://github.com/hashicorp/vault-k8s/blob/ed146a2114aa1cc08c12f6dc29fae5ee23c69552/subcommand/injector/flags.go#L51) was already there to accomplish this and wanted to add the functionality to the Helm Chart

I will be out most of next week, I can get to suggestions when I get back - feel free to make any changes


fbongiovanni29:
Just realized this was already an open PR https://github.com/hashicorp/vault-helm/pull/185/files

PR #219, opened 2020-03-06
Add new vault-k8s environment variables

jasonodonnell:
This adds the new envs included in newest release of Vault K8s: https://github.com/hashicorp/vault-k8s/blob/master/CHANGELOG.md#030-march-5th-2020.

Updated the images to reflect the latest releases.

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -60,6 +60,10 @@ spec:
             - name: AGENT_INJECT_TLS_AUTO_HOSTS
               value: {{ template "vault.fullname" . }}-agent-injector-svc,{{ template "vault.fullname" . }}-agent-injector-svc.{{ .Release.Namespace }},{{ template "vault.fullname" . }}-agent-injector-svc.{{ .Release.Namespace }}.svc
             {{- end }}
+            - name: AGENT_INJECT_LOG_FORMAT
+              value: {{ .Values.injector.logFormat | default "standard" }}
+            - name: AGENT_INJECT_REVOKE_ON_SHUTDOWN
+              value: {{ .Values.injector.revokeOnShutdown | default false }}
           args:
             - agent-inject
             - 2>&1
--- a/test/unit/injector-deployment.bats
+++ b/test/unit/injector-deployment.bats
@@ -223,6 +223,22 @@ load _helpers
   [ "${actual}" = "auth/k8s" ]
 }
 
+@test "injector/deployment: defualt logLevel" {
+  cd `chart_dir`
+  local object=$(helm template \
+      --show-only templates/injector-deployment.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].env' | tee /dev/stderr)
+
+  local actual=$(echo $object |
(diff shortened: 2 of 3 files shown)


test/unit/injector-deployment.bats
```diff
@@ -222,3 +222,54 @@ load _helpers
...
+      yq -r '.spec.template.spec.containers[0].env' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+     yq -r '.[8].name' | tee /dev/stderr)
+  [ "${actual}" = "AGENT_INJECT_REVOKE_ON_SHUTDOWN" ]
+
+  local actual=$(echo $object |
+      yq -r '.[8].value' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
tvoran: Should there be tests for the defaults of these parameters too?

values.yaml
```diff
@@ -22,19 +22,28 @@ injector:
...
-    tag: "1.3.2"
+    tag: "1.3.3"
 
   # Mount Path of the Vault Kubernetes Auth Method.
   authPath: "auth/kubernetes"
 
+  # Configures the log verbosity of the injector
+  logLevel: "info"
+
+  # Configures the log format of the injector
```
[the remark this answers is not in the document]
jasonodonnell: Done!

test/unit/injector-deployment.bats
jasonodonnell: Done!

templates/injector-deployment.yaml
```diff
@@ -60,6 +60,10 @@ spec:
             - name: AGENT_INJECT_TLS_AUTO_HOSTS
               value: {{ template "vault.fullname" . }}-agent-injector-svc,{{ template "vault.fullname" . }}-agent-injector-svc.{{ .Release.Namespace }},{{ template "vault.fullname" . }}-agent-injector-svc.{{ .Release.Namespace }}.svc
             {{- end }}
+            - name: AGENT_INJECT_LOG_FORMAT
+              value: {{ .Values.injector.logFormat | default "standard" }}
+            - name: AGENT_INJECT_REVOKE_ON_SHUTDOWN
+              value: {{ .Values.injector.revokeOnShutdown | default false }}
```
jmaroeder: This looks like it needs to be a string (e.g,. "`default "false"`)

jmaroeder: More details:
```console
$ helm install vault ./ --wait
Error: Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false}],"im|..., bigger context ...|"name":"AGENT_INJECT_REVOKE_ON_SHUTDOWN","value":false}],"image":"hashicorp/vault-k8s:0.3.0","imageP|...
```

jasonodonnell: Thanks! Fixed here: https://github.com/hashicorp/vault-helm/pull/221

PR #221, opened 2020-03-06
Change revoke from bool to string

jasonodonnell:
Fixes a bug where bool value must be string.


rvdh:
@jasonodonnell  / @tvoran Please create a new release with this change, currently deploying the 0.4.0 version of this chart fails because of this issue.

tvoran:
Hi @rvdh, the option this PR affected was added after 0.4.0 (in https://github.com/hashicorp/vault-helm/pull/219), so I'm curious what errors you're seeing with 0.4.0.

rvdh:
@tvoran Ah, hm, it's weird because we installed the 0.4.0 version (but from the master branch) and I got the error that this PR fixes. Well never mind and sorry to bother you. We should install from a tag on github anyway instead of the master branch.

PR #223, opened 2020-03-11
use port names that map to vault.scheme

karabijavad:
Istio makes some assumptions based on service port names.
When you configure vault to serve tls, and the service port name is http, then istio assumes that vault is http, and causes weird tls termination behavior.

This branch just changes the port names. Really, only the service port name needed to change, but i changed a couple others for consistency.

### DIFF
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -37,7 +37,7 @@ spec:
       {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
       nodePort: {{ .Values.server.service.nodePort }}
       {{- end }}
-    - name: {{ include "vault.scheme" . }}-internal
+    - name: https-internal
       port: 8201
       targetPort: 8201
   selector:
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -31,7 +31,7 @@ spec:
   # since this DNS is also used for join operations.
   publishNotReadyAddresses: true
   ports:
-    - name: http
+    - name: {{ include "vault.scheme" . }}
       port: {{ .Values.server.service.port }}
       targetPort: {{ .Values.server.service.targetPort }}
       {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -37,7 +37,7 @@ spec:
       {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
       nodePort: {{ .Values.server.service.nodePort }}
       {{- end }}
-    - name: internal
+    - name: {{ include "vault.scheme" . }}-internal
       port: 8201
       targetPort: 8201
   selector:
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -85,9 +85,9 @@ spec:
             - containerPort: 8200
               name: {{ include "vault.scheme" . }}
             - containerPort: 8201
-              name: internal
+              name: {{ include "vault.scheme" . }}-internal
             - containerPort: 8202
-              name: replication
+              name: {{ include "vault.scheme" . }}-replication
           {{- if .Values.server.readinessProbe.enabled }}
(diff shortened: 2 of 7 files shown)


templates/server-statefulset.yaml
```diff
@@ -83,11 +83,11 @@ spec:
...
           ports:
             - containerPort: 8200
-              name: http
+              name: {{ include "vault.scheme" . }}
             - containerPort: 8201
-              name: internal
+              name: {{ include "vault.scheme" . }}-internal
             - containerPort: 8202
-              name: replication
+              name: {{ include "vault.scheme" . }}-rep
```
karabijavad: fyi, this is "rep", not "replication", because the port name must be under <=15 characters

templates/server-service.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
       nodePort: {{ .Values.server.service.nodePort }}
       {{- end }}
-    - name: internal
+    - name: {{ include "vault.scheme" . }}-internal
```
tvoran: Was this change necessary for istio? Or just going for consistency?

templates/server-statefulset.yaml
tvoran: Same question for the internal the replication ports here.

templates/server-service.yaml
karabijavad: i wasnt utilizing this port, so i am not sure.
however, based on this:
https://istio.io/docs/ops/configuration/traffic-management/protocol-selection/

it says that prefixing port names with "http" or "https" informs istio of the protocol of the traffic. but further down it says that istio can do automatic detection. for the most part this is consistency

templates/server-statefulset.yaml
karabijavad: same answer, pretty much just consistency. the only problem i had that was solved was changing 8200 to http-8200. the rest is consistency.

karabijavad:
to be frank, i wish i could provide more detail about why it _doesnt_ work, but i can not, because the istio internals is still rather confusing. but i do know, based on their docs, that istio should be able to manually detect the protocol, but in my case, when vault has handling https, the service name was http, so there must have been some conflict between the "auto detection" and the "manual specification", or something, because manually replacing the port name to https made it work. apologies for not being able to provide further info

tvoran:
@karabijavad Thanks again for digging into this! Would you be able to add some tests around the port naming?

karabijavad:
@tvoran i would love to! i have never messed with helm testing lol. so it might be a few days for me to get around to it... any chance you can link me to an example PR which i can reference as a template for a simple change like this?

tvoran:
Sure, I'd say first take a look at the testing section of the contributing doc: https://github.com/hashicorp/vault-helm/blob/master/CONTRIBUTING.md#testing

And these tests should give you an example to work from: https://github.com/hashicorp/vault-helm/blob/2b2b0dd2fa9e871d7525e35c7eff5a518bbf9c22/test/unit/server-ha-statefulset.bats#L62-L97

Basically I'm thinking something along the lines of testing the port names in each of the templates you've modified with `Values.global.tlsDisable` set to true and false.

karabijavad:
@tvoran what version of yq do you use? yq tells me that "-r" is an unknown short hand

karabijavad:
@tvoran ok ive added tests for server-service, ui-service and server-statefulset

karabijavad:
@tvoran i just realized something though. isnt port 8201, (the "internal" port) always https? so those port names should just be a static "https-internal" if so.

is the same true for replication?

tvoran:
> im not sure what i should do about the replication port though. should that also be http/https, based on tlsDisable? that is the current behavior, but when i realized that the internal port scheme isnt dictated by the 8200 port scheme, then i wondered if the same was true of replication? im assuming that tlsDisable should dictate the replication port, as it does the 8200 port.

I think this is fine the way you have it now, especially since I don't think port 8202 is actually being used; replication happens over 8201 as far as I know.

karabijavad:
@tvoran yea, the map select allows for the the `ports` elements to move around and me not have to worry about updating the index :)

thank you

PR #225, opened 2020-03-11
Add selectors for injector deployment

sshishov:
Added specs to injector deployment:
- affinity
- toleration
- nodeselector

It will give ability to place the pod on proper node.
Better to have separate configuration for server itself and injector, but then it required more changes.

Signed-off-by: Sergei Shishov <sergei.shishov@dubizzle.com>

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -24,6 +24,9 @@ spec:
         app.kubernetes.io/instance: {{ .Release.Name }}
         component: webhook
     spec:
+      {{ template "vault.affinity" . }}
+      {{ template "vault.tolerations" . }}
+      {{ template "vault.nodeselector" . }}
       serviceAccountName: "{{ template "vault.fullname" . }}-agent-injector"
       securityContext:
         runAsNonRoot: true


tvoran:
Thanks for the patch @sshishov! Though as you mentioned, I think it'd be better for the injector to have it's own configs for these, since it can be deployed without the vault server in this chart.

If you're interested in working on this further, take a look at how the helper for extraEnvironmentVars is used for both injector and server; I suspect something similar could be done for the configs in this PR:
https://github.com/hashicorp/vault-helm/blob/fa13c47858ca89076f84378ff554e205116481d6/templates/_helpers.tpl#L285-L296

https://github.com/hashicorp/vault-helm/blob/fa13c47858ca89076f84378ff554e205116481d6/templates/injector-deployment.yaml#L67

https://github.com/hashicorp/vault-helm/blob/58b96dbc1057d863d334d10d67a3fbaf3b73bb02/templates/server-statefulset.yaml#L84

tvoran:
Thanks again for the contribution! We've decided to continue this in another PR.

Superseded by #234

Issue #117, opened 2019-11-18
How can I input a multi-line annotation with vault?


ado120:
```
  # Extra annotations to attach to the ui service
  # This should be a multi-line string mapping directly to the a map of
  # the annotations to apply to the ui service
  annotations: {}

```
This is how my values looks like:
```
ui:
  enabled: true
  serviceType: LoadBalancer
  externalPort: 8200
  annotations: |
    external-dns.alpha.kubernetes.io/hostname: dns_name
    service.beta.kubernetes.io/aws-load-balancer-security-groups: sg_group


```

Getting a Error: YAML parse error on vault/templates/ui-service.yaml: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field .annotations of type map[string]string

Just to note that I do the same exact thing with the consul-helm chart and it works fine

ado120:
Looks like I had to remove the `|` and it worked however this doesn't fall in-line with the consul-helm chart

tvoran:
Hi @ado120, I see what you mean! The comments in the values file for `ui.annotations` are not quite right since that `annotations` value is being parsed as a yaml map. I'll see about correcting the comments there.

tvoran:
And upon further review, annotations for various objects were a mix of multi-line strings or yaml maps, so in #227 the chart now expects all annotations to be multi-line strings. Thanks again for catching this!

Issue #55, opened 2019-09-17
Use native kubernetes ways of configuration


nesl247:
I'm currently using the helm chart in the incubator (https://github.com/helm/charts/tree/master/incubator/vault), and trying to migrate over to this official chart.

Unfortunately a lot of what is done in this helm chart is so much harder that the one from the incubator. Annotations for the pod for example require you to use a string instead of helm parsing it, allowing for the same configuration one would do for kubernetes itself. Same goes for volume and mounts.

jasonodonnell:
Hi @nesl247!

We're always interested in making this easier to use.  I'm trying to understand where we're falling short here.  Can you provide me an example of the UX you'd prefer or point to code that could be changed?

Thanks!

nesl247:
Hi @jasonodonnell,

See how pod annotations are done here: https://github.com/helm/charts/blob/8180c3ffcf90dd5c2bc40e7006715100d7d6386d/incubator/vault/templates/deployment.yaml#L41 as an example. This is a lot easier to work with rather than having to do yaml strings.

jasonodonnell:
@nesl247,

Appreciate the tip, I think this is a good improvement and makes sense to add. Keeping this open to track the work!

davidmogar:
I didn't see this issue when preparing my change, but it makes a lot of sense. Specially when you have to write sensitive information as passwords. I'm trying to cover the config part, more or less, with this change https://github.com/hashicorp/vault-helm/pull/213.

Another thing that puzzles me but I didn't want to introduce in my change is the fact that deployment mode is split and each one define it's config. A single variable could include the mode and another one the config (with any validation done by Helm).

But for sure, my biggest problem right now is the config specification. Having the possibility to do it in a more dynamic way opens the doors to GitOps and secret management.

nesl247:
Any word on if this would be accepted if I were to submit a PR? Working with this helm chart is a bit painful as it does a lot of custom stuff instead of adopting community standards, and using native features of yaml.

tvoran:
@nesl247 What all do you have in mind? Just changing multiline strings to yaml maps, or something more? 

You mentioned Pulumi in another issue, is that the motivation for the changes? Can you elaborate on the issues you’re encountering?

nesl247:
[...]

PR #227, opened 2020-03-14
Making all annotations multi-line strings

tvoran:
Annotations for various objects were either multi-line strings or yaml
maps strings, so this is making them all multi-line strings for
consistency. Also updated the doc comment for namespaceSelector, since
it's being read as a yaml map (toYaml).

Relates to #117

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -244,7 +244,7 @@ Sets extra ui service annotations
 {{- define "vault.ui.annotations" -}}
   {{- if .Values.ui.annotations }}
   annotations:
-    {{- toYaml .Values.ui.annotations | nindent 4 }}
+    {{- tpl .Values.ui.annotations . | nindent 4 }}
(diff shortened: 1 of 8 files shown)


nesl247:
Why would you do a multi-line string instead of a yaml hash for this? This is the "standard" used in pretty much any helm chart I've seen.

tvoran:
I think the main reason is that a multi-line string can be run through `tpl` so that chart values will be interpolated, and for custom annotations at least that seems like a useful thing. The other reason is the sister chart to this one ([consul-helm](https://github.com/hashicorp/consul-helm)) follows a similar convention.

In any case, this PR was just about making annotations consistent within the chart.

nesl247:
I guess that makes sense for people using helm directly. I'm using `pulumi` which is like `terraform`, so I let `pulumi` handle that part.

Either way, I'll discuss this more in #55.

PR #237, opened 2020-03-22
Fix bug with api server env

jasonodonnell:
`VAULT_API_ADDR` env is incorrect in the template.  This fixes that environment variable by removing `-internal`.


ngarafol:
Hi, can we get new tag, like 0.4.1? There are few fixes like this one, raft support and probably more useful stuff. So in order to avoid using master, tag please :)

jasonodonnell:
@ngarafol We don't use minor releases for new features. This will be in the 0.5.0 release in April once Raft goes GA (it's still in beta). Also the fix was for an error introduced in the raft feature, so it should not be present in a tagged release.

PR #240, opened 2020-03-26
Fix hardcoded service name in raft env

jasonodonnell:
Small bug fix where the name of the cluster wasn't being prepended to the service name when deploying Raft clusters.


danielmittelman:
Nice! I was just going to open a PR for that, cool to see the active development of this chart 👍

PR #250, opened 2020-04-06
Add Vault Helm ent support, service discovery

jasonodonnell:
This adds support for Vault Helm enterprise by adding:
* internal service so pods can communicate directly with each other
* new environment variables for better clustering setup
* acceptance tests for DR and perf replication
* new services for HA mode (service discovery labels)

Will change the image to the official 1.4.0 image when it's released.  Keeping it as a release candidate for evaluating this PR (there's 1.4.0 hard requirements).


jasonodonnell:
Initial doc PR here: https://github.com/hashicorp/vault/pull/8678

Issue #251, opened 2020-04-06
Post install automation


rohitrsh:
Hi,

We have recently deployed vault on Kubernetes; the requirement was to setup end to end vault using helm and integrate with the LDAP backend.

Current, vault helm does not support any post-installation tasks. I have customised it the chart and added a post-installation script. Which is doing the following tasks for me

* Checking vault is up and running.
* Initialising the vault.
* Unsealing the Vault
* Enabling the LDAP auth backend
* Writing the LDAP config.

All the above values can be enabled/disabled using `values.yaml.`

Let me know if you are planning to have such kind of solution in the chart. Would be happy to submit the PR

Thank You
Rohit Sharma

jasonodonnell:
PR welcome for `postStart` lifecycle additions! 💯

rohitrsh:
Closing this one as have created new PR for poststart configs

rg13567:
@rohitrsh I am looking for same solution where i can perform vault init ,unseal and ldap config as part of vault deployment with helm.
Can you please share how did you achieve it ?

bakunowski:
> @rohitrsh I am looking for same solution where i can perform vault init ,unseal and ldap config as part of vault deployment with helm. Can you please share how did you achieve it ?

You might want to take a look at [this](https://github.com/jasonodonnell/vault-agent-demo/blob/master/configs/bootstrap.sh) bootstrap script, which is being run in the `postStart` lifecycle hook as described in [this](https://github.com/hashicorp/vault-helm/pull/315) PR

PR #252, opened 2020-04-08
Adding vault poststart hook/configs

rohitrsh:
@jasonodonnell as discussed in https://github.com/hashicorp/vault-helm/issues/251 adding PR. This performs following tasks.

- Check vault status before execution
- Vault init and unseal it - It can be enabled/disables using `values.yaml`
- Enable ldap auth - This is also can be enabled/disabled using `values.yaml`

More auth modes can be control and added using `vault-config.sh` currently it supports `ldap` only.

### DIFF
--- a/files/vault-config.sh
+++ b/files/vault-config.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+set +x
+while ! nslookup vault </dev/null || ! nc -w1 vault 8200 </dev/null; do
+    echo "Waiting for Vault to Come up!"
+    sleep 0.1
+done
+sleep 10
+
+echo "Vault Up, Will be initlising the it"
+
+export VAULT_ADDR=http://$VAULT_SERVICE_HOST:$VAULT_SERVICE_PORT_HTTP
+echo "vault address is: $VAULT_ADDR"
+
+echo "Initialising the vault"
+vault operator init -n 1 -t 1 > /tmp/stdout
+cat /tmp/stdout | head -n 1 | awk '{print $4}' > /tmp/key
+cat /tmp/stdout | grep -i "Root" |awk '{print $4}' > /tmp/token
+export KEY=$(cat /tmp/key)
+export VAULT_TOKEN=$(cat /tmp/token)
+
+echo "vault key is : $KEY"
+echo "vault token is : $VAULT_TOKEN"
+
+echo "Unsealing the vault"
+vault operator unseal $KEY
+vault status
+
+if [ "{{.Values.initvault.ldapauth.enabled}}" == "true" ]; then
+    echo "Enabling the LDAP auth"
+    export ldap_url="{{.Values.initvault.ldapauth.ldap_url}}"
+    export userattr="{{.Values.initvault.ldapauth.userattr}}"
+    export userdn="{{.Values.initvault.ldapauth.userdn}}"
+    export groupdn="{{.Values.initvault.ldapauth.groupdn}}"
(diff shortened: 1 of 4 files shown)


jasonodonnell:
Hi @rohitrsh, my apologies but there seems to be some confusion about what we talked about.  What I was referring to was the `lifeCycle` management hooks built into K8s, specifically `postStart` which will run after the pod is ready: https://github.com/hashicorp/vault-helm/blob/master/templates/server-statefulset.yaml#L128-L133

This job is useful, however, at this time we're not looking to automate initialization.  Instead this PR should be focused on adding `postStart` so that users can optionally add these types of scripts to their install if they wish to automate it.

rohitrsh:
Hi @jasonodonnell , 

Initially i tried with `postStart` lifeCycle hook. To make this work we might need to update the vault statefulset readiness check here https://github.com/hashicorp/vault-helm/blob/2b137c95d2e04fb0ddabb0f94fdd58a7ecf0e5e0/templates/server-statefulset.yaml#L109

Because till the time its not pass  pod status will be `0/1` and `postStart` hook won't execute until we need to login in pod and initialised the vault.  `postStart` hook waits for the pod pass all the checks and come up `1/1`

I do not wanted make changes in the healthcheck thats why I choose this do be done using helm post-install hook. 

If its fine to change the healthcheck to the port `8200`, Same script can be used in lifeCycle hook.

Let me know your thought on that.

jasonodonnell:
@rohitrsh if you use the HTTP check instead, you can make it go ready depending on different criteria:

```yaml
server:
  readinessProbe:
    enabled: true
    path: /v1/sys/health?standbyok=true&sealedcode=204&uninitcode=204
```

This will make Vault report ready even if it's sealed or uninitialized (liveliness probe can be changed to report different status).

rohitrsh:
@jasonodonnell, cool that just tested that worked. One quick question I have how we are planning to pass the `postStart` commands. via script mapped in `configMap`?

jasonodonnell:
@rohitrsh This can already be done via the `extraVolumes` value.  You define a configmap that contains the script, it will be mounted to `/vault/userconfig/<name of configmap>/myscript.sh`, and you would configure the `postStart` value to run that script.  Hope that makes sense!

rohitrsh:
@jasonodonnell,  Not sure which approach is good to have here `pod lifecycle` hooks or `helm hooks`.

The major problem I could see with `PostStart` approach if `PostStart` script got stuck due to mis-config or any other reason pod will get stuck in `PodInitializing` status forever. In this status pod wont stdtout anything hence its very difficult to troubleshoot. (This I have faced very recently with one of the HELM chart I am working on).

Wherein with `helm hook` we are letting the vault pod come up without distributing its startup and watching it through job, as soon as its up. Job will start performing the tasks written in the script. Also user can manage this the same way current `PostStart` script is proposed. If there is nothing in that script job pod will `echo nothing` and gracefully exit.

With this approach if vault pod gets stuck or job script is having any issue it can be easily troubleshoot. Most important, users custom script will not have any impact on vault pod startup.

Let me know your thought on it.

jasonodonnell:
Hi @rohitrsh,

I think supporting helm hooks would be good too, but I'm closing this PR for now since https://github.com/hashicorp/vault-helm/pull/315 was merged to implement `postStart`.

Perhaps we can open a new issue about adding Helm hook support if this is something you think would be valuable?

rg13567:
@rohitrsh I am looking for same solution where i can perform vault init ,unseal and ldap config as part of vault deployment with helm.
Can you please share how did you achieve it ?

PR #255, opened 2020-04-09
fix: default value types don't match real type

nesl247:
This fixes the default values for several of the fields. There may be other cases, but these are the ones I know of.

The default values were the wrong type, so helm would complain about overriding them with a different type.

This should be fully and properly resolved with another PR for #55, but this will resolve peoples errors immediately.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -133,7 +133,7 @@ server:
     enabled: false
     labels: {}
       # traffic: external
-    annotations: {}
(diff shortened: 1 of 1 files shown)


tvoran:
Thanks, but we're going to leave the default for annotations as `{}` for now, since annotations in the chart now accept either YAML directly or YAML-formatted strings (as of #272).

PR #257, opened 2020-04-13
Setting Ingress to use the active Vault Pod

ttyS0:
The Ingress service defaults to the `vault` service which contains all the running Pods. For some operations, this can cause a error when a `standby` Pod happens to be connected to with the following message:`Error [action description]: redirect would cause protocol downgrade`.

Since the chart creates a `vault-active` service that points to the currently active Pod, this PR adds the `-acvive` suffix to the service used by the Ingress so the active Pod is always being talked to by clients, which alleviates the above error.

### DIFF
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -36,9 +36,9 @@ spec:
         {{- range (.paths | default (list "/")) }}
           - path: {{ . }}
             backend:
-              serviceName: {{ $serviceName }}
+              serviceName: {{ $serviceName }}-active
               servicePort: {{ $servicePort }}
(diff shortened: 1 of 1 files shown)


georgekaz:
I came across this too but this solution is not good because it only creates the -active service if you're using ha. I made a change to achieve the same but with some extra logic. I'll put that PR up but I don't think this one should be accepted Sorry @ttyS0 .

georgekaz:
This PR achieves the same: https://github.com/hashicorp/vault-helm/pull/270

ttyS0:
@georgekaz has a better approach.

PR #259, opened 2020-04-14
service: fix annotations for HA standby/active services

yotsub:

PR #261, opened 2020-04-15
Fixes active and standby service annotations

kri5:
It has been decided in #227 to move annotations to multi-line
string for consistency. Thoses definitions were not following thoses
rules. This commit fixes it.

### DIFF
--- a/templates/server-ha-active-service.yaml
+++ b/templates/server-ha-active-service.yaml
@@ -14,7 +14,7 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
   annotations:
 {{- if .Values.server.service.annotations }}
-{{ toYaml .Values.server.service.annotations | indent 4 }}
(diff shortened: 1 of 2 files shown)


alaindazzi:
i installed the latest helm chart version and kept getting "Error: UPGRADE FAILED: YAML parse error on vault/templates/server-ha-standby-service.yaml: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field .metadata.annotations of type map[string]string".

I fixed it by changing the code to ...
  annotations:
{{- if .Values.server.service.annotations }}
{{- tpl .Values.server.service.annotations . | nindent 4 }}
{{- end }}

Dirc:
Seems duplicate of https://github.com/hashicorp/vault-helm/pull/259

kri5:
@Dirc You're completely right!I don't know how i missed this one!

PR #263, opened 2020-04-15
Ability to provide imagePullSecrets also for injector image

snowping:
Global helm variable "ImagePullSecrets" is not taken into account for injector image "hashicorp/vault-k8". I added the missing template config.


jasonodonnell:
Hi @snowping, thanks for the contribution!

Can you add an injector unit test to check for this secret when set so we have coverage?  You can find them in `test/unit/injector-deployment.bats`.

Let me know if you have any questions!

jasonodonnell:
Feature merged with #298.

PR #268, opened 2020-04-21
Fix ha standby and active annotations

tvoran:


kauedg:
Both files `server-ha-active-service.yaml` and `server-ha-standby-service.yaml` have the same comment:
`# Service for active Vault pod`

Maybe it should be `# Service for standby Vault pod` in `server-ha-standby-service.yaml`

tvoran:
@kauedg Yep, fixed in https://github.com/hashicorp/vault-helm/pull/299

PR #270, opened 2020-04-21
fix annotation bug. Use active service on ingress when ha

georgekaz:
Annotations currently give a warning because they are not actually rendering as yaml like normal charts do.
"Warning: Merging destination map for chart 'vault'. Cannot overwrite table item 'annotations', with non table value: map[]"

changing from annotations: {} to annotations: [] removes this warning.

Additionally I've added some logic that points the ingress at the active server when in ha mode. There are times that pointing at the standby pods causes problems.

### DIFF
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -3,7 +3,7 @@
 {{- if .Values.server.ingress.enabled -}}
 {{- $serviceName := include "vault.fullname" . -}}
 {{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
-{{- $serviceName := printf "%s-%s" $serviceName "active" -}}
+{{- $serviceName = printf "%s-%s" $serviceName "active" -}}
 {{- end }}
 {{- $servicePort := .Values.server.service.port -}}
 {{ if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -2,6 +2,9 @@
 {{- if ne .mode "external" }}
 {{- if .Values.server.ingress.enabled -}}
 {{- $serviceName := include "vault.fullname" . -}}
+{{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
+{{- $serviceName := printf "%s-%s" $serviceName "active" -}}
+{{- end }}
 {{- $servicePort := .Values.server.service.port -}}
 {{ if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
 apiVersion: networking.k8s.io/v1beta1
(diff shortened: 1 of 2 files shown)


templates/server-ingress.yaml
```diff
@@ -2,6 +2,9 @@
 {{- if ne .mode "external" }}
 {{- if .Values.server.ingress.enabled -}}
 {{- $serviceName := include "vault.fullname" . -}}
+{{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
```
jasonodonnell, line 5: This change would only allow ingress to be created for HA mode, but wouldn't support dev or standalone.  This logic will need to be expanded to support the other modes as well.

georgekaz: no it wouldn't. this is specifically appending "-active" to the serviceName defined on line 4, if and only if it's ha. That doesn't mean there won't be an ingress in the other cases, just that it won't include "-active" on the service name

jasonodonnell: Aha, I see now.  Thanks!

values.yaml
```diff
@@ -133,7 +133,7 @@ server:
     enabled: false
     labels: {}
       # traffic: external
-    annotations: {}
+    annotations: []
```
tvoran, line 136: I think we'll want to keep annotations defaulting to `{}` as described in #272.

georgekaz: I agree that using {} is better and is standard practice in most charts, and it works correctly when using toYaml on the annotations, but as this current chart expects the annotations as a multi-line string it gave the warning. The PR you mentioned came in after mine, so personally yes I would keep the {} now to allow toYaml to work and just put up with the warning.

georgekaz:
I've reverted the annotations as requested and actually corrected a mistake in the serviceName. It prints as expected using helm template and helm lint. Not sure how to automate this test on here.

templates/server-ingress.yaml
```diff
@@ -2,6 +2,9 @@
...
+{{- $serviceName := printf "%s-%s" $serviceName "active" -}}
```
tvoran, line 6: I think this needs to be `=` instead of `:=`, because serviceName is already defined above.
```suggestion
{{- $serviceName = printf "%s-%s" $serviceName "active" -}}
```

georgekaz: you're right. I originally did that but then had a problem with it showing `unexpected "=" in operand`. However, that's because I had switched to an older version of helm which appears to have a bug. I will revert this to my original version without the :

templates/server-ingress.yaml
```diff
@@ -2,6 +2,9 @@
 {{- if ne .mode "external" }}
 {{- if .Values.server.ingress.enabled -}}
 {{- $serviceName := include "vault.fullname" . -}}
+{{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
+{{- $serviceName = printf "%s-%s" $serviceName "active" -}}
```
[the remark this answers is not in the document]
georgekaz: I've added tests now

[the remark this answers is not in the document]
tvoran: Thanks!

PR #276, opened 2020-04-24
return vault_api_addr as fqdn instead of ip address

ngarafol:
It makes more sense IMHO to return FQDN as vault_api_addr instead of ip address. 

Discussion is here https://discuss.hashicorp.com/t/req-return-api-addr-as-fqdn-instead-of-ip/7896/2


jasonodonnell:
Hi @ngarafol, #290 made VAULT_API_ADDR configurable, so closing in favor of that change.

PR #277, opened 2020-04-25
Chart configuration options improvements

sergevs:


templates/server-statefulset.yaml
```diff
@@ -46,16 +46,20 @@ spec:
...
       {{ end }}
       securityContext:
-        runAsNonRoot: true
+        runAsNonRoot: {{ .Values.server.runAsNonRoot | default true }}
         runAsGroup: {{ .Values.server.gid | default 1000 }}
         runAsUser: {{ .Values.server.uid | default 100 }}
         fsGroup: {{ .Values.server.gid | default 1000 }}
+      initContainers:
+   {{- if .Values.server.customInitContainers }}
+{{ tpl .Values.server.customInitContainers . | indent 6 }}
```
tvoran, line 55: I think this could support both YAML and a YAML-formatted string, similar to the implementation in #272. And there should be a couple tests for it.

jasonodonnell:
Closing this PR due to many of these features already existing.

PR #283, opened 2020-04-30
Additional configuration options for service

lawliet89:
- Clarified services being created in values
- Added `LoadBalancer` options
- Added option to only select active Vault pod for service


lawliet89:
Closing this since I realised that the service is used for joining raft nodes. I guess these are more suited for the `ui` service.

PR #284, opened 2020-05-02
security: refactor gid and uid

moray95:
- Moves defaults for `server.gid` and `server.uid` to `values.yaml`.
- Adds missing documentation for the fields.

Motivation:

Currently, it's not possible to have an "empty" group or user id in pods. Overriding `server.gid` or `server.uid` to `null` doesn't work because of the `default` clause. However, on Openshift the pods will be assigned a proper group and user depending on security context constraints that applies to it. 

From [Managing Security Context Constraints](https://docs.openshift.com/container-platform/4.3/authentication/managing-security-context-constraints.html):
> When a container or pod does not request a user ID under which it should be run, the effective UID depends on the SCC that emits this pod. Because restricted SCC is granted to all authenticated users by default, it will be available to all users and service accounts and used in most cases.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -127,6 +127,11 @@ server:
   #     memory: 256Mi
   #     cpu: 250m
 
+  # Pod's group id.
+  gid: 1000
+  # Pod's user id.
+  uid: 100
+
   # Ingress allows ingress services to be created to allow external access
   # from Kubernetes to access Vault pods.
   ingress:
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -44,9 +44,9 @@ spec:
       {{ end }}
       securityContext:
         runAsNonRoot: true
-        runAsGroup: {{ .Values.server.gid | default 1000 }}
-        runAsUser: {{ .Values.server.uid | default 100 }}
-        fsGroup: {{ .Values.server.gid | default 1000 }}
+        runAsGroup: {{ .Values.server.gid }}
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -127,6 +127,11 @@ server:
   #     memory: 256Mi
   #     cpu: 250m
 
+  # Pod's group id.
+  gid: 1000
+  # Pod's user id.
+  uid: 1000
```
jasonodonnell, line 133: This should be `100`.  The servers UID and GID by default are `100:1000`.

moray95: Fixed.

jasonodonnell:
@moray95 I think we'll want unit tests covering the `null` use case.  Have you tested this on a live system where the values are null?  I'm wondering if we need to wrap the security contexts with if statements to exclude these values if they aren't set to an integer value?

moray95:
@jasonodonnell Thanks for the feedback. I'll be writing the tests asap. I have tested a deployment with null values on Openshift 4.3 and worked just fine. I don't think verifying the types is very common on Helm charts, but could be added. TBH I would prefer taking the whole securityContext from values instead of just the user and group ids, as done on many other charts. However, I didn't attempt such a change since it would break backwards compatibility. I could change the PR to use this method if you would be interested.

moray95:
@jasonodonnell Tests for the `null` case have been added.

jasonodonnell:
This functionality expanded into more OpenShift support in #319, closing for duplication.

PR #285, opened 2020-05-04
Add configuration options for Vault UI service

lawliet89:
- Configure to select active Vault pod only
- Configure to not publish unready address

### DIFF
--- a/templates/ui-service.yaml
+++ b/templates/ui-service.yaml
@@ -18,7 +18,7 @@ spec:
     app.kubernetes.io/name: {{ include "vault.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     component: server
-    {{- if .Values.ui.activeVaultPodOnly }}
+    {{- if and (.Values.ui.activeVaultPodOnly) (eq .mode "ha") }}
     vault-active: "true"
     {{- end }}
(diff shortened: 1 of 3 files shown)


templates/ui-service.yaml
```diff
@@ -23,7 +18,10 @@ spec:
     app.kubernetes.io/name: {{ include "vault.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     component: server
-  publishNotReadyAddresses: true
+    {{- if .Values.ui.activeVaultPodOnly }}
```
jasonodonnell, line 21: This is only true if in HA mode. I think we don't even need a configurable here.  If it's HA mode, just set the selector criteria, else point to the single deployed pod.

Thoughts?

lawliet89: Sure, I will make the change.

lawliet89: I can add an additional condition to check that HA mode is on.

I think this should be configurable to handle other possible use cases like https://github.com/hashicorp/vault-helm/issues/326. What do you think? I could remove this option if you think otherwise.

jasonodonnell: I think configurable is fine.

lawliet89: I've added the check for HA in 956b101

PR #286, opened 2020-05-05
Run acceptance tests in CI.

sarahethompson:
Run acceptance tests in CircleCI (with minikube).


sarahethompson:
Going to implement with GKE instead of minikube.

PR #289, opened 2020-05-05
[Proposal] Added flag for OpenShift based deployment

radudd:
This is a proposal to make these Helm charts compatible with an OpenShift 4 deployment.
An openshift flag has been added to values.yaml file that is expected to be true in case the deployment is for OpenShift. 

If the flag is set, the deployment performs the following actions:
* creates a Route of type reencrypt instead of Ingress if exposing Vault if openshift flag is set
* removes the security context from server statefulset in order for vault-agent pods to run as the same user as application pod. Relies on https://github.com/hashicorp/vault-k8s/pull/126 
* add Network Policies for both Server and Injector in order to be reachable from outside their namespaces
* signing Services with OpenShift server-ca and mount TLS keys and certificates to vault pods. https://docs.openshift.com/container-platform/4.4/authentication/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate

Additionally:
* Injector is using openlabred/vault-k8s image instead of hashicorp/vault-k8s pending https://github.com/hashicorp/vault-k8s/pull/126
* disable TLS and provide TLS key and certificate file location to Vault config if deploying to OpenShift - this is not done through the flag, but directly set in values.yaml file

The chart has been tested on OpenShift 4.3 and 4.4 both as Standalone or HA deployment.


templates/server-network-policy.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.openshift }}
```
pabrahamsson, line 1: Same as above, seems like something for vanilla k8s as well.

templates/server-ingress.yaml
```diff
@@ -1,3 +1,5 @@
+## In case of OpenShift deployment, a route will be deployed instead of ingress
+{{- if not .Values.openshift }}
```
pabrahamsson, line 2: Is this necessary considering the default: `server.ingress.enabled: false`?

radudd: My idea was to completely ignore server.ingress block if the target deployment platform is OpenShift. Same as ignoring server.route if deployment is on vanilla K8s. (in order to have same concept for configuring those two).
Another reason for this was that OpenShift 4 supports Ingress objects and if one is defined it will be converted to a Route (initially, I was also thinking to use ingress instead of routes, however they doesn't offer the same level of config flexibility). Without this condition in ingress object and  in case deployment is on OpenShift, the user will need to pay extra attention to disable server.ingress in values.yaml, besides enabling routes, otherwise he will get two routes created and the one converted from the ingress will not work.
But if you think it's a better idea to delegate this responsibility to the user, then we can implement this way also.

templates/server-network-policy.yaml
radudd: Same as previous comment. I didn't want to impact with this PR the way how Vault is deploy to vanilla K8S.

templates/injector-network-policy.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.openshift }}
```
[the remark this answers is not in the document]
radudd: This can apply to vanilla Kubernetes indeed, however I didn't want to impact with this PR the way how Vault is currently deployed on vanilla k8s.
Regarding the the second suggestion, yes you are right. I will update this accordingly

values.yaml
```diff
@@ -11,6 +11,9 @@ global:
   # TLS for end-to-end encrypted transport
   tlsDisable: true
 
+# If deploying to OpenShift
+openshift: false
```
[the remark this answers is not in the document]
radudd: Thanks, I will update this.

values.yaml
```diff
@@ -20,9 +23,10 @@ injector:
   externalVaultAddr: ""
 
   # image sets the repo and tag of the vault-k8s image to use for the injector.
+  # image points to openlabred/vault-k8s repository, pending https://github.com/hashicorp/vault-k8s/pull/126
   image:
-    repository: "hashicorp/vault-k8s"
-    tag: "0.3.0"
+    repository: "openlabred/vault-k8s"
```
jasonodonnell, line 28: This needs to be `hashicorp`

values.yaml
```diff
@@ -335,6 +354,22 @@ server:
...
+
+    #  listener "tcp" {
+    #    address = "[::]:8200"
+    #    cluster_address = "[::]:8201"
+    #    tls_cert_file = "/var/run/secrets/kubernetes.io/certs/tls.crt"
+    #    tls_key_file = "/var/run/secrets/kubernetes.io/certs/tls.key"
+    #  }
+    #  storage "file" {
+    #    path = "/vault/data"
+    #  }
```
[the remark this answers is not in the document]
jasonodonnell: @pabrahamsson let's just keep what we have and it can be added to the documentation/comments.  We're trying to not add even more configs.

values.yaml
```diff
@@ -396,6 +434,24 @@ server:
...
+    #    cluster_address = "[::]:8201"
+    #    tls_cert_file = "/var/run/secrets/kubernetes.io/certs/tls.crt"
+    #    tls_key_file = "/var/run/secrets/kubernetes.io/certs/tls.key"
+    #  }
+    #  storage "consul" {
+    #    path = "vault"
+    #    address = "HOST_IP:8500"
+    #  }
+    #  service_registration "kubernetes" {}
+
```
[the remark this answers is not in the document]
jasonodonnell: @pabrahamsson let's just keep what we have and it can be added to the documentation/comments. We're trying to not add even more configs.

values.yaml
eye0fra: I guess will be tag: 0.4.0 from hashicorp right?

jasonodonnell: We'll update this on our end for the next release.

values.yaml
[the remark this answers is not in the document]
jasonodonnell: This still needs updating to `global.openshift`

values.yaml
```diff
@@ -152,6 +158,16 @@ server:
     #    hosts:
     #      - chart-example.local
 
+  # OpenShift only - create a route to expose the service
+  # The created route will be of type reencrypt
+  # https://docs.openshift.com/container-platform/4.4/networking/routes/secured-routes.html#nw-ingress-creating-a-reencrypt-route-with-a-custom-certificate_secured-routes
+  # Services are secured by OpenShift "service-ca" controller
+  # https://docs.openshift.com/container-platform/4.4/authentication/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate
+  route:
+    enabled: true
```
jasonodonnell, line 167: Route should be disabled by default.  Users should explicitly toggle exposing a service.

values.yaml
```diff
@@ -152,6 +158,16 @@ server:
...
+    labels: {}
+    annotations: {}
+    host: chart-example.local
```
jasonodonnell, line 170: We might make a better default host here.  Maybe use the chart name?

values.yaml
radudd: I fixed this with the latest commits

values.yaml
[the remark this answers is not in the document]
radudd: Fixed with latest commits

templates/server-network-policy.yaml
```diff
@@ -0,0 +1,20 @@
...
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: {{ template "vault.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  ingress:
+    - from:
+        - namespaceSelector: {}
+      ports:
+      - port: 8200
+        protocol: TCP
```
jasonodonnell, line 19: 8201 also needs to be added here.

templates/server-statefulset.yaml
```diff
@@ -64,6 +72,8 @@ spec:
           command: {{ template "vault.command" . }}
           args: {{ template "vault.args" . }}
           env:
+            - name: VAULT_LOG_LEVEL
+              value: debug
```
jasonodonnell, line 76: This needs to be removed.

radudd: Thanks, fixed with latest commit. This was actually the reason why the previous unit tests were failing(the ones that were checking environment variables based on indexes)

radudd:
Thanks a lot @tvoran , @jasonodonnell , @pabrahamsson , @eye0fra for reviewing the PR and for your feedback. 
With the latest commits that I pushed, I fixed/added the following:
- added unit tests for OpenShift
- added openshift flag in the global block in configuration
- added condition if injector enabled in the injector NetworkPolicy
- disabled creation of route by default
- removed debug log level

values.yaml
radudd: I used the same naming example as for the ingress, but I can update it

templates/server-network-policy.yaml
jasonodonnell: Bump on this, definitely needs to be in else it breaks Raft storage.

radudd:
If using RAFT integrated storage for HA deployment, the mechanism with the reencrypt routes will not work out-of-the-box.
This because Vault pods are using the TLS certificates generated for the main Vault service for load balancing (_<releaseName>-vault_.<namespace>.svc); but when a node will try to join the RAFT cluster, it will need to call the leader node using the Headless service (i.e. _vault operator raft join <releaseName>-vault-<index>.<releaseName>-internal_.<namespace>.svc). Hence, the certificates validation will fail and the node won't be able to join the cluster.

However, I tested out two possible solutions for deploying Vault with RAFT integrated storage in OpenShift:
1. Don't use TLS for the Vault listener and use Edge routes. The disadvantage of that would be TLS will terminate at Router level, then the traffic will be unencrypted.
2. Create a secondary non-TLS listener for Vault and use this only when the nodes will join RAFT cluster. The main listener will still use TLS and we will be able to create reencrypt routes.

@jasonodonnell, @tvoran let me know your thoughts. Thanks!

jasonodonnell:
@radudd I'm unsure why the route is messing with the internal headless service the Raft nodes are using to join the cluster.  There might be something else going on here (such as CLUSTER_ADDR or API_ADDR?)

I'm wondering if we should have reencrypt sets to `passthrough`?  Communications should stay encrypted until they reach Vault.

Thoughts?

radudd:
Thanks for feedback @jasonodonnell . There are two points here:
1. Node joining existing Raft cluster. For this no route is used, but the headless service. The requester pod will try to join the existing pod in the cluster and it will call it using the headless service **release-vault-internal**.namespace.svc`.
However, Vault listener on pods is configured with the SSL certificates matching the main Vault service: **release-vault**.namespace.svc (the certificates created by OpenShift CA using annotations). 
Hence the certificate validation will fail when the pod will try to join the cluster and the operation will fail. 
CLUSTER_ADDR is using port 8201 and Vault creates an SSL certificate for it, but this can't be used when a node wants to join a cluster, no ?
I think the situation is similar if using Vault with TLS and enabled ingress on vanilla K8s with Raft storage. How would be the recommended configuration in this case ?

2. Route type. Re-encrypt routes terminates TLS at router level but then the router creates another SSL encrypted connection to Vault main service. If we would use passthrough Route instead, the router would just forward the connection to Vault service, but in this case the TLS certificates on Vault pods will not match the route FQDN (so we would need to manage separately the TLS certificates for the route)
I would use re-encrypt routes as we can rely on both OpenShift CA to create TLS certificates https://docs.openshift.com/container-platform/4.4/authentication/certificates/service-serving-certificate.html and on router wildcard certificate.

jasonodonnell:
@radudd,

Thanks for the details!  I see what you mean now but I'm curious, does `vault operator raft join leader-ca-cert=@ca.crt` not help in this situation?

radudd:
Thanks @jasonodonnell for the reply. No, it doesn't help as the certificates are issued for a different service, the trusted CA shouldn't be an issue.

eye0fra:
It should be service-ca.crt in this case

jasonodonnell:
@radudd @eye0fra I tested joining raft by having OpenShift add the CA Bundle to a configmap, then I specified the following on the command line:

```
vault operator raft join \
  -tls-server-name=vault.demo.svc \
  -ca-path=/var/run/secrets/kubernetes.io/ca/service-ca.crt \
  https://vault-0.vault-internal:8200
```

That being said, I'd like to pull the route stuff from this PR so we can merge it.  We can do a follow up PR with route integrations.  Thoughts?

radudd:
@jasonodonnell fine with me. Though this particular problem is not route related, but more on how can we add nodes to raft cluster when Vault pods are using TLS (using the same certificate which match main vault service and not individual ones).
Another approach would be to leave Raft configuration for OpenShift out of this release and create a PR for this later. 
This auto-signing TLS feature is supported since a while in OpenShift, so I would still keep it for creating certificates for Vault TLS listeners.

But regarding the routes, I have actually another small issue, i.e. I see the following message constantly in the Vault pods logs. Thoughts ? 
`http2: server: error reading preface from client router-ip:36068: read tcp vault-ip:8200->router-ip:36068: read: connection reset by peer`
However, this doesn't seem to have any impact with the functionality.
Thougths?

jasonodonnell:
@radudd I forked this PR and am making tweaks to it (PR coming very soon).  We want to get this merged ASAP so we can begin doing a beta for OpenShift support.  That being said I'm keeping route enabled but switched it to `passthrough` for now.  

Additionally I removed the auto-tls feature and for now we will require certificates be added to the deployment.  This isn't ideal since OpenShift has a nice feature to do this for you, however, this isn't different from vanilla K8s deployments.  We definitely want to tap into the power of this auto-tls feature, but right now it's blocking us from moving forward with a small OpenShift beta period.  We can't really punt on Raft support as it's crucial to an isolated, HA cluster deployment.  We will get this feature in ASAP in a separate PR.

eye0fra:
Hi @jasonodonnell, I perfectly agree, thanks for pushing forward.

radudd:
Great @jasonodonnell. Thanks a lot!

Issue #302, opened 2020-05-16
Support placing plugins on the server


allandegnan:
My understanding is vault needs custom plugins to exist on the filesystem at _some_ path but I can't see an obvious way of achieving this via the current chart.

Two methods I've seen are:

1. Stuff binaries in a configMap. 
2. Support emptyDir for extraVolumes.

I've written a PR for option 2.

jasonodonnell:
@allandegnan Binaries definitely shouldn't go into configMaps.  I'm curious how you will populate the emptyDir with the binary?  Using an init container to fetch it?

allandegnan:
@jasonodonnell - Yep, pretty much.  cp from the initContainer.

Save us from having to build & track your (hashi) containers.

jasonodonnell:
#314 makes this possible now.  Closing.

timclifford:
Is there not a better/new way to load custom plugins into vault other than the above? I'm using the second approach but was wondering if custom plugins were better supported now

PR #303, opened 2020-05-16
Added support for emptyDir extraVolumes

allan-degnan-rft:
Should resolve issue #302

### DIFF
--- a/test/unit/server-statefulset.bats
+++ b/test/unit/server-statefulset.bats
@@ -244,6 +244,10 @@ load _helpers
       yq -r '.configMap.secretName' | tee /dev/stderr)
   [ "${actual}" = "null" ]
 
+  local actual=$(echo $object |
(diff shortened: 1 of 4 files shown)


test/unit/server-statefulset.bats
```diff
@@ -347,7 +363,7 @@ load _helpers
   local object=$(helm template \
       --show-only templates/server-statefulset.yaml  \
       --set 'server.standalone.enabled=true' \
-      --set 'server.extraVolumes[0].type=configMap' \
+      --set 'server.extraVolumes[0].type=secret' \
```
allan-degnan-rft, line 366: This change and the other 3 like it are intentional as I believe they're duplicate tests and this was the original intention.

If my PR gets rejected, fixing the tests might still be worthwhile.

jasonodonnell:
#314 was merged, so closing this in favor.

PR #314, opened 2020-05-27
add support for more flexible volume mounts e.g. for use with init containers

georgekaz:
The existing volume mounts implementation via extraVolumes is too opinionated to allow mounting volumes for custom purposes, such as installing a plugin via an init container. Personally I don't like it and would remove it completely, but for backwards compatibility I've left it and added additional server.volumes and server.volumeMounts. Examples in the values files show how this can be used.

Please consider this as it's an easy way to customise your Vault installation.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -97,6 +97,9 @@ extra volumes the user may have specified (such as a secret with TLS).
             secretName: {{ .name }}
           {{- end }}
   {{- end }}
+  {{- if .Values.server.volumes }}
+    {{- toYaml .Values.server.volumes | nindent 8}}
+  {{- end }}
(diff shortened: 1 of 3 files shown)


georgekaz:
Hi Jason. I'll try, I've not used these before so might have questions. I've tested the output is as expected using `helm template` and `helm lint` and am successfully running the output in a cluster at the moment. Happy to learn the unit tests though, will be useful for me.

allan-degnan-rft:
If this gets merged you can kill the need for https://github.com/hashicorp/vault-helm/pull/303 and https://github.com/hashicorp/vault-helm/issues/302.

sssd-dev:
@jasonodonnell Can this be merged, please?

PR #315, opened 2020-05-28
Add postStart lifecycle hook

jasonodonnell:
This adds a configurable `postStart` lifecycle hook to Vault Helm.  Using this hook, in conjunction with readiness probe configurations, it's possible to run scripts on the Vault pods after they start.  This would allow users to create custom scripts to do things like initialize Vault and bootstrap auth methods/policies.

```yaml
server:
  # Mount a secret containing some custom script
  extraVolumes:
  - type: secret
    name: demo-vault

  # Configure readiness probe to report "ready" even if Vault is uninitialized and sealed.
  # Useful when bootstrapping Vault for the first time.
  readinessProbe:
    path: "/v1/sys/health?standbyok=true&sealedcode=204&uninitcode=204"

  postStart:
  - "/bin/sh"
  - "-c"
  - "/vault/userconfig/demo-vault/bootstrap.sh"
```

These can also be configured via `--set`:

```bash
helm install vault \
  --set='server.postStart={/bin/sh,-c,/vault/userconfig/demo-vault/bootstrap.sh}' \
  .
```

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -191,7 +191,7 @@ server:
   # Used to define commands to run after the pod is ready.
   # This can be used to automate processes such as initialization
   # or boostrapping auth methods.
-  postStart: |
+  postStart: []
   # - /bin/sh
   # - -c
   # - /vault/userconfig/myscript/run.sh
--- a/values.yaml
+++ b/values.yaml
@@ -188,6 +188,14 @@ server:
   # Used to set the sleep time during the preStop step
   preStopSleepSeconds: 5
 
+  # Used to define commands to run after the pod is ready.
+  # This can be used to automate processes such as initialization
+  # or boostrapping auth methods.
+  postStart: |
+  # - /bin/sh
+  # - -c
+  # - /vault/userconfig/myscript/run.sh
+
   # extraEnvironmentVars is a list of extra enviroment variables to set with the stateful set. These could be
   # used to include variables required for auto-unseal.
   extraEnvironmentVars: {}
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -159,6 +159,14 @@ spec:
                   # to this pod while it's terminating
                   "sleep {{ .Values.server.preStopSleepSeconds }} && kill -SIGTERM $(pidof vault)",
                 ]
+            {{- if .Values.server.postStart }}
+            postStart:
+              exec:
+                command:
+                {{- range (.Values.server.postStart) }}
+                - {{ . | quote }}
+                {{- end }}
+            {{- end }}
         {{- if .Values.server.extraContainers }}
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -188,6 +188,14 @@ server:
   # Used to set the sleep time during the preStop step
   preStopSleepSeconds: 5
 
+  # Used to define commands to run after the pod is ready.
+  # This can be used to automate processes such as initialization
+  # or boostrapping auth methods.
+  postStart: |
```
tvoran, line 194: Should this default to an empty list?
```suggestion
  postStart: []
```

davidcunningham:
This is a great feature, will more documentation be provided? For example, I have a ConfigMap that creates a script to set a bunch of default auth policies. The script is written to userconfig properly using extraVolumes, but the permissions are set to root so the postStart can not execute the script as permission is denied. Any additional examples would be awesome!

jasonodonnell:
Hi @davidcunningham,

Something like this should work for you (it's a little ugly):

```yaml
server:
  extraVolumes:
  - type: secret
    name: demo-vault
  readinessProbe:
    path: "/v1/sys/health?standbyok=true&sealedcode=204&uninitcode=204"

  postStart:
  - "/bin/sh"
  - "-c"
  - "sleep 5 && cp /vault/userconfig/demo-vault/bootstrap.sh /tmp/bootstrap.sh && chmod +x /tmp/bootstrap.sh && /tmp/bootstrap.sh"
```

We need to add a feature that will allow users to set the mode for the files being mounted.  That should make this nicer.

davidcunningham:
Thanks @jasonodonnell. It's a prettier solution then having a bunch of commands in the postStart!

angeloskaltsikis:
Hey @jasonodonnell , thanks for your great work on this feature.
Currently, we are introducing Vault & also we want to use Vault Terraform Provider to configure some of the required Vault configurations.
To avoid storing plain credentials (or even encrypted ones) in our IaC we decided to utilize AWS Auth method for Vault which fits perfectly for our case and verified that that's true.
While we would like to auto initialize the AWS Auth method with the proper configuration when we install Vault and avoid any need for manual human interaction, we are a bit concerned that introducing something like [this](https://github.com/jasonodonnell/vault-agent-demo/blob/master/configs/bootstrap.sh) has some security implications (storing the root token in a plaintext file in Vault container). 
- Do you think that we are right to be worried about production deployment?
- Can you think or have you found any better alternative?
Thanks in advance 🙏🏽

andreapigatto:
I'm trying to find a way to run a script for enable transit for auto-unseal. `postStart` seems good candidate but in HA I would need ONLY one node to actually run the script and enable transit. Is there a way to do it with `postStart` it or is there a better approach? Thanks!

PR #316, opened 2020-05-28
check that git tag == chart tag on tagged releases

alvin-huang:
It is easy to miss updating the version number in `Chart.yaml` when tagging a release so this is a sanity check to make sure the git tag version matches the version in the `Chart.yaml` file. I have also added a slack notification to send something when there is a failure in the tagged build workflow.


sarahethompson:
👍

PR #317, opened 2020-05-29
Allow setting HA services type

tvoran:
Making the types for active and standby services configurable (just
like the main vault service).

When ha and ingress are enabled in the chart, Ingress uses the `vault-active` service, which is currently a ClusterIP service. The trouble is, Ingress on GKE doesn't support ClusterIP services, so the service type needs to be configurable to work there.

### DIFF
--- a/templates/server-ha-active-service.yaml
+++ b/templates/server-ha-active-service.yaml
@@ -15,13 +15,21 @@ metadata:
   annotations:
 {{ template "vault.service.annotations" .}}
 spec:
-  type: ClusterIP
+  {{- if .Values.server.service.type}}
+  type: {{ .Values.server.service.type }}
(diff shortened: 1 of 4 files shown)


iusergii:
@tvoran, @jasonodonnell  Doesn't this force us to use a`service_registration` with Ingress?

tvoran:
Yes, if you're running with `server.ha.enabled=true`, you'll need to have `service_registration: {}` present in `server.ha.config` or `server.ha.raft.config`.

iusergii:
I'm using DynamoDB as storage and seems `service_registration` has a bug for it. Filed - https://github.com/hashicorp/vault/issues/9153 Can upgrade to 0.6.0 due to this.

PR #319, opened 2020-06-03
Add OpenShift beta support

jasonodonnell:
This adds **beta support for OpenShift Container Platform** and is a fork of #289.

Added in this PR is:
- `global.openshift` value to make the helm chart compatible with OpenShift 4.x
- Added an `emptydir` volume to `/home/vault` so Vault CLI can cache tokens for bootstrapping (permission issues on OpenShift where `/home/vault` isn't owned by arbitrary UID)
- Environment variable for Vault K8s to enable OpenShift support (if enabled)
- OpenShift route support (currently set to `passthrough` to allow Vault to handle TLS)
- Networking policies for OpenShift

This PR differed from #289 by removing auto-tls generation offered by OpenShift.  This feature will need additional work and should be a separate PR to avoid complications with setup of integrated storage (Raft).

Both Vault OSS and Vault Enterprise are supported using this feature.


templates/server-route.yaml
```diff
@@ -0,0 +1,29 @@
...
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{- with .Values.server.route.labels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- template "vault.route.annotations" . }}
+spec:
+  host: {{ .Values.server.route.host }}
+  to:
+    kind: Service
+    name: {{ template "vault.fullname" . }}
```
tvoran, line 21: If `server.ha.enabled=true`, I wonder if this should be pointing at the `active` service (like how non-openshift ingress does now).

jasonodonnell: Added!

PR #323, opened 2020-06-04
Add note to config about sensitive configs

jasonodonnell:
Configs are stored in ConfigMaps due to limitations with Helm templating and Kube secrets requiring to be base64 encoded.  Vault Helm supports both `extraSecretEnvironmentVariables` and mounting secrets to be used as partial configuration files.  I added a note to `values.yaml` pointing to the documentation.

Additionally I added a warning header in the README about disclosing potential security flaws responsibly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,8 @@
 # Vault Helm Chart
 
----
-**Please note**: We take Vault's security and our users' trust very seriously. If 
+> :warning: **Please note**: We take Vault's security and our users' trust very seriously. If 
 you believe you have found a security issue in Vault Helm, _please responsibly disclose_ 
 by contacting us at [security@hashicorp.com](mailto:security@hashicorp.com).
----
 
 This repository contains the official HashiCorp Helm chart for installing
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,5 +1,11 @@
 # Vault Helm Chart
 
+---
+**Please note**: We take Vault's security and our users' trust very seriously. If 
+you believe you have found a security issue in Vault Helm, _please responsibly disclose_ 
+by contacting us at [security@hashicorp.com](mailto:security@hashicorp.com).
+---
```
tvoran, line 7: I think there should be a space between line 6 and 7, otherwise the entire paragraph is big and bold like a header:

image

I'd also suggest something like this:
```suggestion
> :warning: **Please note**: We take Vault's security and our users' trust very seriously. If 
you believe you have found a security issue in Vault Helm, _please responsibly disclose_ 
by contacting us at [security@hashicorp.com](mailto:security@hashicorp.com).
```

image

PR #340, opened 2020-06-25
Added support for kubernetes ManagedCertificate resource

guyzyl:
This is needed for the creation of a valid and managed SSL cert through a Kubernetes resource.
My use case is over GCP, I annotate it to the GCP ingress controller (as can see documented [here](https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs#config-connector))


jasonodonnell:
Thanks for the PR @guyzyl, we spoke about this internally and decided we'd rather not have cloud specific CRDs in the project. Instead operators should create these post install for their environment.

PR #353, opened 2020-07-10
feat(preCommands):  Add ability for helm chart to run some arbitrary …

rayisbadat:
This extends the existing vault.args in _helpers.tpl so that users can insert their own commands before it runs the two sed statements and the docker-entrypoint.sh.


rayisbadat:
Sorry for delay in getting back to you  @jasonodonnell .  I wrote a test for the case of setting it, but wasn't sure how to test for the null case.  Since i am inserting it into an existing section, i didn't want to hard code it to look for those sed command.

Issue #356, opened 2020-07-17
Var declaration syntax error in the server-route and server-ingress


gkarthiks:
**Describe the bug**
The templates for the server-route and server-ingress has the variable serviceName declaration in the lines 6 and 7 correspondingly. They are just equated with "=" sign rather than ":=".

**To Reproduce**
1. Do a helm install command on the vault chart
2. Below error would appear
```Error: parse error in "vault/templates/server-ingress.yaml": template: vault/templates/server-ingress.yaml:7: unexpected "=" in operand```
3. After fixing this same error will appear for the server-route

**Expected behavior**
There should be any syntax errors and the ingress or route should deploy without issue accordingly

**Environment**
* Kubernetes version: 1.11
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): OpenShift
  * Other configuration options or runtime services (istio, etc.): -n/a-
* vault-helm version: 0.6.0

Chart values:

Same default values

**Additional context**
-n/a-

jasonodonnell:
Closing due to Helm version error.

PR #357, opened 2020-07-17
Fix for the var declaration in server-ingress and server-route templates

gkarthiks:
This PR is to fix the templates issue in the server-ingress and server-route with the syntax issue.

Refers: https://github.com/hashicorp/vault-helm/issues/356

### DIFF
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -4,7 +4,7 @@
 {{- if .Values.server.ingress.enabled -}}
 {{- $serviceName := include "vault.fullname" . -}}
 {{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
-{{- $serviceName = printf "%s-%s" $serviceName "active" -}}
+{{- $serviceName := printf "%s-%s" $serviceName "active" -}}
 {{- end }}
 {{- $servicePort := .Values.server.service.port -}}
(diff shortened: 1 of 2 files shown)


jasonodonnell:
Hi @gkarthiks, thanks for the contribution!

One of the unit tests broke in this PR:

```
not ok 207 server/route: OpenShift - route points to active service by when HA
# (in test file test/unit/server-route.bats, line 115)
#   `[ "${actual}" = "RELEASE-NAME-vault-active" ]' failed
# RELEASE-NAME-vault
# /templates/server-route.yaml
# kind: Route
# apiVersion: route.openshift.io/v1
# metadata:
#   name: RELEASE-NAME-vault
#   labels:
#     helm.sh/chart: vault-0.6.0
#     app.kubernetes.io/name: vault
#     app.kubernetes.io/instance: RELEASE-NAME
#     app.kubernetes.io/managed-by: Helm
# spec:
#   host: chart-example.local
#   to:
#     kind: Service
#     name: RELEASE-NAME-vault
#     weight: 100
#   port:
#     targetPort: 8200
#   tls:
#     termination: passthrough
```

Can you take a look at this test @gkarthiks ?

tvoran:
Hi @gkarthiks, what version of helm are you using? I think this was an issue that is fixed in newer versions of helm, though I don't recall the exact version numbers: https://github.com/hashicorp/vault-helm/pull/270#discussion_r431285618

gkarthiks:
@tvoran Ahh. I guess the it was because I was using the helm 2.x version?

tvoran:
@gkarthiks That's probably it! This chart only supports helm 3.

PR #358, opened 2020-07-17
Missing enabled checks for injector-network-policy

delamart:
Looks like the `templates/injector-network-policy.yaml` is missing the checks to disable the template if injector isn't enabled.


delamart:
Hi, can anyone have a look ? Should be a straightforward fix. I currently use the openshift flag but don't use the injector. So I have to manually remove the extra YAML generated.

@jasonodonnell ?

PR #364, opened 2020-07-28
Add volume claim annotations

echoboomer:
Context: When using a tool like ArgoCD, `PersistentVolumeClaim` resources show up as "needs pruning" since they are generated via templating instead of factored in at build time. Despite the use of values that indicate the creation of these claims is a desirable outcome, they simply show up as "out of sync."

Adding the ability to pass annotations into these objects would allow the use of annotations to tell ArgoCD to simply ignore them. This is admittedly a fringe use case, but could provide useful either way. These were created separately to allow turning either of them on or off, but the initial approach was to just put them under `global`, but that seemed wrong.

Unit tests pass: `303 tests, 0 failures`, however no net new unit tests have been added as part of this PR. I'm happy to write one if it's necessary for this addition.

Thanks for your time!

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -179,7 +179,7 @@ storage might be desired by the user.
       {{- if and (eq (.Values.server.dataStorage.enabled | toString) "true") (or (eq .mode "standalone") (eq (.Values.server.ha.raft.enabled | toString ) "true" )) }}
     - metadata:
         name: data
-      {{- template "vault.dataVolumeClaim.annotations" . }}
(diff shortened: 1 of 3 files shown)


echoboomer:
@jasonodonnell Thanks for the feedback! I have added unit tests to cover both cases regarding `typeOf` for either `string` or `yaml`:

```bash
$ bats ./test/unit/server-statefulset.bats
...
69 tests, 0 failures
```

I hope this follows the formatting expected. Please let me know if there are any further required changes. Thank you!

echoboomer:
Hey @jasonodonnell - just circling back to see if I need to add anything else here. Thank you!

PR #338, opened 2020-06-23
Make serviceAccount name a configuration option

davidholsgrove:
**_Note this PR is a clone of original https://github.com/hashicorp/vault-helm/pull/56 authored by @satterly to change the source branch to recently rebased fork. Original PR description is;_**

Allow user to set the Kubernetes serviceAccount name used by vault.

Currently the serviceAccount name is based on the chart release name
which is non-deterministic and so makes using GCloud Workload Identity
service account mapping very difficult.

And according to Google, "Workload Identity is the recommended way to access Google Cloud services from within GKE due to its improved security properties and manageability."
https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

This also follows Helm Best Practices when defining serviceAccount names
https://helm.sh/docs/chart_best_practices/#using-rbac-resources

Example values.yaml

server:
  serviceaccount:
    name: vault-ksa
    annotations:
      iam.gke.io/gcp-service-account: vault-gsa@my-project.iam.gserviceaccount.com

where

    "vault-ksa" is the Kubernetes Service Account
    "vault-gsa" is the GCP Service Account
    "my-project" is the GCP Project name

PR #367, opened 2020-07-31
Make serviceAccount name a configuration option

burningalchemist:
This PR is a continuation of #338 with fixed tests to finally bring the feature to the chart.

### DIFF
--- a/test/unit/server-serviceaccount.bats
+++ b/test/unit/server-serviceaccount.bats
@@ -2,6 +2,41 @@
 
 load _helpers
 
+@test "server/ServiceAccount: specify service account name" {
+  cd `chart_dir`
+
+  local actual=$(helm template \
+      -x templates/server-serviceaccount.yaml  \
+      --set 'server.dev.enabled=true' \
+      --set 'server.serviceAccount.enabled=false' \
+      . | tee /dev/stderr |
+      yq 'length == 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(helm template \
+      -x templates/server-serviceaccount.yaml  \
+      --set 'server.dev.enabled=true' \
+      --set 'server.serviceAccount.name=user-defined-ksa' \
+      . | tee /dev/stderr |
+      yq -r '.metadata.name' | tee /dev/stderr)
+  [ "${actual}" = "user-defined-ksa" ]
+
+  local actual=$(helm template \
+      -x templates/server-serviceaccount.yaml  \
+      --set 'server.dev.enabled=true' \
+      . | tee /dev/stderr |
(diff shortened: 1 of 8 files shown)


burningalchemist:
@jasonodonnell I've fixed the tests, but CLA on @davidholsgrove side seems to be a blocker now. 😢

davidholsgrove:
Apologies, working through approvals to sign the CLA at the moment. Happy for anyone to take over and get this merged if that's an option too.
Thanks!

values.yaml
```diff
@@ -451,6 +451,11 @@ server:
 
   # Definition of the serviceAccount used to run Vault.
   serviceAccount:
+    # Specifies whether a service account should be created
+    enabled: true
```
tvoran, line 455: I wonder if `create` would be a better directive here, since currently `name` can be set in this block regardless.

test/unit/server-serviceaccount.bats
```diff
@@ -2,6 +2,41 @@
...
+      yq -r '.metadata.name' | tee /dev/stderr)
+  [ "${actual}" = "RELEASE-NAME-vault" ]
+
+  local actual=$(helm template \
+      --show-only templates/server-serviceaccount.yaml  \
+      --set 'server.ha.enabled=true' \
+      --set 'server.serviceAccount.annotations.iam\.gke\.io/gcp-service-account=user-defined-gsa@my-project.iam.gserviceaccount.com' \
+      . | tee /dev/stderr |
+      yq -r '.metadata.annotations["iam.gke.io/gcp-service-account"]' | tee /dev/stderr)
+  [ "${actual}" = "user-defined-gsa@my-project.iam.gserviceaccount.com" ]
```
tvoran, line 37: I think this test belongs in the test block below, "specify annotations", and it may be a little redundant.

test/unit/server-serviceaccount.bats
```diff
@@ -2,6 +2,41 @@
 
 load _helpers
 
+@test "server/ServiceAccount: specify service account name" {
+  cd `chart_dir`
+
+  local actual=$( (helm template \
+      --show-only templates/server-serviceaccount.yaml  \
+      --set 'server.dev.enabled=true' \
+      --set 'server.serviceAccount.enabled=false' \
```
tvoran, line 11: I wonder if you could add a test or two that checks the serviceAccount in the statefulSet, when `server.serviceAccount.name` is set, and `server.serviceAccount.enabled` is `true` and `false`.

burningalchemist: Sure, will do 👍

burningalchemist:
Hi @tvoran ! Thanks for your review, please take a look, whenever you have time.

Issue #190, opened 2020-01-31
Ingress does not work with GCE ingress controller on GKE


gorilych:
Ingress events:

```
$ kubectl  -n vault describe ing vault
...
Events:
  Type     Reason             Age                     From                     Message
  ----     ------             ----                    ----                     -------
  Normal   ADD                9m49s                   loadbalancer-controller  vault/vault
  Normal   CreateCertificate  9m49s                   cert-manager             Successfully created Certificate "vault-tls"
  Warning  Sync               3m19s (x17 over 9m45s)  loadbalancer-controller  Error during sync: error running backend syncing routine: googleapi: Error 400: Invalid value for field 'resource.timeoutSec': '5'. TimeoutSec should be less than checkIntervalSec., invalid
```

GCE ingress controller tries to create healthchecks for GCLB which has strict requirement for timeout be less than check interva, see https://cloud.google.com/load-balancing/docs/health-checks and https://github.com/kubernetes/ingress-gce/blob/master/docs/faq/gce.md#can-i-configure-gce-health-checks-through-the-ingress

Unfortunately values for checking interval (`periodSeconds`) and check timeout (`timeoutSeconds`) are hardcoded here https://github.com/hashicorp/vault-helm/blob/master/templates/server-statefulset.yaml#L108 and there is no way to fix this without forking helm chart.

karthikm83:
+1:

This is a bug and should be fixed. Either these values should be left to default or configurable through values.yaml

It makes sense that `periodSeconds` should be higher than `timeoutSeconds`

```
periodSeconds: How often (in seconds) to perform the probe. Default to 10 seconds. Minimum value is 1.

timeoutSeconds: Number of seconds after which the probe times out. Defaults to 1 second. Minimum value is 1.
```

jasonodonnell:
We've made all these values configurable: https://github.com/hashicorp/vault-helm/pull/387

PR #379, opened 2020-08-27
Update statefulset liveness/readiness probe timeout value

karthikm83:
Related to Issue: https://github.com/hashicorp/vault-helm/issues/190


jasonodonnell:
Closing because #387 exposed all configurable probe values.

PR #381, opened 2020-09-01
Allow explicit network policy enablement

corest:
For now, the only way to create a network policy - is to enable `openshift`. 
I don't need all the resources enabled by `openshift` configuration.
But I still need network policy.
This PR allows configuring network policy creation without enabling `openshift`

### DIFF
--- a/templates/server-network-policy.yaml
+++ b/templates/server-network-policy.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.global.openshift }}
+{{- if or (.Values.global.openshift) (eq (.Values.server.networkPolicy.enabled | toString) "true" ) }}
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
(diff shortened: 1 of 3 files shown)


jasonodonnell:
Hi @corest, thanks for the contribution.

This sounds like a good idea but is a little confusing since you may want to install in OpenShift but have the policy disabled.  I think we can just make this configurable using `networkPolicy.enabled`.  Thoughts?

Additionally you'll need to add unit tests to make sure this configurable works as expected.  You can find the `bats` unit tests in `test/unit`: https://github.com/hashicorp/vault-helm/blob/master/test/unit/server-network-policy.bats

corest:
> This sounds like a good idea but is a little confusing since you may want to install in OpenShift but have the policy disabled. I think we can just make this configurable using networkPolicy.enabled. Thoughts?

I wanted to do that first but didn't want to break compatibility for someone who is using `openshift` configuration already and relies on it. But I agree that the network policy configuration should not be something we deploy or not based on specific cloud env. I'll change it with `networkPolicy.enabled` only

corest:
@jasonodonnell PTAL

corest:
Can we merge this @jasonodonnell ?

jasonodonnell (commented):
LGTM. @tvoran can you also take a look at this? It's changing network policy default behavior but since NP isn't OpenShift specific, I'm recommending we break it out as such.

tvoran (approved):
@jasonodonnell Seems ok to me, and we'll probably want to add a note about it to our OpenShift docs once it's released.

PR #383, opened 2020-09-08
hashicorp/vault-helm/values: Using YAML syntax for config instead of a string

arpithank:
Since using strings for config specification is somewhat deprecated since https://github.com/hashicorp/vault-helm/pull/213, this change allows to specify the config using YAML throughout the values file. 

This way it's easier to override values from a base file. (Instead of specifying the entire config string in region/environment specific yaml files for example)

This is my first time contributing! I hope this PR helps.


jasonodonnell:
Thanks @arpithank, but we're going to keep the HCL default for now. YAML is supported but since almost all of our examples are HCL based, we'll want to keep that consistent with the rest of Vault's ecosystem.

Issue #331, opened 2020-06-18
Number of vault-agent-injector replicas cannot be set in values


pamiel:
It is not possible, through the values, to set the number of replicas for the vault-agent-injector.

This somehow leads to consider the vault-agent-injector as a SPOF for the solution: if for any reason the single vault-agent-injector instance is down or not reachable when a "micro-service" Pod is starting, then the side car container of this micro-service Pod will not be created, so the data will not be retrieved from Vault, most probably causing the micro-service Pod to not be able to run properly.

For HA mode, the number of replicas of the vault-agent-injector should at least be hard-coded to 2, or should preferably be configurable using values.

jasonodonnell:
Hi @pamiel, auto TLS in Vault K8s has a bug where each pod of the injector races to update the CA bundle.  Due to this bug, replicas cannot be scaled past 1 when using auto tls (and thus why it's not exposed).

If you manually mount certificates to the injector, it can be safely scaled beyond 1.  

This bug is currently being worked on so this should be resolved soon.  Hope that helps!

pamiel:
Ok, very clear, thanks !

michalschott:
@jasonodonnell was not aware about this bug when raising https://github.com/hashicorp/vault-helm/pull/384, is this still the case?

tvoran:
Hi @pamiel, the number of replicas can now be changed in v0.9.0 of the chart, and some documentation describing the setup will be going up in this PR: https://github.com/hashicorp/vault/pull/10659

PR #384, opened 2020-09-09
vault-agent-injector - adding PDB

michalschott:
Adding ability to add PDB to the injector pods.


dddomin3:
Does the presence of this PR mean this issue is no longer a problem? https://github.com/hashicorp/vault-helm/issues/331

michalschott:
I have updated/rebased PR against recent master branch.

PR #386, opened 2020-09-11
Update Jira sync action versions

tomhjp:
These versions bring a few fixes:
* The action now supports converting the most common bits of markdown syntax into Jira formatting directives
* Replaces Atlassian's comment action, which tries to interpolate bits of text from comments wrapped in {{ github.event_name }} as templates, usually causing an error (in the case I've put there, if I wasn't a vault team member, it would replace the template with `pull_request_target`)
* Remove trailing comma after link that broke the link target Jira selects

### DIFF
--- a/.github/workflows/jira.yaml
+++ b/.github/workflows/jira.yaml
@@ -48,12 +48,12 @@ jobs:
 
     - name: Create ticket
       if: github.event.action == 'opened' && !steps.vault-team-role.outputs.role
-      uses: tomhjp/gh-action-jira-create@v0.1.0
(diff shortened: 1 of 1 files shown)


jasonodonnell (commented):
Should we move these actions into the HashiCorp org so they aren't tied to you @tomhjp ?

jasonodonnell (approved):

tomhjp:
> Should we move these actions into the HashiCorp org so they aren't tied to you @ tomhjp ?

I think this will probably happen at some point, but there didn't seem to be much appetite for it when I asked around after hack week.

PR #387, opened 2020-09-14
Add configurable probe values

jasonodonnell:
This makes hardcoded probe values configurable with some defaults out of the box so users can configure readiness/liveness probes for their environment.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -237,7 +237,7 @@ server:
     path: "/v1/sys/health?standbyok=true"
     # When a probe fails, Kubernetes will try failureThreshold times before giving up
     failureThreshold: 2
-    # Number of seconds after the container has started before probbe initiates
+    # Number of seconds after the container has started before probe initiates
     initialDelaySeconds: 60
     # How often (in seconds) to perform the probe
     periodSeconds: 3
--- a/values.yaml
+++ b/values.yaml
@@ -218,11 +218,33 @@ server:
     enabled: true
     # If you need to use a http path instead of the default exec
     # path: /v1/sys/health?standbyok=true
+    
+    # When a probe fails, Kubernetes will try failureThreshold times before giving up
+    failureThreshold: 2
+    # Number of seconds after the container has started before probbe initiates
+    initialDelaySeconds: 5
+    # How often (in seconds) to perform the probe
+    periodSeconds: 3
+    # Minimum consecutive successes for the probe to be considered successful after having failed
+    successThreshold: 1
+    # Number of seconds after which the probe times out.
+    timeoutSeconds: 5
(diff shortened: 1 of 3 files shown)


tomhjp (approved):
LGTM

values.yaml
```diff
@@ -218,11 +218,33 @@ server:
...
+    # Number of seconds after the container has started before probe initiates
+    initialDelaySeconds: 5
+    # How often (in seconds) to perform the probe
+    periodSeconds: 3
+    # Minimum consecutive successes for the probe to be considered successful after having failed
+    successThreshold: 1
+    # Number of seconds after which the probe times out.
+    timeoutSeconds: 5
+    # If you need to use a http path instead of the default exec
+    # path: /v1/sys/health?standbyok=true
```
tvoran, line 233: It looks like this is duplicated from 219-220.
```suggestion
```

values.yaml
```diff
@@ -218,11 +218,31 @@ server:
...
+    # When a probe fails, Kubernetes will try failureThreshold times before giving up
+    failureThreshold: 2
+    # Number of seconds after the container has started before probe initiates
+    initialDelaySeconds: 5
+    # How often (in seconds) to perform the probe
+    periodSeconds: 3
+    # Minimum consecutive successes for the probe to be considered successful after having failed
+    successThreshold: 1
+    # Number of seconds after which the probe times out.
+    timeoutSeconds: 5
```
tvoran, line 231: I see these were the previous defaults, but I'm wondering if the `timeoutSeconds` should be less than the `periodSeconds`? Seems like you'd want to give a probe the chance to timeout before firing another one? Same question applies to the livessProbe section.

jasonodonnell: Good point, I switched them!

tvoran (approved):

values.yaml
```diff
@@ -218,11 +218,31 @@ server:
...
+    # When a probe fails, Kubernetes will try failureThreshold times before giving up
+    failureThreshold: 2
+    # Number of seconds after the container has started before probe initiates
     initialDelaySeconds: 60
+    # How often (in seconds) to perform the probe
+    periodSeconds: 5
+    # Minimum consecutive successes for the probe to be considered successful after having failed
+    successThreshold: 1
+    # Number of seconds after which the probe times out.
+    timeoutSeconds: 3
```
thomas-riccardi, line 245: was this desired to swap period and timeout values? It was timeout:5 and period:3 before that PR.

tvoran: Yeah, as discussed here: https://github.com/hashicorp/vault-helm/pull/387#discussion_r488131418

PR #389, opened 2020-09-18
Allow configurable egress for server network policy

corest:
There are specific cases when the server pod needs to communicate with some external components.
E.g. when using various auth backends:
- kubernetes - server needs to talk with k8s api
- github - server needs to talk with github

### DIFF
--- a/templates/server-network-policy.yaml
+++ b/templates/server-network-policy.yaml
@@ -19,4 +19,8 @@ spec:
         protocol: TCP
       - port: 8201
         protocol: TCP
+  {{- if .Values.server.networkPolicy.egress }}
+  egress:
+    - {}
+  {{ end }}
 {{ end }}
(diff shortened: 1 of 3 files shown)


jasonodonnell:
@corest Thanks for the contribution!

This use case makes sense but I wonder if we need to allow users to pass in whatever configurations they want for their egress.
For example, instead of `egress` being a boolean, we instead just take whatever YAML the operator wants to add. `server.initContainers` is a good example of this: 
https://github.com/hashicorp/vault-helm/blob/master/templates/server-statefulset.yaml#L61-L63
https://www.vaultproject.io/docs/platform/k8s/helm/configuration#extrainitcontainers

Using this design an operator can add whatever `to` rules they want for their environment and we don't need a bunch of very specific configurables.

Thoughts?

corest:
@jasonodonnell agree
Added test for egress templation

jasonodonnell (approved):

Issue #376, opened 2020-08-25
Token review permission missing when using external vault address


gokuatkai:
Using an external vault address (aka **external** mode) does not create `ClusterRoleBinding` for `system:auth-delegator` and make the request unable to authenticate:

Pod:
```
2020-08-25T15:14:13.077Z [INFO]  auth.handler: authenticating
2020-08-25T15:14:13.170Z [ERROR] auth.handler: error authenticating: error="Error making API request.

URL: PUT https://my-vault.com:8200/v1/auth/kubernetes/login
Code: 403. Errors:

* permission denied" backoff=2.822450632
```
Server:
```
Aug 25 15:13:10 my-vault.com vault[1803]: 2020-08-25T15:13:10.793Z [ERROR] auth.kubernetes.auth_kubernetes_a3c6e8ec: login unauthorized due to: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"tokenreviews.authentication.k8s.io is forbidden: User \"system:serviceaccount:security:vault-testenv-agent-injector\" cannot create resource \"tokenreviews\" in API group \"authentication.k8s.io\" at the cluster scope","reason":"Forbidden","details":{"group":"authentication.k8s.io","kind":"tokenreviews"},"code":403}
```

jasonodonnell:
Hi @gokuatkai, that's correct, no role is setup for Vault to use.  This would be a nice enhancement to add, though, so labeling this as such.

gokuatkai:
Hi, Do you think we could change `server-clusterrolebinding.yaml`
and add some logic to change the `subjects[].name` to `name: {{ template "vault.fullname" . }}-agent-injector` so it adds automatically the role if mode `external` is specified?

gokuatkai:
so it only consists of a clusterrolebinding and a serviceaccount, maybe it can be a separate chart ?

jasonodonnell:
We already do create SA and CRB, however, that's only triggered when Vault server is installed via Helm.  Basically we can add an additional value to also create those if Vault is external.

MPV:
Has this been solved now (with #392 being merged)?

tvoran:
@MPV Yep, looks like it!

Closing this, #392 was included in v0.8.0: https://github.com/hashicorp/vault-helm/blob/master/CHANGELOG.md#080-october-20th-2020

PR #392, opened 2020-09-29
Enable Vault to review kube tokens when using external Vault

jfroche:
We want Vault to perform token reviews with Kubernetes even if we are
using an external Vault.

We need to create the ServiceAccount, Secret and ClusterRoleBinding with
the system:auth-delegator role to enable delegated authentication and
authorization checks [1].

These SA and RBAC objects are created when we deploy the Vault server.
In order to enable the creation of these objects when using an external
Vault, we remove the condition on external mode.

User might want to provide a sensible name (in global.serviceAccount.name) to the service
account such as: vault-auth.

refs #376

[1] https://www.vaultproject.io/docs/auth/kubernetes#configuring-kubernetes

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -70,7 +70,7 @@ injector:
   #      sidecar-injector: enabled
   namespaceSelector: {}
 
-  # Configures failurePolicy of the webhook. By default webhook failures are ignored. 
+  # Configures failurePolicy of the webhook. By default webhook failures are ignored.
   # To block pod creation while webhook is unavailable, set the policy to `Fail` below.
   # See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy
   #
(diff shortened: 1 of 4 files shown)


michalschott:
I'm facing the same problem here, temporary workaround is to do helm template and patch with kustomize. Upvoting.

values.yaml
```diff
@@ -22,6 +22,19 @@ global:
...
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: true
+    # The name of the service account to use.
+    # If not set and create is true, a name is generated using the fullname template
+    name: ""
+    # Extra annotations for the serviceAccount definition. This can either be
+    # YAML or a YAML-formatted multi-line templated string map of the
+    # annotations to apply to the serviceAccount.
+    annotations: {}
```
jasonodonnell, line 37: This should not be moved away from `server`. This will make the change backwards compatible and it's specific to the server, not a global config.

jasonodonnell:
@jfroche This looks good to me but we need to move SA back to server. This will break existing users and while I appreciate increasing the visibility of this setting, it's specific to the `server` and not a global config. I'll merge once this is moved 😄

jfroche:
@jasonodonnell Ok, fixed it

jasonodonnell (approved):
LGTM, thanks!

PR #393, opened 2020-09-30
Add configurable mountPath for audit/data storage

jasonodonnell:
This PR makes the mountPath for data and audit persistent volumes configurable when users prefer different mounts in the filesystem. The defaults are the same as they've always been but are now configurable via values.


tomhjp (approved):
LGTM

tvoran (approved):

PR #394, opened 2020-09-30
add ability to set pod annotations for injector

mrparkers:
this PR adds the ability to add pod annotations to the injector pod.  our use case for this is that we want to install Vault within an Istio enabled namespace and join Vault to the mesh, but we don't want to join the injector pod to the mesh.  so we need to specify pod annotations to exclude it.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -283,6 +283,21 @@ Sets extra pod annotations
 {{- end -}}
 
 {{/*
+Sets extra injector pod annotations
+*/}}
+{{- define "injector.annotations" -}}
+  {{- if .Values.injector.annotations }}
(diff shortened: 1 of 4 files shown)


jasonodonnell:
Thanks for the contribution @mrparkers!

Can you add unit tests so we have coverage on this? The unit tests would go here: https://github.com/hashicorp/vault-helm/blob/master/test/unit/injector-deployment.bats

You can basically copy/modify existing annotation tests: https://github.com/hashicorp/vault-helm/blob/master/test/unit/server-statefulset.bats#L903-L932.

Let me know if you have any questions!

jasonodonnell (approved):
LGTM, thanks!

Issue #390, opened 2020-09-21
Helm v3.3.0 linting declares "rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding" deprecated.


patrickstump:
**Describe the bug**
When using the lates version of helm v3.3.0 and linting the helm chart the error below is received and causes linting (and our pipeline) to fail.
```
Error: 1 chart(s) linted, 1 chart(s) failed
==> Linting vault
[ERROR] templates/server-clusterrolebinding.yaml: the kind "rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding" is deprecated in favor of "rbac.authorization.k8s.io/v1 ClusterRoleBinding"
```
**To Reproduce**
1. Install the latest version of helm 
    - our version ```version.BuildInfo{Version:"v3.3.0", GitCommit:"8a4aeec08d67a7b84472007529e8097ec3742105", GitTreeState:"dirty", GoVersion:"go1.14.7"}```

2. ```helm repo add hashicorp https://helm.releases.hashicorp.com```
3.  ```helm repo update```
4. ```helm pull hashicorp/vault --untar```
5. ```helm lint vault```

This will result in the following error:
```
$ helm lint vault
Error: 1 chart(s) linted, 1 chart(s) failed
==> Linting vault
[ERROR] templates/server-clusterrolebinding.yaml: the kind "rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding" is deprecated in favor of "rbac.authorization.k8s.io/v1 ClusterRoleBinding"
```

**Expected behavior**
The helm linter should return without error.

**Environment**
* Kubernetes version:  NA 
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): NA
  * Other configuration options or runtime services (istio, etc.): NA
* vault-helm version: 
```
helm show chart hashicorp/vault
apiVersion: v2
appVersion: 1.5.2
description: Official HashiCorp Vault Chart
home: https://www.vaultproject.io
icon: https://github.com/hashicorp/vault/raw/f22d202cde2018f9455dec755118a9b84586e082/Vault_PrimaryLogo_Black.png
keywords:
- vault
- security
- encryption
- secrets
- management
- automation
- infrastructure
name: vault
sources:
- https://github.com/hashicorp/vault
- https://github.com/hashicorp/vault-helm
- https://github.com/hashicorp/vault-k8s
version: 0.7.0

```

Chart values:

None: just the linting command

[...]

PR #395, opened 2020-10-05
Update rbac api version to v1

cpieper:
Addresses the error reported in latest Helm release reported here - https://github.com/hashicorp/vault-helm/issues/390

```
Error: malformed chart or values:
	templates/server-clusterrolebinding.yaml: the kind "rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding" is deprecated in favor of "rbac.authorization.k8s.io/v1 ClusterRoleBinding"
```

Deprecation documented here: https://github.com/kubernetes/kubernetes/blob/adbc7c2172ea622db7399f41e893063b00ca1a87/CHANGELOG/CHANGELOG-1.17.md#deprecations-and-removals

### DIFF
--- a/templates/server-clusterrolebinding.yaml
+++ b/templates/server-clusterrolebinding.yaml
@@ -1,7 +1,7 @@
 {{ template "vault.mode" . }}
 {{- if ne .mode "external" }}
 {{- if and (ne .mode "") (and (eq (.Values.global.enabled | toString) "true") (eq (.Values.server.authDelegator.enabled | toString) "true")) }}
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: {{ template "vault.fullname" . }}-server-binding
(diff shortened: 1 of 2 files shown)


templates/server-clusterrolebinding.yaml
```diff
@@ -1,7 +1,7 @@
 {{ template "vault.mode" . }}
 {{- if ne .mode "external" }}
 {{- if and (ne .mode "") (and (eq (.Values.global.enabled | toString) "true") (eq (.Values.server.authDelegator.enabled | toString) "true")) }}
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
lawliet89, line 4: I suggest using `.Capabilities.APIVersions.Has` to see if the Kubernetes version supports this API version to preserve compatibility with older versions of Kubernetes.

e.g.
```suggestion 
{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" -}}
apiVersion: rbac.authorization.k8s.io/v1
{{- else }}
apiVersion: rbac.authorization.k8s.io/v1beta1
{{- end }}
```

Chili-Man:
@cpieper any chance you can address the comment ? Otherwise I can carry this pr forward

cpieper:
@Chili-Man I've added @lawliet89 's recommendation to include the `capabilities` check.

cpieper:
@lawliet89 Could you take another look at these changes? Any feedback regarding the acceptance of this PR would be appreciated.

lawliet89:
Hey @cpieper I don't have write access to this repo. @jasonodonnell would be more helpful.

jasonodonnell (approved):

PR #396, opened 2020-10-06
Allow injection of TLS config for OpenShift routes

aRestless:
**Purpose of this PR:** On OpenShift it is common to let the cluster-wide router do the TLS termination, either through a wildcard certificate for its standard domain, or a per-route certificate. See https://docs.openshift.com/container-platform/4.5/networking/routes/secured-routes.html. These termination policies "edge" and "reencrypt" support additional configuration parameters on the route, specifically for the per-route certificate. The current chart version in master does not support to set these values or to even change the termination policy.

**Solution:** Add a `tls` block to `server.route` that is injected directly into the route config (`.spec.tls`) and set the default in values.yml to passthrough termination to not change default behavior.

**Alternative:** Include and document any values that are supported by `.spec.tls` in `server.route.tls` instead of injecting the whole YAML block.

**Outlook:** The default standalone config in the chart disables TLS at the vault listener, which is somewhat okay for in-cluster traffic, but a more production-ready default for OpenShift routes would be to assume termination policy "edge" instead of "passthrough", and to not set the route host. Not setting the route host generates a default hostname based on the OpenShift project name. Typical OpenShift clusters have a wildcard certificate set up for these default hostnames.

This would be a breaking change since it would change the chart's default behavior, which is why I left this part out of the PR for now.

### DIFF
--- a/templates/server-route.yaml
+++ b/templates/server-route.yaml
@@ -27,7 +27,7 @@ spec:
   port:
     targetPort: 8200
   tls:
-    termination: passthrough
+    {{- toYaml .Values.server.route.tls | nindent  4 }}
 {{- end }}
 {{- end }}
 {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -176,12 +176,17 @@ server:
     #      - chart-example.local
 
   # OpenShift only - create a route to expose the service
-  # The created route will be of type passthrough
+  # By default the created route will be of type passthrough
   route:
     enabled: false
     labels: {}
     annotations: {}
     host: chart-example.local
+    # tls will be passed directly to the route's TLS config, which
+    # can be used to configure other termination methods that terminate
+    # TLS at the router
+    tls:
+      termination: passthrough
 
   # authDelegator enables a cluster role binding to be attached to the service
   # account.  This cluster role binding can be used to setup Kubernetes auth


leofin:
I open a PR very close to this one a few days ago. I liked more your approach. Please @HashiCoop review it?

tvoran (commented):
Apologies for the delay! I agree this seems like a good approach for the `tls` block; would you mind adding a couple tests (one for the default and one for a set value) to test/unit/server-route.bats?

mbaldessari:
I added some tests here in https://github.com/mbaldessari/vault-helm/commit/4be1572ef027fd89937a34047e82dc500dd75af9 @aRestless could you merge something along those lines in your PR here ? (If not we can add them on top after merging this one, I guess)

mbaldessari:
> Apologies for the delay! I agree this seems like a good approach for the `tls` block; would you mind adding a couple tests (one for the default and one for a set value) to test/unit/server-route.bats?

@tvoran Could we maybe merge this one as is and I'll make a PR out of https://github.com/mbaldessari/vault-helm/commit/4be1572ef027fd89937a34047e82dc500dd75af9 to add tests on top or should I propose a new PR squashing this + tests? (I suspect I cannot update this existing PR)

tvoran:
Hi @mbaldessari, if you don't mind, please open a new PR from your branch (mbaldessari:tests-on-top-of-396), and we can continue reviewing it there. No need to squash it unless you run into trouble resolving conflicts or something.

mbaldessari:
Hi @tvoran, thanks! I pushed https://github.com/hashicorp/vault-helm/pull/686

tvoran:
Superseded by https://github.com/hashicorp/vault-helm/pull/686

PR #404, opened 2020-10-20
Allow running pods in the host network

acroca:
In our setup we have some vault instances running externally and new instances running in K8S using consul storage.
If one of the K8S instances get elected as active, all traffic is forwarded from the external instances to the active instance in K8S, so running it in the host network (where the existing instances are running as well) will make it easily reachable.

I'm not convinced this is the best way to make vault pods visible in the host network, I'm happy to try any other.

If this is a good approach, I will add the unit tests.

### DIFF
--- a/templates/server-psp.yaml
+++ b/templates/server-psp.yaml
@@ -22,7 +22,14 @@ spec:
     {{- if eq (.Values.server.dataStorage.enabled | toString) "true" }}
     - persistentVolumeClaim
     {{- end }}
+  {{- if .Values.server.hostNetwork }}
+  hostNetwork: {{ .Values.server.hostNetwork }}
+  hostPorts:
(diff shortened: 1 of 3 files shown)


jasonodonnell (commented):
Hi @acroca, sorry for the delay, its been busy here :)

This is an interesting change. I'm wondering how this is better than using `nodePort`. Using the downward API, it's possible to configure the IP address of the host you're running on. Using that IP address and node port, pods running on that host can communicate with the pod.

Thoughts?

acroca:
Hi @jasonodonnell. I will try that and let you know, thanks for the suggestion.

jasonodonnell:
Closing due to inactivity. If you think this is a mistake, please let me know!

PR #406, opened 2020-10-21
Fix Raft config provided in YAML

gw0:
According to the improvement in #213 I attempted to use YAML config in Raft mode. It turns out, that the type check in this situation does not work correctly.

```
  ha:
    enabled: true
    #config: {}  # uncomment as a workaround for this bug to provide the correct type
    raft:
      enabled: true
      config:
        ...(yaml config)...
```

Results in:

```
Error: template: vault/templates/server-config-configmap.yaml:26:18: executing "vault/templates/server-config-configmap.yaml" at <.Values.server.ha.raft.config>: wrong type for value; expected string; got map[string]interface {}
```

### DIFF
--- a/templates/server-config-configmap.yaml
+++ b/templates/server-config-configmap.yaml
@@ -15,7 +15,11 @@ metadata:
 data:
   extraconfig-from-values.hcl: |-
   {{- if or (eq .mode "ha") (eq .mode "standalone") }}
+  {{- if eq (.Values.server.ha.raft.enabled | toString) "true" }}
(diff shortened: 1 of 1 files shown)


jasonodonnell (commented):
Sorry for the delay @gw0, good catch on this bug. It seems this broke a bunch of unit tests, though?

gw0:
Yeah, I did not manually run the unit tests and I do not want to log into CircleCI in order to see the details of what is failing.

PR #408, opened 2020-10-21
Update mutating webhook API Version

lawliet89:
`admissionregistration.k8s.io/v1beta1` is [deprecated](https://v1-16.docs.kubernetes.io/docs/setup/release/notes/#deprecations-and-removals) since 1.16 and will be removed in 1.19.

This PR will set the right API Version if it is supported.

**However**, there is a change in behaviour from `v1beta1` to `v1`. Specifically, the default (unspecified) [`failurePolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy) changes from `Ignore` to `Fail`. Should we set the default in `values.yaml` to `Ignore` to preserve the previous behaviour? Or this can be labelled a breaking change?

### DIFF
--- a/templates/injector-mutating-webhook.yaml
+++ b/templates/injector-mutating-webhook.yaml
@@ -13,6 +13,10 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 webhooks:
   - name: vault.hashicorp.com
+    sideEffects: None
+    admissionReviewVersions:
+    - "v1beta1"
+    - "v1"
     clientConfig:
       service:
         name: {{ template "vault.fullname" . }}-agent-injector-svc
--- a/templates/injector-mutating-webhook.yaml
+++ b/templates/injector-mutating-webhook.yaml
@@ -1,5 +1,5 @@
 {{- if and (eq (.Values.injector.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
-{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" -}}
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" }}
 apiVersion: admissionregistration.k8s.io/v1
 {{- else }}
 apiVersion: admissionregistration.k8s.io/v1beta1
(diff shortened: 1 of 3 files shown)


jasonodonnell:
@lawliet89 Thanks! I think we should set it to `Ignore` in values by default since that's the expected behavior today.

lawliet89:
@jasonodonnell I have set it to `Ignore` by default.

I also tried adding tests to see if we can test if the `apiVersion` is set correctly. `helm template` has the `--api-versions` flag but it seems to be an additive operation.

```
$ helm template       --show-only templates/injector-mutating-webhook.yaml        --set 'injector.enabled=true'       --namespace foo --api-versions '["admissionregistration.k8s.io/v1beta1"]'       .
---
# Source: vault/templates/injector-mutating-webhook.yaml
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  name: RELEASE-NAME-vault-agent-injector-cfg
  labels:
    app.kubernetes.io/name: vault-agent-injector
webhooks:
  - name: vault.hashicorp.com
    clientConfig:
      service:
        name: RELEASE-NAME-vault-agent-injector-svc
        namespace: foo
        path: "/mutate"
      caBundle: ""
    rules:
      - operations: ["CREATE", "UPDATE"]
        apiGroups: [""]
        apiVersions: ["v1"]
        resources: ["pods"]
    failurePolicy: Ignore
```

So I have not

lawliet89:
I have also added new fields that are required for `v1` as mentioned in https://github.com/hashicorp/consul-helm/pull/658#discussion_r510256304

lawliet89:
@jasonodonnell Could you review this again please?

jasonodonnell:
@lawliet89 looks good but I'm wondering if `vault-k8s` also needs to be updated? `auto-tls` updates the caBundle when certificates change: https://github.com/hashicorp/vault-k8s/blob/master/agent-inject/handler.go#L81-L82

jasonodonnell:
@lawliet89 I asked a similar question here: https://github.com/hashicorp/vault-k8s/pull/168

lawliet89:
I am not too familiar with the kubernetes api library. I'm not sure if a v1 api library function can work with v1beta1 endpoint. It's not quite the same change as this PR imo.

jasonodonnell (approved):
LGTM, tested out fine here.

Issue #361, opened 2020-07-27
alb Ingress - http to https redirection.


davorceman:
**Is your feature request related to a problem? Please describe.**
I can't create alb ingress with the ssl redirection. Currently after deploying with this in my vaules file
```
  ingress:
    enabled: true
    labels: {}
    annotations:
      kubernetes.io/ingress.class: alb
      alb.ingress.kubernetes.io/scheme: ${ingress_scheme}
      alb.ingress.kubernetes.io/healthcheck-path: /v1/sys/health
      alb.ingress.kubernetes.io/success-codes: 429,200
      alb.ingress.kubernetes.io/security-groups: ${ingress_security_group}
      alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80,"HTTPS":443}]'
      alb.ingress.kubernetes.io/certificate-arn: ${ingress_certificate_arn}
      alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
      alb.ingress.kubernetes.io/tags: ${ingress_tags}
    hosts:
      - host: ${ingress_host}
        paths: ["/*"]
  service:
    type: NodePort
```
... there is no redirection from http to https

**Describe the solution you'd like**
I want to have possibility to add one more path in rules for SSL redirection.
Currently code is
```
  rules:
  {{- range .Values.server.ingress.hosts }}
    - host: {{ .host | quote }}
      http:
        paths:
        {{- range (.paths | default (list "/")) }}
          - path: {{ . }}
            backend:
              serviceName: {{ $serviceName }}
              servicePort: {{ $servicePort }}
        {{- end }}
  {{- end }}
```
and I don't see the way to add
```
- path: /*
  backend:
    serviceName: ssl-redirect
    servicePort: use-annotation
```
as a second path to the same host.

peimanja:
Same issue here.

PR #413, opened 2020-11-17
Add ability to provide extra Ingress backends (refs #361).

rossigee:
A workaround to provide the ability to add extra backends, as required to get SSL redirection to work properly when using this chart with an AWS Application Load Balancer.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vault
-version: 0.8.1
+version: 0.8.0
 appVersion: 1.5.4
 description: Official HashiCorp Vault Chart
 home: https://www.vaultproject.io
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vault
-version: 0.8.0
+version: 0.8.1
 appVersion: 1.5.4
 description: Official HashiCorp Vault Chart
 home: https://www.vaultproject.io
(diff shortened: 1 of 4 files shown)


templates/server-ingress.yaml
```diff
@@ -41,6 +41,12 @@ spec:
     - host: {{ .host | quote }}
       http:
         paths:
+        {{- range .extraBackends }}
```
jasonodonnell, line 44: It's not obvious to me how this is different than the existing configurable `.paths`. Wondering if you can clarify?

Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vault
-version: 0.8.0
+version: 0.8.1
```
jasonodonnell, line 3: We'll update the version when we release the chart. Please remove this change.

jasonodonnell (commented):
Hi @rossigee, thanks for the contribution! Please see my comments above.

templates/server-ingress.yaml
rossigee: Sure. I think @davorceman described the problem this solves well in #361.

To clarify the difference: The problem with the existing configurable `.paths` is it's assumptions of the values made by '$servicename' and '$serviceport'.

ollandos:
Same issue here. Can we have this?

samjmarshall:
@jasonodonnell this has been implemented by PR https://github.com/hashicorp/vault-helm/pull/460 :) Thanks for pushing it through. My team and I really appreciate it dude!

rossigee:
Superceded by https://github.com/hashicorp/vault-helm/pull/460 :rocket:

PR #415, opened 2020-11-19
don't set VAULT_DEV_ROOT_TOKEN_ID by default in dev mode

hpio:
Do not set VAULT_DEV_ROOT_TOKEN_ID environment variable by default when running in dev mode as it can already be set using `extraSecretEnvironmentVars` or `extraEnvironmentVars`. 
At the moment it is impossible to set VAULT_DEV_ROOT_TOKEN_ID to a custome value as variable will be duplicated in a resulting yaml file.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -140,7 +140,7 @@ Set's additional environment variables based on the mode.
 {{- define "vault.envs" -}}
   {{ if eq .mode "dev" }}
             - name: VAULT_DEV_ROOT_TOKEN_ID
-              value: {{ .Values.server.dev.devRootToken | default "root" }}
+              value: {{ .Values.server.dev.devRootToken }}
   {{ end }}
 {{- end -}}
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -135,6 +135,14 @@ for users looking to use this chart with Consul Helm.
 {{- end -}}
 
 {{/*
+Set's additional environment variables based on the mode.
+*/}}
+{{- define "vault.envs" -}}
+  {{ if eq .mode "dev" }}
+            - name: VAULT_DEV_ROOT_TOKEN_ID
+              value: {{ .Values.server.dev.devRootToken | default "root" }}
+  {{ end }}
+{{- end -}}
 
 {{/*
 Set's which additional volumes should be mounted to the container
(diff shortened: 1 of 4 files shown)


jasonodonnell (changes requested):
Hi @hpio, thanks for the contribution!

I think it might be better to just make a config under the `server.dev` settings that allow you to override the default token. This has been set to `root` since we've 0.1.0 so changing this behavior now isn't desirable since users could be relying on it for CI/testing purposes.

Definitely agree that the static value should be configurable so `server.dev.devRootToken` with a default value of `root` would be the way to go.

Hope that helps.

hpio:
thanks for the feedback @jasonodonnell . I made suggested changes and hopefully everything's good this time. I was not aware that token has been set to `root` for a long period of time as I only recently started using Vault.

templates/_helpers.tpl
```diff
@@ -140,7 +140,7 @@ Set's additional environment variables based on the mode.
 {{- define "vault.envs" -}}
   {{ if eq .mode "dev" }}
             - name: VAULT_DEV_ROOT_TOKEN_ID
-              value: "root"
+              value: {{ .Values.server.dev.devRootToken | default "root" }}
```
jasonodonnell, line 143: Instead of having `default "root"` here, set this in the `values.yaml` file and have it uncommented. We've been moving away from backend defaults to be more transparent in the `values.yaml` file.

values.yaml
```diff
@@ -417,6 +417,9 @@ server:
   dev:
     enabled: false
 
+    # Overrides dev root token value which by default is set to "root"
+    #devRootToken: "root"
```
jasonodonnell, line 421: Uncomment as per previous comment.

jasonodonnell (changes requested):
Small tweaks but looks good!

Additionally unit tests will need to be added for coverage on the configurable. Two simple tests demonstrating the default value is `root` and that it changes when configured would suffice.

Let me know if you have any questions.

hpio:
hey @jasonodonnell , hopefully latest changes are enough. Thanks for your help

jasonodonnell:
Changes look good @hpio at a glance. When I return next week I will give them a test and merge. Thanks!

jasonodonnell (approved):

jasonodonnell:
Thanks @hpio!

Issue #427, opened 2020-12-07
Configuration option to apply labels to agent-injector


logileifs:
**Is your feature request related to a problem? Please describe.**
I need to apply labels to the vault-agent-injector so it can bypass some network policies, currently there is no option in the helm chart to assign labels to it.

**Describe the solution you'd like**
Something like the extraLabels option for the server block would be good enough

**Describe alternatives you've considered**
I don't really know of any alternatives other than disabling the network policies and compromising security in the cluster

**Additional context**
Screenshot of agent-injector failing to start when network policies are in place
<img width="1518" alt="Screenshot 2020-12-07 at 14 01 03" src="https://user-images.githubusercontent.com/2317815/101359766-a6d19380-3894-11eb-8fe4-85148a909442.png">

jasonodonnell:
Fixed in #428

PR #428, opened 2020-12-07
support extraLabels for vault-agent-injector

logileifs:
fixes issue https://github.com/hashicorp/vault-helm/issues/427

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -23,6 +23,9 @@ spec:
         app.kubernetes.io/name: {{ template "vault.name" . }}-agent-injector
         app.kubernetes.io/instance: {{ .Release.Name }}
         component: webhook
+        {{- if  .Values.injector.extraLabels -}}
+          {{- toYaml .Values.injector.extraLabels | nindent 8 -}}
+        {{- end -}}
       {{ template "injector.annotations" . }}
     spec:
       {{ template "injector.affinity" . }}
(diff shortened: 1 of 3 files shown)


jasonodonnell (commented):
Hi @logileifs, thanks for the contribution!

This looks good but will require unit tests to make sure the configurable works and has coverage. Would you mind adding some? You can find the `bats` injector unit tests here: https://github.com/hashicorp/vault-helm/blob/master/test/unit/injector-deployment.bats.

Here's a test that's equivalent: https://github.com/hashicorp/vault-helm/blob/master/test/unit/server-statefulset.bats#L891-L901.

jasonodonnell:
@logileifs Also the CLA will need to be signed 😄

logileifs:
Hey @jasonodonnell thanks for the tips, I added a test for the extraLabels on the injector and I have already signed the CLA, I don't know why it's not updating 😕

logileifs:
Screenshot 2020-12-07 at 15 57 29

It's already signed but when I click recheck, nothing happens

jasonodonnell (commented):
One last change, this needs to be added to the `values.yaml` file. You can set it to blank just we like we do for `server`:

```yaml
injector:
  ...
  extraLabels: {}
```

jasonodonnell (approved):
LGTM, thanks for the contribution!

logileifs:
Thanks for the quick response @jasonodonnell and just out of curiosity, when can I expect to see this change in a release?

jasonodonnell:
We release monthly (if there's things to be released) on the third Tuesday, so good chance you'll see it then.

PR #433, opened 2020-12-14
Add optional LoadBalancer service for HA mode

tomhjp:
This adds an optional `LoadBalancer` service type when operating in HA mode, specifically to address cross-cluster replication, but more widely useful as well.

Still TODO: Acceptance tests.

Unit tests pass locally:
```bash
docker build -t vault-helm-test -f test/docker/Test.dockerfile .
docker run --rm -v `pwd`:/root vault-helm-test bats -t /root/test/unit/server-ha-lb.bats
1..15
ok 1 server/ha-lb-Service: generic annotations
ok 2 server/ha-lb-Service: disable with ha.enabled false
ok 3 server/ha-lb-Service: disable with ha.lb.enabled false
ok 4 server/ha-lb-Service: disable with server.service.enabled false
ok 5 server/ha-lb-Service: type LoadBalancer
ok 6 server/ha-lb-Service: clusterIP empty by default
ok 7 server/ha-lb-Service: externalTrafficPolicy Local and publishNotReadyAddresses false as defaults
ok 8 server/ha-lb-Service: externalTrafficPolicy can be set
ok 9 server/ha-lb-Service: publishNotReadyAddresses can be set
ok 10 server/ha-lb-Service: port and targetPort will be 8200 by default
ok 11 server/ha-lb-Service: port and targetPort can be set
ok 12 server/ha-lb-Service: nodeport can be set
ok 13 server/ha-lb-Service: nodeport can't set when type isn't NodePort
ok 14 server/ha-lb-Service: vault port name is http, when tlsDisable is true
ok 15 server/ha-lb-Service: vault port name is https, when tlsDisable is false
```

### DIFF
--- a/templates/server-ha-active-service.yaml
+++ b/templates/server-ha-active-service.yaml
@@ -1,4 +1,5 @@
 {{ template "vault.mode" . }}
+{{- if ne .mode "external" }}
 {{- if and (eq .mode "ha" ) (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
 # Service for active Vault pod
 apiVersion: v1
@@ -36,4 +37,5 @@ spec:
     app.kubernetes.io/instance: {{ .Release.Name }}
     component: server
     vault-active: "true"
+{{- end }}
 {{- end }}
--- a/templates/server-ha-active-service.yaml
+++ b/templates/server-ha-active-service.yaml
@@ -1,6 +1,5 @@
 {{ template "vault.mode" . }}
-{{- if ne .mode "external" }}
-{{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
+{{- if and (eq .mode "ha" ) (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
 # Service for active Vault pod
 apiVersion: v1
 kind: Service
@@ -37,5 +36,4 @@ spec:
     app.kubernetes.io/instance: {{ .Release.Name }}
     component: server
     vault-active: "true"
-{{- end }}
 {{- end }}
--- a/templates/server-ha-lb.yaml
+++ b/templates/server-ha-lb.yaml
@@ -1,10 +1,10 @@
 {{ template "vault.mode" . }}
-{{- if and (eq .mode "ha" ) (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
-# Service for standby Vault pod
+{{- if and (eq .mode "ha" ) (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") (eq (.Values.server.ha.lb.enabled | toString) "true") }}
+# LoadBalancer for active Vault pod
 apiVersion: v1
-kind: Service
+kind: LoadBalancer
 metadata:
-  name: {{ template "vault.fullname" . }}-standby
+  name: {{ template "vault.fullname" . }}-lb
   namespace: {{ .Release.Namespace }}
   labels:
     helm.sh/chart: {{ include "vault.chart" . }}
@@ -19,6 +19,11 @@ spec:
   {{- end}}
   {{- if .Values.server.service.clusterIP }}
   clusterIP: {{ .Values.server.service.clusterIP }}
+  {{- end }}
+  {{- if .Values.server.ha.local }}
+  externalTrafficPolicy: Local
+  {{- else }}
+  externalTrafficPolicy: Cluster
   {{- end }}
   publishNotReadyAddresses: true
   ports:
@@ -35,5 +40,5 @@ spec:
     app.kubernetes.io/name: {{ include "vault.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     component: server
-    vault-active: "false"
+    vault-active: "true"
 {{- end }}
--- a/templates/server-ha-lb.yaml
+++ b/templates/server-ha-lb.yaml
@@ -12,20 +12,11 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
   annotations:
-{{ template "vault.service.annotations" .}}
+{{ template "vault.lb.annotations" .}}
 spec:
-  {{- if .Values.server.service.type}}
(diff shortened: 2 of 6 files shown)


templates/server-ha-lb.yaml
```diff
@@ -0,0 +1,35 @@
...
+  ports:
+    - name: {{ include "vault.scheme" . }}
+      port: {{ .Values.server.service.port }}
+      targetPort: {{ .Values.server.service.targetPort }}
+      {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
+      nodePort: {{ .Values.server.service.nodePort }}
+      {{- end }}
+    - name: https-internal
+      port: 8201
+      targetPort: 8201
```
jasonodonnell, line 29: Think we also need 8202 as well

templates/server-ha-standby-service.yaml
```diff
@@ -1,6 +1,6 @@
 {{ template "vault.mode" . }}
 {{- if ne .mode "external" }}
-{{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
+{{- if and (eq .mode "ha" ) (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
```
jasonodonnell, line 3: Same comment as above.

templates/server-ha-lb.yaml
```diff
@@ -0,0 +1,35 @@
...
+  type: LoadBalancer
+  externalTrafficPolicy: {{ .Values.server.ha.lb.externalTrafficPolicy }}
+  publishNotReadyAddresses: {{ .Values.server.ha.lb.publishNotReadyAddresses }}
+  ports:
+    - name: {{ include "vault.scheme" . }}
+      port: {{ .Values.server.service.port }}
+      targetPort: {{ .Values.server.service.targetPort }}
+      {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
+      nodePort: {{ .Values.server.service.nodePort }}
+      {{- end }}
```
jasonodonnell, line 26: I don't think we need node port configurations.

templates/server-ha-lb.yaml
tomhjp: Oops good call

templates/server-ha-active-service.yaml
```diff
@@ -1,6 +1,6 @@
 {{ template "vault.mode" . }}
 {{- if ne .mode "external" }}
-{{- if and (eq .mode "ha" ) (and (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true")) }}
+{{- if and (eq .mode "ha" ) (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
```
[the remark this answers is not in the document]
tomhjp: Sure, raised #434

tomhjp:
Thanks for the quick review, @jasonodonnell. Ready for another round now.

templates/server-ha-lb.yaml
```diff
@@ -0,0 +1,35 @@
+{{ template "vault.mode" . }}
+{{- if and (eq .mode "ha" ) (eq (.Values.server.service.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") (eq (.Values.server.ha.lb.enabled | toString) "true") }}
+# LoadBalancer for active Vault pod
+apiVersion: v1
+kind: LoadBalancer
```
jasonodonnell, line 5: This should be `Service`

david-becher:
Is there any progress here? Cross-cluster replication needs a LoadBalancer, which exposes both port 8200 and 8201. I even think, this could/should be the default for the UI LoadBalancer. I would really like to see this chart support cross-cluster replication by exposing 8201 as well. Only being able to set up replication within the same K8s cluster somehow defeats the purpose of replication in the first place...

bkonick:
> Is there any progress here? Cross-cluster replication needs a LoadBalancer, which exposes both port 8200 and 8201. I even think, this could/should be the default for the UI LoadBalancer. I would really like to see this chart support cross-cluster replication by exposing 8201 as well. Only being able to set up replication within the same K8s cluster somehow defeats the purpose of replication in the first place...

@david-becher I actually have both 8200 and 8201 exposed by defining this in the values yaml

```yaml
server:
  service:
      annotations:
        service.beta.kubernetes.io/aws-load-balancer-internal: "true"
      type: "LoadBalancer"
```

That yields me three distinct LoadBalancer services all listening on 8200 and 8201:
```
NAME                     TYPE           CLUSTER-IP      EXTERNAL-IP                                                                        PORT(S)                         AGE
service/vault            LoadBalancer   10.x.x.x     blah1.elb.amazonaws.com   8200:32183/TCP,8201:31241/TCP   29d
service/vault-active     LoadBalancer   10.x.x.x   blah2.elb.amazonaws.com   8200:32663/TCP,8201:30630/TCP   29d
service/vault-internal   ClusterIP      None            <none>                                                                             8200/TCP,8201/TCP               29d
service/vault-standby    LoadBalancer   10.x.x.x    blah3.elb.amazonaws.com   8200:31777/TCP,8201:30724/TCP   29d
```

david-becher:
Hi @bkonick 

Thats correct and I have used this before, but it’s creating three distinct LoadBalancers when I actually only need one LB. Maybe I have not grasped the full concept of how to use all the different LBs, but having only one LB would reduce the costs imposed by cloud resources.

Maybe the chart should be adapted in a way where one can configure each of the services individually as LB or ClusterIP so that one can pick and choose the LB which is actually needed.

bkonick:
> Hi @bkonick
> 
> Thats correct and I have used this before, but it’s creating three distinct LoadBalancers when I actually only need one LB. Maybe I have not grasped the full concept of how to use all the different LBs, but having only one LB would reduce the costs imposed by cloud resources.
> 
> Maybe the chart should be adapted in a way where one can configure each of the services individually as LB or ClusterIP so that one can pick and choose the LB which is actually needed.

We use `vault` for client access as this includes all pods (active and perf standby). We use `vault-active` for replication as that traffic needs to be sent only to the active pod. We don't really use `vault-standby` though. We have used this pattern successfully with a replication chain that includes:
* NA <=> EU performance replication
* NA <=> second NA site DR replication
* EU <=> second EU site DR replication

I think at minimum both `vault` and `vault-active` are desired, unless you don't plan on using performance standby pods in which case you would just want `vault-active`.

I agree that having full control over which actually get provisioned would be helpful.

PR #434, opened 2020-12-14
Remove redundant logic

tomhjp:
Noticed these while raising #433. There are already a lot of brackets in these expressions, so it's nice to remove some for easier reading.


tomhjp:
Ping on this @jasonodonnell. Not urgent, but would be nice to either close or merge.

jasonodonnell (approved):

PR #435, opened 2020-12-15
Update server route template

josedev-union:
- Add more params for route tls config
- Use auto-generated host name when `.Values.server.route.host` empty,
like `{ROUTE_NAME}-{PROJECT_NAME}.{OCP_CLUSTER_APP_DOMAIN}`


josedev-union:
seems nobody is looking this PR

jasonodonnell:
Hi @josedev-union, thanks for the contribution. Sorry for the delay but we have a company shutdown for a weeks at the end of the year.

values.yaml
```diff
@@ -195,9 +195,17 @@ server:
...
+    labels: {}
+    # Host name for the Route
+    host: ""
+    tls:
+      # If `true`, TLS is enabled for the Route
+      enabled: true
+      # Insecure edge termination policy of the Route. Can be `None`, `Redirect`, or `Allow`
+      insecureEdgeTerminationPolicy: Redirect
+      # TLS termination of the route. Can be `edge`, `passthrough`, or `reencrypt`
+      termination: edge
```
jasonodonnell, line 208: It's not suggested to use edge or reencrypt TLS termination with Vault, the server should be handling all TLS termination (hence the passthrough).

cuttingedge1109: Well, I got your point.
But we can allow options for the other kinds of termination. The usecases can be various. For our case, we manage the TLS cert centrally and all tls are terminated on route level.

cuttingedge1109: @jasonodonnell Well, I got your point.
But we can allow options for the other kinds of termination. The usecases can be various. For our case, we manage the TLS cert centrally and all tls are terminated on route level.

rasheedamir: @jasonodonnell by any chance this PR will be accepted?

jasonodonnell: Yes, I think we can accept this, but we should make the default `passthrough`, since that's been the default for some time now.

cuttingedge1109:
@jasonodonnell I set `passthrough` as the default value.

rasheedamir:
@jasonodonnell can this PR be merged now?

hanzala1234:
any chance to get this PR merged?

mbaldessari:
This needs some rebasing now that https://github.com/hashicorp/vault-helm/commit/92da5125777b87b7475f546e0a7127b9517fe735 has merged

hsimon-hashicorp:
We won't be able to merge this without all contributors signing the CLA, unfortunately, too.

PR #436, opened 2020-12-17
Support deploying multiple injector replicas with auto-TLS

tomhjp:
This PR implements support for the leader election functionality added in https://github.com/hashicorp/vault-k8s/pull/198.

Acceptance test passing locally:

```bash
$ bats -t test/acceptance/injector-leader-elector.bats
1..1
ok 1 injector: testing leader elector
```

By default, no behaviour change other than some pod anti-affinity settings which I think are harmless, but please correct me if you think the default injector affinity settings should stay empty.

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -42,34 +42,6 @@ spec:
         runAsUser: {{ .Values.injector.uid | default 100 }}
       {{- end }}
       containers:
-        {{- if and (eq (.Values.injector.leaderElector.enabled | toString) "true") (gt (.Values.injector.replicas | int) 1) }}
-        - name: leader-elector
-          image: {{ .Values.injector.leaderElector.image }}
-          args:
-            - --election="{{ template "vault.fullname" . }}-agent-injector-leader"
-            - --election-namespace={{ .Release.Namespace }}
-            - --http=0.0.0.0:4040
-          livenessProbe:
-            httpGet:
-              path: /
-              port: 4040
-              scheme: HTTP
-            failureThreshold: 2
-            initialDelaySeconds: 1
-            periodSeconds: 2
-            successThreshold: 1
-            timeoutSeconds: 5
-          readinessProbe:
-            httpGet:
-              path: /
-              port: 4040
-              scheme: HTTP
-            failureThreshold: 2
-            initialDelaySeconds: 2
-            periodSeconds: 2
-            successThreshold: 1
-            timeoutSeconds: 5
-        {{- end }}
         - name: sidecar-injector
           {{ template "injector.resources" . }}
           image: "{{ .Values.injector.image.repository }}:{{ .Values.injector.image.tag }}"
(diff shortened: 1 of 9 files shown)


templates/injector-deployment.yaml
```diff
@@ -84,6 +84,14 @@ spec:
             - name: AGENT_INJECT_TELEMETRY_PATH
               value: "/metrics"
             {{- end }}
+            {{- if and (eq (.Values.injector.leaderElector.enabled | toString) "true") (gt (.Values.injector.replicas | int) 1) }}
```
jasonodonnell, line 87: I see this being used a few times, should we just make it a helper function?

templates/injector-leader-endpoint.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if and (eq (.Values.injector.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") (eq (.Values.injector.leaderElector.enabled | toString) "true") (gt (.Values.injector.replicas | int) 1) }}
+# This is created here so it can be cleaned up easily, since if
+# the endpoint is left around the leader won't expire for about a minute.
+apiVersion: v1
```
jasonodonnell, line 4: This needs Helm annotations:

```yaml
     app.kubernetes.io/name: {{ include "vault.name" . }}-agent-injector
```

values.yaml
```diff
@@ -27,6 +27,14 @@ injector:
   # True if you want to enable vault agent injection.
   enabled: true
 
+  replicas: 1
+
+  # If multiple replicas are specified, by default a leader-elector side-car
+  # will be created so that only one injector attempts to create TLS certificates.
+  leaderElector:
+    enabled: true
+    image: "gcr.io/google_containers/leader-elector:0.4"
```
jasonodonnell, line 36: For consistency, can you make this:

```
image:
  repository: "gcr.io/google_containers/leader-elector"
  tag: "0.4"
```

templates/injector-leader-endpoint.yaml
tomhjp: Good shout. Should I add the same to the secret too?

jasonodonnell: Yup, everything we make should have them.

tomhjp: Done

values.yaml
tomhjp: Done

tvoran (approved):
I agree with Jason's comment about a helper function for the template line, and we may also want to add an election ttl configurable option: https://github.com/hashicorp/vault-k8s/blob/master/deploy/injector-deployment.yaml#L29

But I'd say neither of those are blockers, and can be done in subsequent PRs if we like.

tomhjp:
I've added configurable TTL, and acceptance test is still passing locally:

```bash
bats -t test/acceptance/injector-leader-elector.bats
1..1
ok 1 injector: testing leader elector
```

tvoran (approved):

Issue #385, opened 2020-09-11
Feature request: Let's Encrypt for Vault itself


glerchundi:
**Is your feature request related to a problem? Please describe.**

In case Vault is secured end-to-end with self signed certificates (which seems to be the most common way of deploying it), anyone who is going to access to Vault needs to have the CA pubkey to verify the authenticity of the certificate & avoid MiTM attacks.

This requires to have a method to provision this CA in a secure way, which is not completely trivial. It can be done by putting it in a place where it is already trusted/verified like the company website: `https://mycompany.com/ca.pem`.

**Describe the solution you'd like**

Ideally I would like to avoid the people administrating Vault the hassle of trusting this CA. I don't know if it's even possible but here it is my proposal:

* Extenal load balancer (1.2.3.4:8200)
* Vault address listening to this external load balancer: `listener "tcp" { address "1.2.3.4:8200" }`. `api_addr` & `cluster_addr` would use internal kubernetes pod ip addresses.
* Vault also serving TLS in this address `listener "tcp" { tls_{cert,key}_file "/cert/from/letsencrypt-acme.{cert,key}" }`
* Use `cert-manager` and add support for automatic retrieval&renewal

**Describe alternatives you've considered**

Providing a custom CA cert out of band.

glerchundi:
Coming from https://github.com/hashicorp/vault/issues/9711.

fbongiovanni29:
Would really like to have this feature or maybe a guide on how to do this

alwaysastudent:
Would like to do this outside of K8s, on a baremetal standalone cluster.

Chili-Man:
this would be great; have any of you guys tried doing this but with [cert-manager](https://github.com/jetstack/cert-manager) to manage the Let's Encrypt secrets?

iuriaranda:
> this would be great; have any of you guys tried doing this but with [cert-manager](https://github.com/jetstack/cert-manager) to manage the Let's Encrypt secrets?

We're using Vault this way actually, and it works well. The only problem we have is that when cert-manager renews the certificate, we have no way of notifying Vault so it reloads it from the secret volume mount. We're looking for a clean solution to accomplish that atm. Any ideas?

glerchundi:
@iuriaranda would you mind adding more info.?

* What challenging method are you using?
* If it's not the DNS based one, how do you solve the proxying problem with HTTP/TLS, using an Envoy/Nginx in front of Vault?
[...]

PR #437, opened 2020-12-22
Add UI targetPort option

inigohu:
Use custom `targetPort` for UI service. See the usecase in https://github.com/hashicorp/vault-helm/issues/385#issuecomment-749560213


glerchundi:
Thanks for making it happen @inigohu!!

glerchundi:
Would it be possible to take a look at this @tvoran or @jasonodonnell?

PR #441, opened 2021-01-08
fix(injector): label component value in anti-affinity rule

rasta-rocket:
Hello folks 👋 

First of all, thanks a lot 👏 for the great work you did on the replicas topic related to the injector 💪 

Find that issue related to the label called "component" in the anti-affinity rule which doesn't match the one in the deployment manifest (webhook vs injector).

Also it could be an idea to rename all related label to injector. Tell me what you think about it. 

Cheers 😉


tvoran (approved):
Good catch! I think leaving it as "webhook" is ok for now.

PR #442, opened 2021-01-11
fix(injector): label component (app.kubernetes.io/name) value in anti-affinity rule

rasta-rocket:
Hey I missed that label also 😅 
Same as https://github.com/hashicorp/vault-helm/pull/441
Cheers


tvoran (approved):
Thanks!

PR #443, opened 2021-01-11
Tests: Increase the timeout for leader elector ready

tvoran:
Bumps the timeout waiting for the injector replicas (with
leader-elector containers) to be "Ready" to 5 minutes. Default was 30
seconds.

Hopefully this helps with intermittent tests failures with the leader-elector, for example: https://app.circleci.com/pipelines/github/hashicorp/vault-helm/599/workflows/f1a7b2ce-b5d0-41a5-923a-84cf79a10460/jobs/728


tomhjp (approved):
Thanks for fixing - seems totally reasonable 👍

PR #446, opened 2021-01-15
Set VAULT_DEV_LISTEN_ADDRESS in dev mode

tvoran:
Addresses https://github.com/hashicorp/vault-k8s/issues/206

Since vault in dev mode now binds to localhost (127.0.0.1) by default, this adds the environment variable VAULT_DEV_LISTEN_ADDRESS to the server statefulset so that vault will accept traffic from outside the pod.

Also adds a couple tests for the new var, and modifies one that was affected by the extra env var added to the pod spec.


jasonodonnell (approved):
LGTM!

PR #456, opened 2021-02-03
Add objectSelector to webhookconfiguration

guru1306:
When all pods in the k8s cluster are deleted in a namespace, some of the pods are starting without vault if the pods start before the vault injector  The failure policy can be set to fail in the webhook to force the deployments to be retried until the injector is available. But if we want the pod creation to fail only for certain pods then an object selector is needed. For instance, consider the vault and application pods are deployed in the same namespace with the failure policy set to fail. The Vault injector(webhook server) is also a pod in the same namespace that doesn't start as the webhook is not available. It’s a deadlock. Hence an object selector is needed to skip the webhook for certain pods.

Without this change, if the pods are deployed in the same namespace and the failure policy is set to fail, then entire deployment will not happen as explained above 

I have provided examples in values. yaml.

Once an object selector is set, only those pods with the label will go through the mutating webhook server.

### DIFF
--- a/templates/injector-mutating-webhook.yaml
+++ b/templates/injector-mutating-webhook.yaml
@@ -32,6 +32,10 @@ webhooks:
     namespaceSelector:
 {{ toYaml .Values.injector.namespaceSelector | indent 6}}
 {{ end }}
+{{- if .Values.injector.objectSelector }}
+    objectSelector:
+{{ toYaml .Values.injector.objectSelector | indent 6}}
+{{ end }}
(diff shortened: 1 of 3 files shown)


guru1306:
Could someone please take a look at the changes?

guru1306:
Nobody is looking at this PR. Can someone take a look at it?

guru1306:
@jasonodonnell  Could you take a look at this?

jasonodonnell (approved):
LGTM, thanks!

PR #458, opened 2021-02-09
add artifacthub ownership file

alvin-huang:
This PR adds an `artifacthub-repo.yml` file specifying a service user I created to claim ownership of the vault helm chart according to https://github.com/artifacthub/hub/blob/master/docs/repositories.md#ownership-claim. I will need another follow-up PR once the ownership is claimed to apply the `repositoryID` field as described in https://github.com/artifacthub/hub/blob/master/docs/repositories.md#verified-publisher.

### DIFF
--- a/artifacthub-repo.yml
+++ b/artifacthub-repo.yml
@@ -1,0 +1,3 @@
+owners:
+  - name: hashicorp
+    email: rel-eng+artifacthub@hashicorp.com


alvin-huang:
This file should exist on the repo index (helm.releases.hashicorp.com) and not the package index. This can be seen here: https://artifacthub.io/packages/search?page=1&org=hashicorp

PR #461, opened 2021-02-16
Add CSI secrets store provider

tomhjp:
Adds support for the Vault secrets store CSI provider.

```bash
$ bats test/acceptance/csi.bats 
 ✓ csi: testing deployment 

1 test, 0 failures
```


templates/csi-daemonset.yaml
```diff
@@ -0,0 +1,61 @@
+{{- if and (eq (.Values.csi.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
```
jasonodonnell, line 4: We should add the ability to apply annotations to these pods like we do for our other deployments.

templates/csi-daemonset.yaml
```diff
@@ -0,0 +1,61 @@
...
+  name: {{ template "vault.fullname" . }}-csi-provider
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "vault.name" . }}-csi-provider
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+spec:
+  updateStrategy:
+    type: RollingUpdate
```
jasonodonnell, line 13: We might want to make this configurable with RollingUpdate as the default.

templates/csi-serviceaccount.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if and (eq (.Values.csi.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
```
jasonodonnell, line 4: We should add the ability to apply annotations to the service account.

values.yaml
```diff
@@ -629,3 +629,40 @@ ui:
...
+  # True if you want to install a secrets-store-csi-driver-provider-vault daemonset.
+  #
+  # Requires installing the secrets-store-csi-driver separately, see:
+  # https://github.com/kubernetes-sigs/secrets-store-csi-driver#install-the-secrets-store-csi-driver
+  #
+  # With the driver and provider installed, you can mount Vault secrets into volumes
+  # similar to the Vault Agent injector, and you can also sync those secrets into
+  # Kubernetes secrets.
+  enabled: false
+
```
jasonodonnell, line 644: We should add an `externalVaultAddr` configurable like we have in the injector: https://www.vaultproject.io/docs/platform/k8s/helm/examples/external. Vault isn't always running on K8s (like Vault HCP) so we want to override any kind of templating configuration we do by default.

templates/csi-clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and (eq (.Values.csi.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") (semverCompare "> 0.0.7" .Values.csi.image.tag) }}
```
jasonodonnell, line 1: I don't think we're guaranteed the tag wasn't modified by the user, so this check might not be enough. I don't think there's much we can do about that, though, so let's just make sure to document these requirements on the website.

values.yaml
tomhjp: Weirdly, there is no configuration of the vault address in the provider - that all happens in the SecretProviderClass, so nothing to do here yet. But we probably will change that, at which point we'll want to add `externalVaultAddr` as you suggest.

tomhjp:
Thanks for the review @jasonodonnell  - all comments addressed, and acceptance test still passing:

```text
$ bats test/acceptance/csi.bats 
 ✓ csi: testing deployment 

1 test, 0 failures
```

templates/csi-daemonset.yaml
```diff
@@ -0,0 +1,67 @@
...
+kind: DaemonSet
+metadata:
+  name: {{ template "vault.fullname" . }}-csi-provider
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "vault.name" . }}-csi-provider
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+  {{ template "csi.daemonSet.annotations" . }}
+spec:
```
jasonodonnell, line 12: This daemonset is missing readiness/liveness probes and as such are always reporting healthy.

jasonodonnell (approved):
Missing health probes but LGTM and tested fine!

test/acceptance/csi-test/nginx.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: nginx
+---
+kind: Pod
+apiVersion: v1
+metadata:
+  name: nginx
+spec:
+  terminationGracePeriodSeconds: 0
+  containers:
+  - image: hashicorp.jfrog.io/docker/nginx
```
tvoran, line 14:
```suggestion
  - image: docker.mirror.hashicorp.services/docker/nginx
```

test/acceptance/csi.bats
```diff
@@ -0,0 +1,58 @@
...
+  result=$(kubectl --namespace=acceptance exec nginx -- cat /mnt/secrets-store/bar1)
+  [[ "$result" == "hello1" ]]
+}
+
+# Clean up
+teardown() {
+  if [[ ${CLEANUP:-true} == "true" ]]
+  then
+      echo "helm/pvc teardown"
+      helm delete vault
```
tvoran, line 54:
```suggestion
      helm --namespace=acceptance delete vault
```

test/acceptance/csi-test/nginx.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ServiceAccount
+metadata:
+  name: nginx
+---
+kind: Pod
+apiVersion: v1
+metadata:
+  name: nginx
+spec:
+  terminationGracePeriodSeconds: 0
```
tvoran, line 12:
```suggestion
  serviceAccountName: nginx
  terminationGracePeriodSeconds: 0
```

test/acceptance/csi-test/vault-kv-secretproviderclass.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  provider: vault
+  parameters:
+    roleName: "kv-role"
+    vaultAddress: http://vault:8200
+    objects: |
+      array:
+        - |
+          objectName: "bar1"
+          objectPath: "v1/secret/kv1"
```
tvoran, line 15:
```suggestion
      - objectName: "bar1"
        secretPath: "secret/data/kv1"
        secretKey: "bar1"
```

test/acceptance/csi.bats
```diff
@@ -0,0 +1,58 @@
...
+  kubectl --namespace=acceptance wait --for=condition=Ready --timeout=5m pod -l app.kubernetes.io/name=vault
+  kubectl --namespace=acceptance wait --for=condition=Ready --timeout=5m pod -l app.kubernetes.io/name=vault-csi-provider
+
+  # Set up k8s auth and a kv secret.
+  cat ./test/acceptance/csi-test/vault-policy.hcl | kubectl --namespace=acceptance exec -i vault-0 -- vault policy write kv-policy -
+  kubectl --namespace=acceptance exec vault-0 -- vault auth enable kubernetes
+  kubectl --namespace=acceptance exec vault-0 -- sh -c 'vault write auth/kubernetes/config \
+    token_reviewer_jwt="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+    kubernetes_host="https://$KUBERNETES_PORT_443_TCP_ADDR:443" \
+    kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt'
```
tvoran, line 33:
```suggestion
    kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
    disable_iss_validation=true'
```

test/acceptance/csi.bats
```diff
@@ -0,0 +1,58 @@
...
+  kubectl --namespace=acceptance exec vault-0 -- vault write auth/kubernetes/role/kv-role \
+    bound_service_account_names=vault-csi-provider \
```
tvoran, line 35:
```suggestion
    bound_service_account_names=nginx \
```

tvoran (commented):
These are changes I made to run the acceptance test against the current vault-csi-provider, in case that's what we want to target with the chart.

test/acceptance/csi-test/nginx.yaml
tomhjp: Thanks, good catch

values.yaml
```diff
@@ -685,5 +685,10 @@ csi:
     # annotations to apply to the serviceAccount.
     annotations: {}
 
+  # Used to define custom readinessProbe settings
+  readinessProbe: {}
+  # Used to enable a livenessProbe for the pods.
+  livenessProbe: {}
```
jasonodonnell, line 691: The defaults you defined up front should be in the `values.yaml` file. This makes them easier to discover by users instead of going to the website or looking at the source code.

tomhjp:
> These are changes I made to run the acceptance test against the current vault-csi-provider, in case that's what we want to target with the chart.

Thanks for these suggestions @tvoran, that's good to have a working patch - currently we are installing 0.0.7 by default, so that's the version I've targeted in the acceptance test. I think those changes should probably go hand-in-hand with updating the default once we have the next release to target. WDYT?

jasonodonnell (approved):

tvoran:
@tomhjp That sounds fine, though I think the probes will need to be disabled in order to use this chart with 0.0.7?

tomhjp:
> That sounds fine, though I think the daemonset probes will need to be disabled if this chart is going to default to 0.0.7?

Good point, I've put them behind a boolean flag that defaults to false. We can tear that flag out again in the near future.

Issue #344, opened 2020-06-29
NodePort value conflict between services


pamiel:
Dear community,

On chart v0.6.0, when setting the `service.type` to `NodePort` and specifying a `service.nodePort` value, then the following error is raised when trying to deploy using Helm:
```
Error: release vault failed, and has been uninstalled due to atomic being set: Service "vault" is invalid: spec.ports[0].nodePort: Invalid value: 32445: provided port is already allocated
```

This is due to the fact that the single `service.nodePort` value is used in both the `vault` (`server-service.yaml` file), `vault-active`(`server-ha-active-service.yaml` file) and `vault-standby` (`server-ha-standby-service.yaml` file) services.
We should rather have different `nodePort` values, one per service.

Note that if `service.type` is set to `NodePort` but no `service.nodePort` value is set, then things are working correctly and a random nodePort is allocated to each service.

sbaier1:
this is still the case and it's pretty inconvenient when trying to deploy vault in a HA setup with a static NodePort.

scottrbaxter:
> for example to link vault using an existing load balancer in front

Agreed, I feel like this would be a very common practice for vault clusters. Anyone with a HA cluster in Kubernetes that also needs their company to access the UI via browser are stuck on vault-helm <`v0.6.0`, or am I missing something here? What's the way forward for users deploying this chart that want to use an existing load balancer?

mikearruda:
Any update on this one?

shayfisher:
Just wondering - if one uses nodePort to access vault - the only service one should care about is vault-active isn't it?
I'm thinking on a PR to make this service use the supplied value and let the others select randomly.
Anyone has a different use-case where they access the standby by nodePort?

pamiel:
Hi @shayfisher ,
That was a good idea, but I think that the "vault-ui" service is also eligible to have a NodePort. So unfortunately, 2 Services ("vault-active" and "vault-ui") for one single nodePort value... :(

shayfisher:
@pamiel 
UI uses : Values.ui.serviceNodePort
While service uses: Values.server.service.nodePort
So one could set two different values and there will be no port collision.
This should work IMHO

pamiel:
Oh, yes sorry, my mistake  !! This is a different value for this nodePort of the UI !!

mikearruda:
> @pamiel
> UI uses : Values.ui.serviceNodePort
> While service uses: Values.server.service.nodePort
[...]

PR #463, opened 2021-02-26
* Make ha standby service generate random nodeport -

shayfisher:
In order to prevent collision with the ha active svc
* Make the generic server svc generate a random
nodeport if using HA mode, otherwise use the
value supplied by the user on the chart

This addresses this issue: 
https://github.com/hashicorp/vault-helm/issues/344

### DIFF
--- a/templates/server-ha-standby-service.yaml
+++ b/templates/server-ha-standby-service.yaml
@@ -27,7 +27,7 @@ spec:
       port: {{ .Values.server.service.port }}
       targetPort: {{ .Values.server.service.targetPort }}
       {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
-      nodePort: {{ .Values.server.service.nodePort }}
+      nodePort:
       {{- end }}
     - name: https-internal
       port: 8201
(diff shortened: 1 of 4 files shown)


shayfisher:
This PR should make the user able to pick the nodePort port for the active svc, while generating port values for other svc that are also being created as nodePort (standby and generic server nodeport)

shayfisher:
Hi @jasonodonnell,
Do you know if I could see the reason the build failed?
It has a generic error with nothing informative.
I'm not sure that this contribution is what the devs meant, but it solves the problem we are facing.
The other way around it is probably to be very verbose in the chart and let the user configure the type for each service individually.
Would love to hear how to make progress on this!
Thanks

shayfisher:
Hi @jasonodonnell ,
It does make more sense but I didn't know if you guys prefer to leave less choices for the end user in order to make the chart less verbose.
If I understand you correctly, a change should be made to introduce a new chart variable for the port of the standby service if NodePort was selected.
This way the user selects both the active and the standby ports.
Confirm and I will make the change.
Thanks

jasonodonnell:
@shahbazn Yes, I think a configurable on the port number makes sense.

shahbazn:
Hey @jasonodonnell maybe you wanted to tag @shayfisher in the last comment?

jasonodonnell:
My apologies @shahbazn, indeed I made a mistake 😄

swenson:
I believe this PR was superseded by https://github.com/hashicorp/vault-helm/pull/610

PR #464, opened 2021-03-01
Add externalTrafficPolicy support for services (excluding the UI service)

shayfisher:


test/unit/server-ha-standby-service.bats
```diff
@@ -168,3 +168,63 @@ load _helpers
...
+
+@test "server/ha-standby-Service: vault externalTrafficPolicy set to Local lowercase" {
+  cd `chart_dir`
+
+  local actual=$(helm template \
+      --show-only templates/server-ha-standby-service.yaml \
+      --set 'server.service.externalTrafficPolicy=local' \
+      --set 'server.ha.enabled=true' \
+      . | tee /dev/stderr |
+      yq -r '.spec.externalTrafficPolicy | tee /dev/stderr)
```
jasonodonnell, line 180: This is missing a `'` and so are the other tests you added here, which is causing failures when running the tests.

templates/server-ha-active-service.yaml
```diff
@@ -15,6 +15,11 @@ metadata:
   annotations:
 {{ template "vault.service.annotations" .}}
 spec:
+  {{- if and (.Values.server.service.externalTrafficPolicy ) (print .Values.server.service.externalTrafficPolicy | lower | title | eq "Local"  )}}
```
jasonodonnell, line 18: I believe this setting is only used when the service type is `LoadBalancer`, so an additional check might be needed.

shayfisher: This configuration is valid for NodePort and it allows you to get the sourceIP of the end user -
https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport
Thia also works when running with GKE as per the document I've referenced.
With this information in hand do you believe a change should be made?

test/unit/server-ha-standby-service.bats
shayfisher: Just committed the change from my phone :)

templates/server-ha-active-service.yaml
jasonodonnell: I think this still needs additional logic to check for the service type. For example the default is ClusterIP so this setting wouldn't be used by that. Thoughts?

shayfisher: I think that the simplest to implement addition will be to take care only the NodePort + externalTrafficPolicy=Local case
Just pushed the change

templates/server-ha-active-service.yaml
```diff
@@ -15,6 +15,11 @@ metadata:
   annotations:
 {{ template "vault.service.annotations" .}}
 spec:
+  {{- if and (.Values.server.service.externalTrafficPolicy ) (print .Values.server.service.externalTrafficPolicy | lower | title | eq "Local" ) (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
+  externalTrafficPolicy: {{ .Values.server.service.externalTrafficPolicy| lower | title | quote }}
+  {{- else }}
+  externalTrafficPolicy: {{ default "Cluster" | quote }}
+  {{- end}}
```
benashz, line 22: I feel like it may be simpler to document the possible traffic policies in `values.yaml` and then do:

```suggestion
  {{- if .Values.server.service.externalTrafficPolicy  }}
  externalTrafficPolicy: {{ .Values.server.service.externalTrafficPolicy }}
  {{- end }}
```

templates/server-service.yaml
```diff
@@ -15,6 +15,11 @@ metadata:
   annotations:
 {{ template "vault.service.annotations" .}}
 spec:
+  {{- if and (.Values.server.service.externalTrafficPolicy ) (print .Values.server.service.externalTrafficPolicy | lower | title | eq "Local"  )}}
+  externalTrafficPolicy: {{ .Values.server.service.externalTrafficPolicy | lower | title | quote }}
+  {{- else }}
+  externalTrafficPolicy: {{ default "Cluster" | quote }}
+  {{- end}}
```
benashz, line 22: I feel like it may be simpler to document the possible traffic policies in `values.yaml` and then do:

```suggestion
  {{- if .Values.server.service.externalTrafficPolicy  }}
  externalTrafficPolicy: {{ .Values.server.service.externalTrafficPolicy }}
  {{- end }}
```

values.yaml
```diff
@@ -170,6 +170,7 @@ injector:
 
   # Injector service specific config
   service:
+    externalTrafficPolicy: "Cluster"
```
benashz, line 173: Quotes on the value are not needed:

```suggestion
    # Set the Service's externalTrafficPolicy
    # For the set of possible value see: https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
    externalTrafficPolicy: Cluster
```

templates/server-ha-active-service.yaml
shayfisher: @benashz I thought to create a "fail-proof" deployment but I'm guessing that would be easier.
I would change this to the suggested value.

values.yaml
shayfisher: Got it, I will look for the second instance of this value and update accordingly while removing this configuration under the injector service.

templates/server-ha-standby-service.yaml
```diff
@@ -15,6 +15,11 @@ metadata:
   annotations:
 {{ template "vault.service.annotations" .}}
 spec:
+  {{- if and (.Values.server.service.externalTrafficPolicy ) (print .Values.server.service.externalTrafficPolicy | lower | title | eq "Local"  )}}
```
benashz, line 18: Seems to be a missing update here.

templates/server-ha-active-service.yaml
```diff
@@ -15,6 +15,9 @@ metadata:
   annotations:
 {{ template "vault.service.annotations" .}}
 spec:
+  {{- if .Values.server.service.externalTrafficPolicy }}
+  externalTrafficPolicy: {{ .Values.server.service.externalTrafficPolicy | lower | title }}
```
benashz, line 19: Are you certain that the extra template pipelines are necessary? I would say just take what is provided and let k8s sort it out.

test/unit/server-ha-active-service.bats
```diff
@@ -157,3 +157,53 @@ load _helpers
       yq -r '.spec.ports | map(select(.port==8200)) | .[] .name' | tee /dev/stderr)
   [ "${actual}" = "https" ]
 }
+
+@test "server/ha-active-Service: vault externalTrafficPolicy set to Local lowercase" {
```
benashz, line 161: I think just testing that the value provided is honoured is good enough  here.  The two cases are:
- empty string `""`
- non empty value like `Cluster`

values.yaml
```diff
@@ -395,6 +395,14 @@ server:
 
   # Enables a headless service to be used by the Vault Statefulset
   service:
+    # The externalTrafficPolicy can be set to either Cluster or Local
```
benashz, line 398: I think we can be little more succint here. The linked doc contains this information already.

templates/server-ha-active-service.yaml
benashz: Since we have multiple `Services` I think it makes sense to have `externalTrafficPolicy` configurable per.

values.yaml
shayfisher: I will just keep the options available and leave the reference.

test/unit/server-ha-active-service.bats
shayfisher: @benashz So you suggest adding a default value of "Cluster" for instance ? 
I'm not handling the case where an empty string isn't provided.
This would probably fall back to K8S defaults (Which currently I'm not sure what is).

templates/server-ha-active-service.yaml
shayfisher: @benashz I need to check what happens when one configures a service in K8S and uses a lower case value.
I suspect it might be failing to default in case of unknown value.
As per the services - you think this configuration might be necessary for the UI and the injector svcs also ? 
I'm trying not to overload the user with too many configurations.

test/unit/server-ha-active-service.bats
shayfisher: @benashz Just checked - made a new svc with a value not in the correct case -  got the following error:
The Service "my-service" is invalid: spec.externalTrafficPolicy: Invalid value: "local": ExternalTrafficPolicy must be empty, Cluster or Local.

So - empty is just fine.
Non empty value would fail if it isn't Local or Cluster.

templates/server-ha-active-service.yaml
shayfisher: Checked it as shown below - the pipelines are indeed necessary in case the user enters local/cluster instead of Local/Cluster.
It depends on how we believe the user will be able to cope with filling the values as expected by K8S.

benashz: I would prefer that we enforce that the user provides one of: `""`, `Cluster` or `Local`. This should be documented as the allowed values.

PR #466, opened 2021-03-09
Adds support to explicitly configure network policy in injector

ryanmt:
Similar to the conversion around PR #381, network policies are useful for the injector independent of openshift.  This allows support for those use cases but similarly will require some configuration changes for openshift users.


templates/injector-network-policy.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and (eq (.Values.injector.enabled | toString) "true" ) (and (eq (.Values.global.enabled | toString) "true") (eq (.Values.global.openshift | toString) "true") ) }}
+{{- if eq (.Values.injector.networkPolicy.enabled | toString) "true" }}
```
jasonodonnell, line 1: I believe this would create a networkPolicy even if you disabled the injector, which shouldn't happen.

jasonodonnell:
Bumping this 😄

ryanmt:
I ran into some issues I didn't end up chasing down yet.  I'll close this since I don't have immediately plans to revisit.

PR #475, opened 2021-03-24
Target vault-csi-provider release 0.1.0

tomhjp:
Now that `vault-csi-provider` 0.1.0 with liveness and readiness probes is released, target that by default.

```bash
bats test/acceptance/csi.bats 
 ✓ csi: testing deployment 

1 test, 0 failures
```

Also manually tested as working with some updates to the `vault-csi-provider`'s integration tests.


jasonodonnell (approved):
LGTM!

PR #482, opened 2021-03-29
Remove extra word in readme

coignetp:
The README seems to have a sentence with an extra word, so I propose to remove it.


tvoran (approved):
Thanks!

Issue #468, opened 2021-03-16
Upgrade the Ingress Api versions to accomodate the v1


rokarajan:
**Describe the bug**
Upon deploying vault using helm chart in kubernetes version 1.19 came across the message below so wnated to flag if you can accomodate the new api versions for ingress.

W0316 10:18:29.729348       1 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0316 10:18:29.936108       1 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress

**To Reproduce**
1. Install vault using helm chart in kubernets version 1.19

**Expected behavior**
As per this documentation: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
upgrade ingress with networking.k8s.io/v1 

**Environment**
* Kubernetes version: 
EKS 1.19

Chart version:
.0.9.1 app version 1.6.2
```yaml
# Be sure to scrub any sensitive values!
```

tony-kerz:
https://github.com/hashicorp/vault-helm/pull/483

tvoran:
I believe this has been addressed in #590, thanks.

PR #483, opened 2021-03-30
[ingress] allow api-version spec

tony-kerz:
```
helm template --debug -n ae-system --set server.ingress.enabled=1 my-release .
```

```yaml
---
# Source: vault/templates/server-ingress.yaml
apiVersion: networking.k8s.io/v1beta1


kind: Ingress
metadata:
  name: my-release-vault
  namespace: ae-system
  labels:
    helm.sh/chart: vault-0.10.0
    app.kubernetes.io/name: vault
    app.kubernetes.io/instance: my-release
spec:
  rules:
    - host: "chart-example.local"
      http:
        paths:
          - path: /
            backend:
              serviceName: my-release-vault
              servicePort: 8200
---
```

```
helm template --debug -n ae-system --set server.ingress.enabled=1 --set server.ingress.apiVersion=networking.k8s.io/v1 my-release .
```

```yaml
---
# Source: vault/templates/server-ingress.yaml
apiVersion: networking.k8s.io/v1

kind: Ingress
metadata:
  name: my-release-vault
  namespace: ae-system
  labels:
    helm.sh/chart: vault-0.10.0
    app.kubernetes.io/name: vault
    app.kubernetes.io/instance: my-release
spec:
  rules:
    - host: "chart-example.local"
      http:
        paths:
          - path: /
            backend:
              serviceName: my-release-vault
              servicePort: 8200
---
```

### DIFF
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -7,7 +7,11 @@
 {{- $serviceName = printf "%s-%s" $serviceName "active" -}}
 {{- end }}
 {{- $servicePort := .Values.server.service.port -}}
-{{ if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
+{{ if .Values.server.ingress.apiVersion }}
+apiVersion: {{ .Values.server.ingress.apiVersion }}
+{{ else if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+apiVersion: networking.k8s.io/v1
+{{ else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
 apiVersion: networking.k8s.io/v1beta1
 {{ else }}
 apiVersion: extensions/v1beta1
--- a/test/unit/server-ingress.bats
+++ b/test/unit/server-ingress.bats
@@ -121,3 +121,26 @@ load _helpers
       yq -r '.spec.rules[0].http.paths[0].backend.serviceName' | tee /dev/stderr)
   [ "${actual}" = "RELEASE-NAME-vault" ]
 }
+
+@test "server/ingress: api version capabilities default" {
+  cd `chart_dir`
+
+  local actual=$(helm template \
+      --show-only templates/server-ingress.yaml \
+      --set 'server.ingress.enabled=true' \
+      . | tee /dev/stderr |
+      yq -r '.apiVersion' | tee /dev/stderr)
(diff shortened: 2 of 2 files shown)


jasonodonnell (commented):
Hi @tony-kerz, thanks for the contribution!

Can you help me understand why this change is being introduced? We don't offer api version configurables on any other objects so I'm curious what issues you are running into?

Additionally for any PR with configurables to be merged, there must be unit test coverage. You can find the bats unit tests in `test/unit`, but please hold off until we understand the scope of this change more.

tony-kerz:
hi @jasonodonnell,

sure, yeah, the helm chart won't install on k8s v1.19.x, see [this issue](https://github.com/hashicorp/vault-helm/issues/468), is that sufficient motivation?

best,
tony.

jasonodonnell:
Hi @tony-kerz, thanks for pointing me to that issue. I've now linked it to this PR.

I'm very hesitant to allow users to configure this on their own. Instead I'm thinking this should use `if else` statements and lookup the capability of the server. The APIs don't change very often so we can expand when needed, but it would be one less thing for the user to configure.

Thoughts?

tony-kerz:
the capabilities lookup is cool, but u can see how it fell short this time around and now the chart is unusable in certain contexts until a new version is released. 

i've been bitten by the shifting api version situation numerous times in the past, and custom forks are always an extra hassle, so i might endorse a [belt-and-suspenders](https://www.merriam-webster.com/dictionary/belt-and-suspenders) approach which uses a "case" type thing with all known versions at a point in time, but also allows an override to keep the chart in play during periods of change.

jasonodonnell:
@tony-kerz, I'm fine with a switch case. Would you mind updating the PR to support it?

tony-kerz:
sure, will do, thanks jason!

on a side note, i tried to run [the `ingress` bats test](https://github.com/hashicorp/vault-helm/blob/master/test/unit/server-ingress.bats) and choked on the (undocumented?) dependency on [python `yq`](https://github.com/kislyuk/yq) because i myself have migrated to [the go version of `yq`](https://github.com/mikefarah/yq) to reduce the surface area of my runtimes.

any thoughts on cutting over to the go portable `yq`?

tony-kerz:
thx for looking at this @jasonodonnell i think #498 should be fine for this, let's just focus on merging that ;)

PR #488, opened 2021-03-31
Add logLevel and logFormat values for Vault

jasonodonnell:
This PR exposes two Vault logging configurations (log level and log format) via explicit server values. These were always configurable in the Vault configuration files, however, these values make them easier to change.

Fixes #487.

### DIFF
--- a/test/unit/server-statefulset.bats
+++ b/test/unit/server-statefulset.bats
@@ -414,6 +414,78 @@ load _helpers
 }
 
 #--------------------------------------------------------------------
+# log level
+
+@test "server/standalone-StatefulSet: default log level to info" {
+  cd `chart_dir`
+  local object=$(helm template \
+      --show-only templates/server-statefulset.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].env' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+     yq -r '.[11].name' | tee /dev/stderr)
+  [ "${actual}" = "VAULT_LOG_LEVEL" ]
+
+  local actual=$(echo $object |
+      yq -r '.[11].value' | tee /dev/stderr)
+  [ "${actual}" = "info" ]
+}
+
+@test "server/standalone-StatefulSet: log level can be changed" {
+  cd `chart_dir`
+  local object=$(helm template \
+      --show-only templates/server-statefulset.yaml  \
+      --set='server.logLevel=debug' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].env' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+     yq -r '.[11].name' | tee /dev/stderr)
+  [ "${actual}" = "VAULT_LOG_LEVEL" ]
+
+  local actual=$(echo $object |
+      yq -r '.[11].value' | tee /dev/stderr)
+  [ "${actual}" = "debug" ]
+}
+
+#--------------------------------------------------------------------
+# log format
+
+@test "server/standalone-StatefulSet: default log format to standard" {
+  cd `chart_dir`
+  local object=$(helm template \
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -188,6 +188,14 @@ server:
   # See https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies
   updateStrategyType: "OnDelete"
 
+  # Configure the logging verbosity for the Vault server.
+  # Supported log levels include: trace, debug, error, warn, info
+  logLevel: "info"
```
tomhjp, line 193: In order of precedence, Vault prefers `Flags > Env var > Config file`[1], so defaulting this to "info" may surprise some folks who have log level set in their config file. Same goes for logFormat of course.

[1]: https://www.vaultproject.io/docs/configuration#log_level

values.yaml
```diff
@@ -188,6 +188,14 @@ server:
   # See https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies
   updateStrategyType: "OnDelete"
 
+  # Configure the logging verbosity for the Vault server.
+  # Supported log levels include: trace, debug, error, warn, info
```
tomhjp, line 192: Can you put these in order?

tomhjp (commented):
I'm ok with this approach too, just needs some tests 👍

values.yaml
tomhjp: Error is the least verbose, then warn then info, then debug and trace is the most verbose.

tvoran: I kinda like the idea of defaulting to empty, and only setting the logLevel env var if this has been set by the user, since this is something that could also be set in the vault config file.

values.yaml
```diff
@@ -188,6 +188,14 @@ server:
...
+  logLevel: "info"
+
+  # Configure the logging format for the Vault server.
+  # Supported log formats include: standard, json
+  logFormat: "standard"
```
tvoran, line 197: Similar thought here, could this default to `""`, and then only set VAULT_LOG_FORMAT env if values.server.logFormat has been set by the user?

values.yaml
ncabatoff: Shouldn't `error` be last?

tomhjp: +1, in descending order of verbosity it should be trace, debug, info, warn, error

values.yaml
jasonodonnell: Good point @tvoran, I added that.

values.yaml
jasonodonnell: Fixed.

values.yaml
jasonodonnell: Fixed as per above.

test/unit/server-statefulset.bats
```diff
@@ -413,6 +413,62 @@ load _helpers
   [ "${actual}" = "true" ]
 }
 
+#--------------------------------------------------------------------
+# log level
+
+@test "server/standalone-StatefulSet: default log level to info" {
```
tomhjp, line 419: nit:

```suggestion
@test "server/standalone-StatefulSet: default log level to empty" {
```

test/unit/server-statefulset.bats
```diff
@@ -413,6 +413,62 @@ load _helpers
...
+
+  local value=$(echo $objects |
+      yq -r 'map(select(.name=="VAULT_LOG_LEVEL")) | .[] .value' | tee /dev/stderr)
+  [ "${value}" = "debug" ]
+}
+
+#--------------------------------------------------------------------
+# log format
+
+@test "server/standalone-StatefulSet: default log format to standard" {
```
tomhjp, line 447: nit:

```suggestion
@test "server/standalone-StatefulSet: default log format to empty" {
```

tomhjp (approved):
LGTM

PR #489, opened 2021-04-04
Custom value of agent port 

HamzaZo:
This PR allows users to set the different port of `AGENT_INJECT_LISTEN`, other than the default one.
Ref [issue](https://github.com/hashicorp/vault-k8s/issues/240)

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -52,7 +52,7 @@ spec:
           {{- end }}
           env:
             - name: AGENT_INJECT_LISTEN
-              value: ":8080"
+              value: {{ printf ":%v" .Values.injector.port | default ":8080" }}
             - name: AGENT_INJECT_LOG_LEVEL
               value: {{ .Values.injector.logLevel | default "info" }}
             - name: AGENT_INJECT_VAULT_ADDR
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -103,7 +103,7 @@ spec:
           livenessProbe:
             httpGet:
               path: /health/ready
-              port: 8080
+              port: {{ .Values.injector.port }}
               scheme: HTTPS
             failureThreshold: 2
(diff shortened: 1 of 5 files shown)


jasonodonnell (commented):
Hi @HamzaZo, thanks for the contribution!

This looks good but is missing a few things:

* The health probes need to be updated with the custom port,
* Bats unit tests need to be added to ensure coverage on this new configurable. You can find those in `test/unit`.

Thanks again!

HamzaZo:
Hello @jasonodonnell, thanks for the review.

I added the missing configs 

Thanks!

templates/injector-deployment.yaml
```diff
@@ -113,7 +113,7 @@ spec:
           readinessProbe:
             httpGet:
               path: /health/ready
-              port: 8080
+              port: {{ printf ":%v" .Values.injector.port | default ":8080" }}
```
jasonodonnell, line 116: The `| default ":8080"` is actually not needed since the default is in the `values.yaml` file. Can you remove it?

jasonodonnell (commented):
Thanks @HamzaZo! It looks like the service will also need unit tests since those values are effected by this configurable as well.

templates/injector-deployment.yaml
HamzaZo: sorry for the copy/past mistake, done

HamzaZo:
@jasonodonnell I update the configuration, could you check it? Thanks

values.yaml
```diff
@@ -29,6 +29,9 @@ injector:
 
   replicas: 1
 
+  # Configures the port the agent should listen on
```
jasonodonnell, line 32:
```suggestion
  # Configures the port the injector should listen on
```

HamzaZo:
Hi @jasonodonnell,
Can you check if everything looks good?
thanks

jasonodonnell (approved):
LGTM, thanks!

PR #495, opened 2021-04-13
[injector] Add port name in injector service

javiercri:
- Add port name in injector service.

### DIFF
--- a/templates/injector-service.yaml
+++ b/templates/injector-service.yaml
@@ -11,7 +11,8 @@ metadata:
   {{ template "injector.service.annotations" . }}
 spec:
   ports:
-  - port: 443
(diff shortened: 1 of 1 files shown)


templates/injector-service.yaml
```diff
@@ -11,7 +11,8 @@ metadata:
   {{ template "injector.service.annotations" . }}
 spec:
   ports:
-  - port: 443
+  - name: {{ include "vault.scheme" . }}
```
jasonodonnell, line 14: The injector is always HTTPs and isn't based on `vault.scheme`. So this can be hard coded to `https`.

javiercri: Done

jasonodonnell (approved):
LGTM, thanks!

PR #498, opened 2021-04-15
feat: Allow networking.k8s.io/v1

dirtycajunrice:
allows for the newest networking stable api

### DIFF
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -8,7 +8,8 @@
 {{- $serviceName = printf "%s-%s" $serviceName "active" -}}
 {{- end -}}
 {{- $servicePort := .Values.server.service.port -}}
-{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" -}}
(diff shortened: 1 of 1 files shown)


philomory:
I would love to see this merged and released, is that likely to happen soon?

DirtyCajunRice:
@philomory Was actually 1 small bug. Fixed now... but not sure who to ping to move it along.

Kerwood:
Can we get this one going ? It is much needed.

The unit test fail because the test is using `v1beta1` logic on a `v1` manifest.
This is the unit test that fails: https://github.com/hashicorp/vault-helm/blob/master/test/unit/server-ingress.bats#L46

As you can see in below screenshot from the failed test, the api version is  `v1beta1` but the `paths` array is `v1`. Therefore, the test `yq  -r '.spec.rules[0].http.paths[0].backend.serviceName  | length > 0'` will fail.
image


I am not sure how the unit test distinguishes between the different versions or if it is possible ?

tvoran:
Thanks for the work here! It looks like this has been addressed in https://github.com/hashicorp/vault-helm/pull/310 and https://github.com/hashicorp/vault-helm/pull/634, so I'll close this for now.

PR #500, opened 2021-04-19
Added dns config into server statefulset

awilkins-frel:
This allows you to add dnsConfig, I need it because it allows me to add in my own search domain to the vault servers.


awilkins-frel:
Any chance of getting this included ? Not being able to add a dns search domain is still holding me back from being able to use the official repo for this

PR #508, opened 2021-04-27
Fixing condition evaluation on `openshift` value

scellef:
Removing errant `not` in `openshift: true` condition evaluation.


jasonodonnell:
Hi @scellef, this is not a bug, but intended. In OpenShift, pod specs are mutated with security contexts (randomized UID, GID 0 and other rules).

If you're seeing issues specifically due to security contexts, let me know. I've seen issues where storage classes don't properly setup permissions for volumes (such as vsphere CSI), but those can be fixed via additional storage class configurations.

Let me know!

PR #510, opened 2021-05-07
Set minkubeversion and added chart-verifier tests

tvoran:
Sets min kubeVersion in Chart.yaml to 1.14. Added [chart-verifier](https://github.com/redhat-certification/chart-verifier) bats tests, and configured to run them in CI alongside the unit tests. Some verification tests that haven't been addressed yet are skipped.

I chose 1.14 since it's the oldest k8s version that kind can run with the [last release](https://github.com/kubernetes-sigs/kind/releases/tag/v0.10.0), and therefore I was able to run the acceptance tests with it (besides csi), but that's certainly open for suggestions.

The chart-verifier circleci job builds the test binary from a specific git SHA, since the project is under active development and neither the [git repo](https://github.com/redhat-certification/chart-verifier) nor the [docker image](https://quay.io/repository/redhat-certification/chart-verifier?tab=tags) has release tags.

### DIFF
--- a/test/chart/verifier.bats
+++ b/test/chart/verifier.bats
@@ -1,0 +1,77 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+setup_file() {
+    cd `chart_dir`
+    export VERIFY_OUTPUT="/$BATS_RUN_TMPDIR/verify.json"
+    export CHART_VOLUME=vault-helm-chart-src
+    # Note: currently `latest` is the only tag available in the chart-verifier repo.
+    local IMAGE="quay.io/redhat-certification/chart-verifier:latest"
+
+    local run_cmd="chart-verifier"
+    local chart_src="."
+
+    if [ ! -e $USE_DOCKER ]; then
+        chart_src="/chart"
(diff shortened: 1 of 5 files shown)


test/chart/verifier.bats
```diff
@@ -0,0 +1,77 @@
...
+@test "is-helm-v3" {
+    check_result is-helm-v3
+}
+
+@test "not-contains-crds" {
+    check_result not-contains-crds
+}
+
+@test "helm-lint" {
+    check_result not-contains-crds
```
tomhjp, line 49: Looks like a few of these `check_result` arguments need updating

tomhjp (approved):
LGTM, and great tests! Presumably the intention is that for running locally we can use docker, but CI uses the installed version?

test/chart/verifier.bats
tvoran: Ha, whoops. Fixed in e9c7517.

tvoran:
@tomhjp Yeah, for running locally we can either use an installed version or docker. CI is only using an installed version.

PR #512, opened 2021-05-11
match kubeVersion on semver pre-releases

tvoran:
Since clouds like GKE set their kubeVersion as a pre-release (e.g. v1.17.17-gke.6700).

See https://github.com/helm/helm/issues/3810 for details.


tomhjp (approved):
Nice catch finding this before it was released

jasonodonnell (approved):

PR #513, opened 2021-05-11
added values json schema

tvoran:
Generated the schema using the helm [schema-gen plugin](https://github.com/karuppiah7890/helm-schema-gen), and added extra data types to fields that allow it, such as annotations, tolerations, enabled, etc. Also enables the "contains-value-schema" chart-verifier test. Added a blurb to the test readme about running the chart-verifier tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+Improvements:
+* Set chart kubeVersion and added chart-verifier tests [GH-510](https://github.com/hashicorp/vault-helm/pull/510)
+
 ## 0.11.0 (April 14th, 2021)
 
 Features:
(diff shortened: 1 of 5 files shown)


jasonodonnell (approved):
LGTM!

tvoran:
> And am I right thinking the schema is automatically exercised by helm as we run through the unit/acceptance tests?

Yeah, since our tests use `helm template`, they'll use the schema for validation. It looks like the helm template, lint, install, and upgrade commands all use the schema: https://helm.sh/docs/topics/charts/#schema-files

And the chart-verifier also runs `helm lint`, so I think that's another instance of schema validation in our tests.

tomhjp (approved):
Thanks for adding the Makefile target and docs. LGTM 👍

PR #517, opened 2021-05-11
Issue-516 ImagePullSecret Fix for csi-deamonset

mehmetsalgar:
Fix for the missing ImagePullSecret configuration at csi-deamonset.yaml

### DIFF
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -75,6 +75,6 @@ spec:
        {{- end }}
       {{- if .Values.global.imagePullSecrets }}
       imagePullSecrets:
-      {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
-  {{- end }}
+        {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
+      {{- end }}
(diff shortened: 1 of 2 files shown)


templates/csi-daemonset.yaml
```diff
@@ -73,4 +73,8 @@ spec:
        {{- if .Values.csi.volumes }}
          {{- toYaml .Values.csi.volumes | nindent 8}}
        {{- end }}
+      {{- if .Values.global.imagePullSecrets }}
+      imagePullSecrets:
+      {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
+  {{- end }}
```
tomhjp, line 79: The indentation just looks a bit off here:

```suggestion
      {{- if .Values.global.imagePullSecrets }}
      imagePullSecrets:
        {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
      {{- end }}
```

tomhjp (commented):
Thanks for spotting and raising this. LGTM, it would be good to get some unit tests similar to those on the server statefulset please, and then I think this will be ready to land.

PR #521, opened 2021-05-12
[Issue-520] tolerations for csi-deamonset

mehmetsalgar:
Missing toleration config at DeamonSet

### DIFF
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -28,6 +28,7 @@ spec:
       {{ template "csi.pod.annotations" . }}
     spec:
       serviceAccountName: {{ template "vault.fullname" . }}-csi-provider
+      {{- template "csi.daemonSet.tolerations" . }}
       containers:
         - name: {{ include "vault.name" . }}-csi-provider
           {{ template "csi.resources" . }}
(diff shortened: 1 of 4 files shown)


values.yaml
```diff
@@ -728,6 +728,7 @@ csi:
     # YAML-formatted multi-line templated string map of the annotations to apply
     # to the daemonSet.
     annotations: {}
+    tolerations: null
```
tvoran, line 731: I think this could use a comment block describing format, etc. Something like this?

```suggestion

    # Toleration Settings for provider pods
    # This should be a multi-line string matching the Toleration array
    # in a PodSpec.
    tolerations: null

```

templates/csi-daemonset.yaml
```diff
@@ -28,6 +28,7 @@ spec:
       {{ template "csi.pod.annotations" . }}
     spec:
       serviceAccountName: {{ template "vault.fullname" . }}-csi-provider
+      {{- template "csi.daemonSet.tolerations" . }}
```
tvoran, line 31: Since tolerations are set here in the pod spec portion of the daemonSet, I'm thinking this might fit better in the `csi.pod` section.

```suggestion
      {{- template "csi.pod.tolerations" . }}
```

tvoran (commented):
Looks pretty good, though should these should be defined under `csi.pod` instead of `csi.daemonSet`?

tvoran (approved):
Thanks!

Issue #249, opened 2020-04-06
Need ability to resuse existing Cluster Role and Cluster Role Binding


kaishah:
When I create multiple Vault deployments on a single cluster, the helm chart is unable to reuse the existing Cluster Roles and Cluster Role Bindings.

UPGRADE FAILED: rendered manifests contain a new resource that already exists. Unable to continue with update: existing resource conflict: namespace: , name: vault-agent-injector-binding, existing_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding, new_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding

ngarafol:
You can just disable auth delegator in server options:

`  server:$
       authDelegator:$
          enabled: false$
`
Check indentation but its server->authdelegator->enabled

kaishah:
I want to use the authDelegator with an existing ClusterRole and the helm chart shouldn't fail if the ClusterRole already exists

ngarafol:
Problem lies in templates/server-clusterrolebinding.yaml IIRC. As I said you could try to disable authdelegator and adjust values.yaml accordingly so it uses clusterrole you want. It should work like this.

kaishah:
The chart might be used in both cases, if the ClusterRole exists or not. It can reuse a ClusterRole if it exists, and not throw an error, or create one if it doesn't exist.

I don't want to edit any templates, ideally only the values, so that we can pull any updates to the chart from this repository without conflicts.

manisha-tanwar:
Hi @kaishah Were you able to use existing CRD? In my case I don't have access to create CRD and can get it from Admin team but don't know what role should I ask them to create and how to use them here. Any suggestions?

iomarcovalente:
I believe this to be related to https://github.com/helm/helm/pull/7649 
I fixed it by annotating relevant resources with command:
`kubectl annotate $KIND $NAME meta.helm.sh/release-namespace=hcvault`

astenman:
@kaishah : Any news on this? I just encountered the same problem. We have a production vault running in namespace "vault" and I want to make an additional installation in namespace "vault-dev" for testing.

anapsix:
This issues also comes up when Vault chart is used as subchart.
[...]

PR #523, opened 2021-05-17
use namespace in the name of cluster-wide resource

anapsix:
.. to avoid resource name collision without having to override full name,
  or change·release name, or use other name overrides

Signed-off-by: Anastas Dancha <anapsix@random.io>

first solution I've mentioned in #249


sig-abyreddy:
@hashicorp-cla Can we merge this feature?

swenson:
Thanks for this.

I understand that this works around the problem you mentioned (but doesn't solve it by allowing reuse of the cluster resources?), but it doesn't make sense to me to include the namespace in a resource that is explicitly not in a namespace.

PR #531, opened 2021-05-25
Adding helm test for vault server

tvoran:
Also adds acceptance test for 'helm test' and updates the chart-verifier version.

### DIFF
--- a/templates/tests/server-test.yaml
+++ b/templates/tests/server-test.yaml
@@ -9,7 +9,7 @@ metadata:
 spec:
   containers:
     - name: {{ .Release.Name }}-server-test
-      image: {{ .Values.server.image.repository }}:{{ .Values.server.image.tag }}
+      image: {{ .Values.server.image.repository }}:{{ .Values.server.image.tag | default "latest" }}
       imagePullPolicy: {{ .Values.server.image.pullPolicy }}
       env:
         - name: VAULT_ADDR
--- a/templates/tests/server-test.yaml
+++ b/templates/tests/server-test.yaml
@@ -9,8 +9,8 @@ metadata:
 spec:
   containers:
     - name: {{ .Release.Name }}-server-test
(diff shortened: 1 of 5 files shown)


test/chart/verifier.bats
```diff
@@ -33,8 +39,8 @@ teardown_file() {
     fi
 }
 
-@test "has-minkubeversion" {
-    check_result has-minkubeversion
+@test "has-kubeversion" {
+    check_result has-kubeversion
```
tvoran, line 43: This change is because the test was renamed in the latest `chart-verifier`.

templates/tests/server-test.yaml
```diff
@@ -0,0 +1,39 @@
...
+kind: Pod
+metadata:
+  name: "{{ .Release.Name }}-server-test"
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: {{ .Release.Name }}-server-test
+      image: {{ .Values.server.image.repository }}:{{ .Values.server.image.tag | default "latest" }}
```
tomhjp, line 12: I'm surprised to see the `| default "latest"` snippet here. Given we package the chart with a default tag in `values.yaml`, is it possible for tag to not be set? Unless people really go out of their way I suppose..

tomhjp (approved):
Nice! LGTM

templates/tests/server-test.yaml
tvoran: Yeah, good catch, removed it in a498b29.

tvoran: Well I take it back. It looks like since the test is templated as part of the chart install, there's a test that fails if the tag doesn't default to "latest" like this: https://github.com/hashicorp/vault-helm/blob/ec67b5dd45c60346f709630e0d9acbd6f538ac69/test/unit/server-dev-statefulset.bats#L49-L60

So put it back for now.

Issue #534, opened 2021-05-26
server.ha.disruptionBudget.maxUnavailable: Invalid type. Expected: `null`, given: `integer`


rule88:
https://github.com/hashicorp/vault-helm/blob/b21b37b07a439e8c0ae6ddd133aca71c719a01ca/values.schema.json#L487

This change prevents upgrading from 0.11.0 to 0.12.0 when using variable `server.ha.disruptionBudget.maxUnavailable`.

rule88:
PR: https://github.com/hashicorp/vault-helm/pull/535

rule88:
Error message:

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
│ vault:
│ - server.ha.disruptionBudget.maxUnavailable: Invalid type. Expected: null, given: integer
```

tvoran:
Fixed in #535

PR #535, opened 2021-05-26
change maxUnavailable to integer

rule88:
change maxUnavailable from `null` to `integer` to enable upgrade from 0.11.0 to 0.12.0 when using the specific variable.

solves:

- https://github.com/hashicorp/vault-helm/issues/534

### DIFF
--- a/values.schema.json
+++ b/values.schema.json
@@ -484,7 +484,10 @@
                                     "type": "boolean"
                                 },
                                 "maxUnavailable": {
-                                    "type": "integer"
+                                    "type": [
+                                        "null",
+                                        "integer"
+                                    ]
                                 }
                             }
                         },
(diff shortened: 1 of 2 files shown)


values.schema.json
```diff
@@ -484,7 +484,7 @@
                                     "type": "boolean"
                                 },
                                 "maxUnavailable": {
-                                    "type": "null"
+                                    "type": "integer"
```
tvoran, line 487: I think this needs to also allow for a type of 'null' since that's the default in values.yaml.

```suggestion
                                    "type": [
                                        "null",
                                        "integer"
                                    ]
```

tvoran (changes requested):
Good catch! Can you also add a test in [test/unit/server-ha-disruptionbudget.bats](https://github.com/hashicorp/vault-helm/blob/2b2b0dd2fa9e871d7525e35c7eff5a518bbf9c22/test/unit/server-ha-disruptionbudget.bats) that sets server.ha.disruptionBudget.maxUnavailable to a number? I think that would've caught this.

tvoran (approved):

PR #536, opened 2021-05-27
process extraInitContainers and extraContainers with tpl

anapsix:
For more flexibility when using this chart as sub-chart, process `server.extraInitContainers` and `server.extraContainer` with `tpl`.

Signed-off-by: Anastas Dancha <anapsix@random.io>

Sub-chart's `values.yaml` example:
```yaml
vault:
  server:
    extraContainers:
    - name: my-vault-sidecar
      image: >-
        {{- $sidecarImageRepo := default "my-vault-sidecar-image-repo" .Values.global.myVaultSidecar.image.repo -}}
        {{- $sidecarImageTag := default "my-vault-sidecar-image-tag" .Values.global.myVaultSidecar.image.tag -}}
        {{- printf "%s:%s" $sidecarImageRepo $sidecarImageTag }}
      imagePullPolicy: "{{ .Values.global.myVaultSidecar.image.pullPolicy }}"
...
```


anapsix:
I don't understand why `bats-unit-test` is failing. It's succeeding on my local when running `make test-image test-unit`.
I even tried rebuilding the `vault-helm-test` image with `--no-cache`, in case I was using something out of date.

anapsix:
interesting.. 
tried running unit tests using `docker.mirror.hashicorp.services/hashicorpdev/vault-helm-test@sha256:62b548d0b1bc07c2ec37133c60a441414c62fe3ee34692588f0c2c56be5d177e` image (the one used by CI job), and got the same failure as in CI job..
Digging further into it, I've ran the `helm template` command corresponding to the new tests cases I've added, and noticed Helm was unable to render the template.

That docker image has Helm CLI `v3.2.0`
However, when I upgraded it to `v3.6.0`, tests succeeded.

Pretty sure the test image is using a version of Helm CLI which has some bug preventing it from rendering YAML related to the `tpl` I've added.

anapsix:
I believe there is a bug in Helm version used in CI test image running "bats" test, as described in #539

PR #538, opened 2021-05-28
Add OpenShift default images

tvoran:
Adds OpenShift UBI images as defaults if `global.openshift=true` or the `apps.openshift.io/v1` API is present in the deployment system. Otherwise the defaults are the images from dockerhub. The images can also be set by user overrides in either case.

Also sets `global.openshift=true` for the chart-verifier tests (since the helm setup in chart-verifier [doesn't enable any openshift APIs](https://github.com/redhat-certification/chart-verifier/blob/e2c03bd1a4aea20deb0a4a03ebfde254b1672050/pkg/chartverifier/checks/helm.go#L197)), and checks that the `images-are-certified` chart-verifier test passed.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -63,7 +63,7 @@ injector:
     # OpenShift is detected)
     openshift:
       repository: "registry.connect.redhat.com/hashicorp/vault-k8s"
-      tag: "0.10.0-ubi"
+      tag: "0.10.1-ubi"
 
   # agentImage sets the repo and tag of the Vault image to use for the Vault Agent
   # containers.  This should be set to the official Vault image.  Vault 1.3.1+ is
--- a/values.yaml
+++ b/values.yaml
@@ -51,16 +51,36 @@ injector:
 
   # image sets the repo and tag of the vault-k8s image to use for the injector.
   image:
-    repository: "hashicorp/vault-k8s"
-    tag: "0.10.0"
+    repository: ""
+    tag: ""
     pullPolicy: IfNotPresent
+    # Kubernetes defaults (these will be used if global.openshift=false and
+    # OpenShift is not detected)
+    kubernetes:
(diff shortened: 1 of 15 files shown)


values.yaml
```diff
@@ -203,10 +223,20 @@ server:
...
   image:
-    repository: "vault"
-    tag: "1.7.2"
+    repository: ""
+    tag: ""
     # Overrides the default Image Pull Policy
     pullPolicy: IfNotPresent
+    # Kubernetes defaults (these will be used if global.openshift=false and
+    # OpenShift is not detected)
+    kubernetes:
```
tomhjp, line 232: Would it be possible to keep the non-openshift config keys in the same place as before? Having `kubernetes` here feels a little odd to me, and I think it could cause confusion for those not familiar with openshift and and not sure whether it's relevant to them.

I assume you've done it this way so if a user specifies `server.image.repository` but deploys to openshift, their value doesn't get ignored? But IMHO I think that would be ok - it seems reasonable that in the default case we always use `server.image.repository`, unless you're deploying to openshift in which case we always use `server.image.openshift.repository`.

tomhjp (commented):
LGTM, just one discussion point about how we'd like to handle defaults and `values.yaml` config. I'd be interested to hear more on your and other people's thoughts on that.

values.yaml
tvoran: Yeah I structured it like this because there doesn't seem to be a way to tell if a user has overridden the default for a value. So this way preserves the existing behavior of using `server.image.repository` if a user has it in their chart deployment overrides. I was err'ing on the side of caution since folks are already using the chart on openshift.

tomhjp: Good point, I hadn't thought about that upgrade flow 🤔

tvoran:
closing in favor of #549

PR #545, opened 2021-06-04
Added webhook-certs volume mount to sidecar injector

rmgpinto:
This PR adds the `webhook-certs` volume mount to sidecar injector.
Without it, the agent injector can't access the TLS certificate files.

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -167,10 +167,6 @@ spec:
             timeoutSeconds: 5
         {{- end }}
 {{- if .Values.injector.certs.secretName }}
-          volumeMounts:
-            - name: webhook-certs
-              mountPath: /etc/webhook/certs
-              readOnly: true
       volumes:
         - name: webhook-certs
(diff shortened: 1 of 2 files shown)


rmgpinto:
@tvoran or @tomhjp ?

tvoran (changes requested):
Good catch!

Can you also remove the volumeMount block from lines 170-173? It doesn't belong in the leader-elector container, and with injector.replicas=1 two volumeMount blocks will be rendered.

Would you also mind updating the [bats test](https://github.com/hashicorp/vault-helm/blob/5f2e1d470fc1f04972232f90c43883873bae0485/test/unit/injector-deployment.bats#L126-L126) to check for the webhook-certs mount? This test is an example: https://github.com/hashicorp/vault-helm/blob/030d3cd89db0e6d99ba5038bd12e2d0153112586/test/unit/csi-daemonset.bats#L283-L303

tvoran (approved):
Thanks!

PR #546, opened 2021-06-04
add configuration for active/standy services for server

jamesgoodhouse:
Currently there is no ability to configure services for active and standby services, as they inherit their settings from the main server service configuration. This is a problem, for example, when you need to attach different annotations to the different services. Instead of just allowing for setting of annotations, I opted to allow configuring the services for more flexibility.

### DIFF
--- a/values.schema.json
+++ b/values.schema.json
@@ -685,6 +685,28 @@
                         },
                         "targetPort": {
                             "type": "integer"
+                        },
+                        "active": {
+                            "type": "object",
+                            "properties": {
+                                "annotations": {
+                                    "type": [
+                                        "object",
+                                        "string"
+                                    ]
+                                }
+                            }
+                        },
+                        "standby": {
+                            "type": "object",
+                            "properties": {
+                                "annotations": {
+                                    "type": [
+                                        "object",
+                                        "string"
+                                    ]
+                                }
+                            }
                         }
                     }
                 },
--- a/values.yaml
+++ b/values.yaml
@@ -459,6 +459,52 @@ server:
     # to the service.
     annotations: {}
 
+    # Enables a headless service that contains the active pod only
+    active:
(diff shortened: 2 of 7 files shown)


velkovb:
This would solve https://github.com/hashicorp/vault-helm/issues/588

eye0fra:
Related #674

jamesmbourne:
This would be really helpful - we are facing this same issue meaning multiple load balancers are provisioned without any mechanism to control which services they are linked with.

peikk0:
Any update on this? I would also find this change helpful to be able to set annotations for a particular service.

values.schema.json
```diff
@@ -685,6 +685,28 @@
                         },
                         "targetPort": {
                             "type": "integer"
+                        },
+                        "active": {
+                            "type": "object",
+                            "properties": {
+                                "annotations": {
```
swenson, line 692: These appear to be missing entries for `type`, `nodePort`, clusterIP`.

values.yaml
```diff
@@ -459,6 +459,52 @@ server:
     # to the service.
     annotations: {}
 
+    # Enables a headless service that contains the active pod only
```
swenson, line 462: These sections don't enable this behavior, so I would remove this comment.

values.yaml
```diff
@@ -459,6 +459,52 @@ server:
...
+    active:
+      # clusterIP controls whether a Cluster IP address is attached to the
```
swenson, line 464: Maybe add a comment about these values defaulting to the `values.server.service` values if missing, except `annotations` (which will be appended)?

tekicat:
Revived this PR in https://github.com/hashicorp/vault-helm/pull/896

tvoran:
Superseded by #896

PR #547, opened 2021-06-04
Adding server.enterpriseLicense

tvoran:
Sets up a vault-enterprise license for autoloading on
startup. Mounts an existing secret to /vault/license and sets
VAULT_LICENSE_PATH appropriately.

The UX will be similar to [consul-helm](https://www.consul.io/docs/k8s/installation/deployment-configurations/consul-enterprise).

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -107,6 +107,7 @@ extra volumes the user may have specified (such as a secret with TLS).
         - name: vault-license
           secret:
             secretName: {{ .Values.server.enterpriseLicense.secretName }}
+            defaultMode: 0440
   {{- end }}
 {{- end -}}
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -103,6 +103,11 @@ extra volumes the user may have specified (such as a secret with TLS).
   {{- if .Values.server.volumes }}
     {{- toYaml .Values.server.volumes | nindent 8}}
   {{- end }}
+  {{- if (and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey) }}
+        - name: vault-license
+          secret:
+            secretName: {{ .Values.server.enterpriseLicense.secretName }}
+  {{- end }}
 {{- end -}}
 
 {{/*
@@ -165,6 +170,11 @@ based on the mode configured.
   {{- end }}
   {{- if .Values.server.volumeMounts }}
     {{- toYaml .Values.server.volumeMounts | nindent 12}}
+  {{- end }}
+  {{- if (and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey) }}
+            - name: vault-license
+              mountPath: /vault/license
+              readOnly: true
   {{- end }}
 {{- end -}}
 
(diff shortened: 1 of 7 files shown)


values.yaml
```diff
@@ -198,6 +198,17 @@ server:
   # If not set to true, Vault server will not be installed. See vault.mode in _helpers.tpl for implementation details
   enabled: true
 
+  # [Enterprise Only] This value refers to a Kubernetes secret that you have
```
ncabatoff, line 201: Do we want to be a little more explicit here?  We could note that for enterprise Vault 1.8 and later, enterpriseLicense is required unless using the binaries with baked-in licenses (pro/prem) or if the cluster was created with an older Vault version and the license was stored.

I'm not sure that we have to, it's a genuine question.

tvoran: Good question; I think we can be more explicit about the cases where a license is required in the vault-helm [docs pages](https://www.vaultproject.io/docs/platform/k8s/helm), since setting `enterpriseLicense` here is just one way of setting the license on the vault pods.

Specifically I'm thinking of adding a page similar to https://www.consul.io/docs/k8s/installation/deployment-configurations/consul-enterprise.

values.yaml
```diff
@@ -198,6 +198,18 @@ server:
...
+  # created that contains your enterprise license. If you are not using an
+  # enterprise image or if you plan to introduce the license key via another
+  # route, then set these fields to empty (""). Requires Vault Enterprise 1.8 or
+  # later.
+  enterpriseLicense:
+    # The name of the Kubernetes secret that holds the enterprise license. The
+    # secret must be in the same namespace that Vault is installed into.
+    secretName: ""
+    # The key within the Kubernetes secret that holds the enterprise license.
+    secretKey: ""
```
jasonodonnell, line 211: Nit: should we have a reasonable default key name (like `license`)?

templates/_helpers.tpl
```diff
@@ -103,6 +103,11 @@ extra volumes the user may have specified (such as a secret with TLS).
   {{- if .Values.server.volumes }}
     {{- toYaml .Values.server.volumes | nindent 8}}
   {{- end }}
+  {{- if (and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey) }}
+        - name: vault-license
+          secret:
+            secretName: {{ .Values.server.enterpriseLicense.secretName }}
```
jasonodonnell, line 109: Nit: should we set the perms on this file?

tvoran: Sure, something like `defaultMode: 0440`?

values.yaml
tvoran: Yeah, I think that makes sense. It'll make examples in the docs a little simpler too.

templates/_helpers.tpl
jasonodonnell: Works for me!

values.yaml
tvoran: Done in 95beeccfd6162ccf80a6bf97e64338fc5238f2f1

templates/_helpers.tpl
tvoran: Done in 95beeccfd6162ccf80a6bf97e64338fc5238f2f1

PR #549, opened 2021-06-09
Add openshift overrides

tvoran:
Adds default overrides for OpenShift (values.openshift.yaml) and uses
them in the chart-verifier tests. Also updates chart-verifier to 1.0.0.

This is essentially an alternative to #538, and will require setting the overrides in values.yaml somewhere along the pipeline before the chart lands in the OpenShift helm chart repo.


jasonodonnell (approved):
LGTM!

PR #557, opened 2021-06-24
Use imagePullSecrets as list containing only names

rahulsadanandan:
While using vault chart in platform chart implementations, it's causing an error with the way imagePullSecrets is defined.
Most of the common chart implementations have imagePullSecrets to accepts only names as a list.
this pr won’t any breaking change. We have added a flag that will accept imagePullSecret as list containing only names
Default way
```
   imagePullSecrets:
     - name: <secret_name>
```
will also accept
```
  imagePullSecrets:
    - <secret_name>
```


chukka:
@tvoran Can you please review this

PR #560, opened 2021-06-30
injector: add templateConfig.exitOnRetryFailure annotation

calvn:
This PR adds `templateConfig.exitOnRetryFailure` annotation to the Helm chart in order to make the new parameter introduced in https://github.com/hashicorp/vault-k8s/pull/267 globally configurable.

We should hold off on merging this until the next vault-k8s release that contains the new param on agent-inject has shipped though.

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -107,6 +107,8 @@ spec:
               value: "{{ .Values.injector.agentDefaults.memLimit }}"
             - name: AGENT_INJECT_DEFAULT_TEMPLATE
               value: "{{ .Values.injector.agentDefaults.template }}"
+            - name: AGENT_INJECT_TEMPLATE_CONFIG_EXIT_ON_RETRY_FAILURE
+              value: "{{ .Values.injector.agentDefaults.templateConfig.exitOnRetryFailure }}"
             {{- include "vault.extraEnvironmentVars" .Values.injector | nindent 12 }}
           args:
             - agent-inject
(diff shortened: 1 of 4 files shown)


templates/injector-deployment.yaml
```diff
@@ -107,6 +107,8 @@ spec:
               value: "{{ .Values.injector.agentDefaults.memLimit }}"
             - name: AGENT_INJECT_DEFAULT_TEMPLATE
               value: "{{ .Values.injector.agentDefaults.template }}"
+            - name: AGENT_INJECT_TEMPLATE_CONFIG_EXIT_ON_RETRY_FAILURE
+              value: "{{ .Values.injector.agentDefaults.templateConfig.exitOnRetryFailure }}"
```
calvn, line 111: If I add  `{{ ... | default true }}` the `injector/deployment: can set agent template_config.exit_on_retry_failure"` bats test fails, and I'm not sure why.

tvoran: Defaults should be only set in values.yaml since it's easier to keep track, so this is fine.

calvn: I saw that `.Values.injector.revokeOnShutdown` has a default value so I wanted to give it a try. I'm still a bit curious on why `--set` on the test doesn't work if I add a default over there though.

tvoran: I suspect that's because revokeOnShutdown is a boolean, so if it's `false` that the same as it being unset, which is when the `| default true }` would be used.

tvoran (approved):
Can you also add this to the json schema? https://github.com/hashicorp/vault-helm/blob/master/test/README.md#generating-the-values-json-schema

benashz:
I think we are only missing the update to `values.schema.json`, otherwise 👍

benashz (approved):
I think we are only missing the update to `values.schema.json`, otherwise 👍

PR #565, opened 2021-07-06
feature: Support configuring various properties as YAML directly.

benashz:
Supported properties include: pod toleration, pod affinity, and node selectors.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -299,12 +299,12 @@ Sets the injector node selector for pod placement
 {{- define "injector.nodeselector" -}}
   {{- if .Values.injector.nodeSelector }}
       nodeSelector:
-        {{- $tp := typeOf .Values.injector.nodeSelector }}
-        {{- if eq $tp "string" }}
-          {{ tpl .Values.injector.nodeSelector . | nindent 8 | trim }}
-        {{- else }}
-          {{- toYaml .Values.injector.nodeSelector | nindent 8 }}
-        {{- end }}
+      {{- $tp := typeOf .Values.injector.nodeSelector }}
+      {{- if eq $tp "string" }}
+        {{ tpl .Values.injector.nodeSelector . | nindent 8 | trim }}
+      {{- else }}
+        {{- toYaml .Values.injector.nodeSelector | nindent 8 }}
+      {{- end }}
   {{- end }}
 {{- end -}}
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -224,9 +224,9 @@ Set's the affinity for pod placement when running in standalone and HA modes.
 {{- define "vault.affinity" -}}
   {{- if and (ne .mode "dev") .Values.server.affinity }}
       affinity:
-        {{- $tp := typeOf .Values.server.affinity }}
+        {{ $tp := typeOf .Values.server.affinity }}
         {{- if eq $tp "string" }}
-          {{- tpl .Values.server.affinity . | nindent 8 }}
+          {{- tpl .Values.server.affinity . | nindent 8 | trim }}
         {{- else }}
           {{- toYaml .Values.server.affinity | nindent 8 }}
         {{- end }}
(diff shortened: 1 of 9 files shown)


templates/_helpers.tpl
```diff
@@ -274,7 +299,12 @@ Sets the injector node selector for pod placement
 {{- define "injector.nodeselector" -}}
   {{- if .Values.injector.nodeSelector }}
       nodeSelector:
-        {{ tpl .Values.injector.nodeSelector . | indent 8 | trim }}
+        {{- $tp := typeOf .Values.injector.nodeSelector }}
+        {{- if eq $tp "string" }}
+          {{- tpl .Values.injector.nodeSelector . | indent 8 }}
```
tvoran, line 304: I'm curious why `trim` was removed from all the string cases.

tvoran (commented):
Looks pretty good! Just a few questions.

values.schema.json
```diff
@@ -80,7 +80,11 @@
                             ]
                         },
                         "tolerations": {
-                            "type": ["null", "string"]
+                            "type": [
+                                "null",
```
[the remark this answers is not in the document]
benashz: I think that it is still needed to avoid breaking preexisting configuration e.g in the case where someone has set it explicitly to `null`.  I believe validation is done for the following commands `install`, `upgrade`, `lint`, and `template`.

templates/_helpers.tpl
```diff
@@ -264,7 +284,12 @@ Set's the node selector for pod placement when running in standalone and HA mode
 {{- define "vault.nodeselector" -}}
   {{- if and (ne .mode "dev") .Values.server.nodeSelector }}
       nodeSelector:
-        {{ tpl .Values.server.nodeSelector . | indent 8 | trim }}
+      {{- $tp := typeOf .Values.injector.tolerations }}
```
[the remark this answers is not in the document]
benashz: Good catch :)

templates/_helpers.tpl
benashz: I was getting YAML parse errors in the tests with `trim`, but I see now that we were getting extra white space trimmed by the Go template itself. 

The following change allows us to use `trim` without breaking the YAML.
```
@@ -551,7 +551,7 @@ Sets the injector toleration for pod placement
       tolerations:
       {{- $tp := typeOf .Values.csi.pod.tolerations }}
       {{- if eq $tp "string" }}
-        {{- tpl .Values.csi.pod.tolerations . | nindent 8 }}
+        {{ tpl .Values.csi.pod.tolerations . | nindent 8 | trim }}
       {{- else }}
         {{- toYaml .Values.csi.pod.tolerations | nindent 8 }}
       {{- end }}

```

templates/_helpers.tpl
```diff
@@ -274,7 +299,12 @@ Sets the injector node selector for pod placement
 {{- define "injector.nodeselector" -}}
   {{- if .Values.injector.nodeSelector }}
       nodeSelector:
-        {{ tpl .Values.injector.nodeSelector . | indent 8 | trim }}
+        {{- $tp := typeOf .Values.injector.nodeSelector }}
+        {{- if eq $tp "string" }}
+          {{ tpl .Values.injector.nodeSelector . | nindent 8 | trim }}
+        {{- else }}
+          {{- toYaml .Values.injector.nodeSelector | nindent 8 }}
+        {{- end }}
```
tvoran, line 307: The string case has some extra indenting here when rendered, so we may want to shift the whole block back like this:

```suggestion
      {{- $tp := typeOf .Values.injector.nodeSelector }}
      {{- if eq $tp "string" }}
        {{ tpl .Values.injector.nodeSelector . | nindent 8 | trim }}
      {{- else }}
        {{- toYaml .Values.injector.nodeSelector | nindent 8 }}
      {{- end }}
```

tvoran (approved):
One more small suggestion.

templates/_helpers.tpl
benashz: Fixed in latest commit.

PR #566, opened 2021-07-08
Use imagePullSecrets as list containing only names

rahulsadanandan:
While using vault chart in platform chart implementations, it's causing an error with the way imagePullSecrets is defined.
Most of the common chart implementations have imagePullSecrets to accepts only names as a list.
this pr won’t have any breaking change. We have added a flag that will accept imagePullSecret as list containing only names
Default way
```
   imagePullSecrets:
     - name: <secret_name>
```
will also accept
```
  imagePullSecrets:
    - <secret_name>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Improvements:
 * Add support for running the acceptance tests against a local `kind` cluster [GH-567](https://github.com/hashicorp/vault-helm/pull/567)
 * Add `server.ingress.activeService` to configure if the ingress should use the active service [GH-570](https://github.com/hashicorp/vault-helm/pull/570)
 * Add `server.route.activeService` to configure if the route should use the active service [GH-570](https://github.com/hashicorp/vault-helm/pull/570)
+* Use imagePullSecrets as list containing only names [GH-566](https://github.com/hashicorp/vault-helm/pull/566)
 
 
 ## 0.13.0 (June 17th, 2021)
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -623,3 +623,21 @@ Inject extra environment populated by secrets, if populated
 {{ "https" }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Resolve imagePullSecrets value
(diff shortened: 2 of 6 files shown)


rahulsadanandan:
@tvoran Can you please review this
cc @chukka @rimusz

rimusz:
We are trying to integrate Vault chart in https://github.com/jfrog/charts/tree/master/stable/pipelines which is also is a part of the https://github.com/jfrog/charts/tree/master/stable/jfrog-platform where that change we would like to use.

tvoran:
Hi folks, can you elaborate a bit more on why the current format is problematic?

chukka:
Hi @tvoran , as @rimusz  pointed out above, We want to use Vault chart as dependency chart in both of our  JFrog helm charts - [pipelines](https://github.com/jfrog/charts/tree/master/stable/pipelines) and[ jfrog-platform ](https://github.com/jfrog/charts/tree/master/stable/jfrog-platform)

We use  imagePullSecrets  as list  https://github.com/jfrog/charts/blob/master/stable/jfrog-platform/values.yaml#L7 

```
imagePullSecrets:
    - <secret_name>
```

and Since Vault chart will be  dependency chart , we would like to pass same structure there too. However, when we checked your implementation  it accepts as below. 
```
   imagePullSecrets:
     - name: <secret_name>
```
Hence, we raised this PR which will allow us to use vault chart without `name` without any breaking change. Hope this helps

**Note :** we use bitnami dependency [charts](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml#L8) which also follow same convention under global values

chukka:
@tvoran Did you get a chance to review my above comment ?

chukka:
@tvoran @benashz Reminder ! Please review

benashz:
Hi @chukka @rahulsadanandan, I  just wanted to let you know that we are currently looking at this PR and will get back to you soon.

Thanks.

benashz:
@rahulsadanandan we have merged your feature request with #576. It will be available in the next release.

Thank you for your interest and contribution!

PR #570, opened 2021-07-15
Add ingress/route configurable to specify active/general service

jasonodonnell:
Currently when using ingress/routes, if HA mode is enabled, these services will always point to the active service. There are use cases, such as performance standbys, where the ingress should instead point to the general Vault service instead. 

This adds two new boolean configs `server.ingress.activeService` and `server.route.activeService` to make this configurable.

### DIFF
--- a/templates/server-route.yaml
+++ b/templates/server-route.yaml
@@ -2,7 +2,7 @@
 {{- if ne .mode "external" }}
 {{- if .Values.server.route.enabled -}}
 {{- $serviceName := include "vault.fullname" . -}}
-{{- if eq .mode "ha" }}
+{{- if and (eq .mode "ha" ) (eq (.Values.server.route.activeService | toString) "true") }}
 {{- $serviceName = printf "%s-%s" $serviceName "active" -}}
 {{- end }}
 kind: Route
--- a/test/unit/server-ingress.bats
+++ b/test/unit/server-ingress.bats
@@ -131,7 +131,7 @@ load _helpers
   [ "${actual}" = "nginx" ]
 }
 
-@test "server/ingress: uses active service when ha - yaml" {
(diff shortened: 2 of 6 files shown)


test/unit/server-ingress.bats
```diff
@@ -157,4 +172,19 @@ load _helpers
       . | tee /dev/stderr |
       yq -r '.spec.rules[0].http.paths[0].backend.serviceName' | tee /dev/stderr)
   [ "${actual}" = "RELEASE-NAME-vault" ]
-}
\ No newline at end of file
+}
+
+@test "server/ingress: uses regular service when not ha and activeService is true- yaml" {
```
benashz, line 177:
```suggestion
@test "server/ingress: uses regular service when not ha and activeService is true - yaml" {
```

templates/server-route.yaml
```diff
@@ -1,33 +1,33 @@
...
-    name: {{ $serviceName }}
-    weight: 100
-  port:
-    targetPort: 8200
-  tls:
-    termination: passthrough
-{{- end }}
-{{- end }}
-{{- end }}
+{{- if .Values.global.openshift }}
```
benashz, line 1: Seems to be quite a bit of whitespace change, dos2unix :)

benashz (approved):
LGTM!

templates/server-route.yaml
jasonodonnell: It does look strange, doesn't it?

benashz: Looks like we had DOS line feeds. Seems to have been the only file affected.

```
$ git ls-files -z | xargs -0 file|grep CRLF
templates/server-route.yaml:                                ASCII text, with CRLF line terminators
```

PR #572, opened 2021-07-19
Add imagePullSecrets on server test

shiinro:
When deploying Vault on a private cluster and a private secured repository, server-test pull image fails cause of credentials

I propose  to add imagePullSecrets on the pod

### DIFF
--- a/templates/tests/server-test.yaml
+++ b/templates/tests/server-test.yaml
@@ -7,8 +7,7 @@ metadata:
   annotations:
     "helm.sh/hook": test
 spec:
-  imagePullSecrets:
-    {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
+  {{- include "imagePullSecrets" . | nindent 6 }}
   containers:
     - name: {{ .Release.Name }}-server-test
       image: {{ .Values.server.image.repository }}:{{ .Values.server.image.tag | default "latest" }}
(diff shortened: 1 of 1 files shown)


templates/tests/server-test.yaml
```diff
@@ -7,6 +7,8 @@ metadata:
   annotations:
     "helm.sh/hook": test
 spec:
+  imagePullSecrets:
+    {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
```
benashz, line 11: After rebasing master:

```suggestion
  {{- include "imagePullSecrets" . | nindent 6 }}
```

templates/tests/server-test.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   annotations:
     "helm.sh/hook": test
 spec:
+  {{- include "imagePullSecrets" . | nindent 6 }}
```
tvoran, line 10: I think the indentation just needs to be a little less here:

```suggestion
  {{- include "imagePullSecrets" . | nindent 2 }}
```

tvoran (commented):
You'll need to merge in the latest changes from master as well to get the `imagePullSecrets` helper template.

tvoran (approved):
The tests pass when I merge in master locally, so I think this is good to go.

PR #576, opened 2021-07-23
feature: imagePullSecrets from string array.

benashz:
* allow configuring `imagePullSecrets` from an array of strings in
  addition to the already supported array of maps

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -623,3 +623,20 @@ Inject extra environment populated by secrets, if populated
 {{ "https" }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+imagePullSecrets generates pull secrets from either string or map values.
+A map value must be indexable by the key 'name'.
+*/}}
+{{- define "imagePullSecrets" -}}
+{{- with .Values.global.imagePullSecrets -}}
+imagePullSecrets:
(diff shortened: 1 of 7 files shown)


tomhjp (approved):
LGTM

tvoran (approved):
Nice solution!

tvoran:
We may also want to update the comment in values.yaml to mention that an array of strings is now supported as well.

templates/_helpers.tpl
```diff
@@ -623,3 +623,20 @@ Inject extra environment populated by secrets, if populated
...
+{{/*
+imagePullSecrets generates pull secrets from either string or map values.
+A map value must be indexable by the key 'name'.
+*/}}
+{{- define "imagePullSecrets" -}}
+{{- with .Values.global.imagePullSecrets -}}
+imagePullSecrets:
+{{- range . -}}
+{{- if typeIs "string" . }}
+  - name: {{ . }}
```
chukka, line 636: @benashz shouldn't be just `- {{ . }}` ?

cc @tvoran

benashz: @chukka for `imagePullSecrets` we need a mapping of `name: <value>`, so in the case of string values we need to create the mapping  dynamically from the string value.

PR #579, opened 2021-07-29
Update chart verifier

tvoran:
Updates the version of [chart-verifier](https://github.com/redhat-certification/chart-verifier) in use in CI to 1.2.1. 

The tests it runs appear to all be the same, just with slightly different output/names.

The tests can be run locally with bats by using the chart-verifier docker image:

```console
USE_DOCKER=true bats test/chart/verifier.bats
```

Or by installing the chart-verifier binary and running the tests:

```console
go get github.com/redhat-certification/chart-verifier@1.2.1

bats test/chart/verifier.bats
```


benashz (approved):
LGTM!

PR #584, opened 2021-08-10
Add injector.webhookAnnotations chart option

tvoran:
Allows setting annotations on the mutating webhook config.

For example, if cert-manager is managing the certificates for vault-k8s, that would require [setting the CA cert in the chart values (`injector.certs.caBundle`)](https://www.vaultproject.io/docs/platform/k8s/helm/examples/injector-tls-cert-manager#configuration). But, if the [`cert-manager.io/inject-ca-from: <namespace>/<secret>` annotation](https://cert-manager.io/docs/concepts/ca-injector/) is set on the mutating webhook config, cert-manager will keep the mutating webhook config updated with the CA from the Secret.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -354,6 +354,21 @@ Sets extra injector service annotations
 {{- end -}}
 
 {{/*
+Sets extra injector webhook annotations
+*/}}
+{{- define "injector.webhookAnnotations" -}}
+  {{- if .Values.injector.webhookAnnotations }}
+  annotations:
+    {{- $tp := typeOf .Values.injector.webhookAnnotations }}
+    {{- if eq $tp "string" }}
+      {{- tpl .Values.injector.webhookAnnotations . | nindent 4 }}
+    {{- else }}
+      {{- toYaml .Values.injector.webhookAnnotations | nindent 4 }}
+    {{- end }}
+  {{- end }}
+{{- end -}}
+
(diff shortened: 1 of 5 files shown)


templates/_helpers.tpl
```diff
@@ -353,6 +353,21 @@ Sets extra injector service annotations
...
 {{- end -}}
 
+{{/*
+Sets extra injector webhook annotations
+*/}}
+{{- define "injector.webhookAnnotations" -}}
+  {{- if .Values.injector.webhookAnnotations }}
+  annotations:
+    {{- $tp := typeOf .Values.injector.webhookAnnotations }}
+    {{- if eq $tp "string" }}
```
tomhjp, line 363: Is it worth supporting string for new options?

tomhjp (approved):
LGTM, should we also update the schema? Happy to leave that to a separate PR if we'd rather batch those changes up/roll them at release time.

templates/_helpers.tpl
tvoran: Good question, though I feel like even with new options we need to keep supporting both types for consistency with the other annotation options in the chart.

tvoran:
@tomhjp Thanks for the reminder! Schema updated in https://github.com/hashicorp/vault-helm/pull/584/commits/c2a384b5adf785771679201f63cf4e123367fedb.

PR #589, opened 2021-08-13
Vault Injector Web Hook should be registered before deployment of the dependent pods

gau31rav:
Issue:  If vault webhook is not registered before deployment of a dependent pod, then vault(init/side-car) containers will not be added to the pod.

As helm doesn't guarantee the order of creation of webhook, there is a race condition where the dependent pod gets deployed before the webhook and hence sidecar injection failed.

In this case, the vault has to be installed as a separate chart. To avoid that, adding annotations to the webhook making sure that it gets created first.

Solution:  Added helm hook annotations to register the webhook during pre-installation/pre-upgrade steps.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -99,6 +99,7 @@ extra volumes the user may have specified (such as a secret with TLS).
             secretName: {{ .name }}
           {{- end }}
             defaultMode: {{ .defaultMode | default 420 }}
+            optional: {{ .optional | default "false" }}
   {{- end }}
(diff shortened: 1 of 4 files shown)


mballoni:
This is interesting.
 I suspect we may have been facing this race condition described here. Intermittently some pods are initializing without the Vault init container and sidecar and we still haven't figure out why. I suspect now it may be the case.
 @gau31rav on a side note, would you have any tips on how I could confirm if this is the case on my described scenario?
 
 Anyway, I hope this gets accepted soon, great job.

Issue #587, opened 2021-08-13
Add support for Ingress networking.k8s.io/v1


velkovb:
**Is your feature request related to a problem? Please describe.**
networking.k8s.io/v1 and networking.k8s.io/v1beta1 have different syntax and it is good to support the new version as the old one will get deprecated soon.

jeroenlandheer:
We're having this as a blocking issue atm, we cannot use the helm chart on our K8s 1.22 cluster. (Ingress = v1beta)

PR #590, opened 2021-08-13
ingress stable networking api

vrumjantsev:
Moved ingress to stable api 
https://github.com/hashicorp/vault-helm/issues/587

### DIFF
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -9,8 +9,10 @@
 {{- end }}
 {{- $servicePort := .Values.server.service.port -}}
 {{- $kubeVersion := .Capabilities.KubeVersion.Version }}
-{{ if semverCompare ">= 1.19" $kubeVersion }}
+{{ if semverCompare ">= 1.19.0-0" $kubeVersion }}
 apiVersion: networking.k8s.io/v1
+{{ else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
+apiVersion: networking.k8s.io/v1beta1
 {{ else }}
 apiVersion: extensions/v1beta1
 {{ end }}
@@ -49,7 +51,7 @@ spec:
         {{- range (.paths | default (list "/")) }}
           - path: {{ . }}
             backend:
-              {{ if semverCompare ">= 1.19" $kubeVersion }}
+              {{ if semverCompare ">= 1.19.0-0" $kubeVersion }}
               service:
                 name: {{ $serviceName }}
                 port:
@@ -58,7 +60,6 @@ spec:
               serviceName: {{ $serviceName }}
               servicePort: {{ $servicePort }}
               {{ end }}
-
         {{- end }}
   {{- end }}
 {{- end }}
--- a/templates/server-ingress.yaml
+++ b/templates/server-ingress.yaml
@@ -8,11 +8,7 @@
 {{- $serviceName = printf "%s-%s" $serviceName "active" -}}
 {{- end }}
 {{- $servicePort := .Values.server.service.port -}}
-{{ if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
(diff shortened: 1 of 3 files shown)


templates/server-ingress.yaml
```diff
@@ -8,11 +8,7 @@
 {{- $serviceName = printf "%s-%s" $serviceName "active" -}}
 {{- end }}
 {{- $servicePort := .Values.server.service.port -}}
-{{ if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
-apiVersion: networking.k8s.io/v1beta1
-{{ else }}
-apiVersion: extensions/v1beta1
-{{ end }}
+apiVersion: networking.k8s.io/v1
```
tvoran, line 11: Since networking.k8s.io/v1 Ingress has only been available since k8s 1.19, I think we'll need to maintain some of the older apiVersion's here, back to k8s 1.14 if possible.

vrumjantsev: done

guyguy333:
Thanks for this PR, it is required for K8S 1.22. I hope it will be merged soon :)

templates/server-ingress.yaml
```diff
@@ -8,8 +8,9 @@
 {{- $serviceName = printf "%s-%s" $serviceName "active" -}}
 {{- end }}
 {{- $servicePort := .Values.server.service.port -}}
-{{ if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
-apiVersion: networking.k8s.io/v1beta1
+{{- $kubeVersion := .Capabilities.KubeVersion.Version }}
+{{ if semverCompare ">= 1.19" $kubeVersion }}
+apiVersion: networking.k8s.io/v1
 {{ else }}
```
tvoran, line 14: I think the semver version needs to be 1.19.0-0, to avoid problems with pre-release style version strings, as in GKE: https://github.com/helm/helm/issues/3810

Also, just so we don't change the API version unexpectedly on folks, can we do something like this?

```suggestion
{{- $kubeVersion := .Capabilities.KubeVersion.Version }}
{{ if semverCompare ">= 1.19.0-0" $kubeVersion }}
apiVersion: networking.k8s.io/v1
{{ else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
apiVersion: networking.k8s.io/v1beta1
{{ else }}
```

templates/server-ingress.yaml
```diff
@@ -48,8 +49,16 @@ spec:
         {{- range (.paths | default (list "/")) }}
           - path: {{ . }}
             backend:
+              {{ if semverCompare ">= 1.19" $kubeVersion }}
```
tvoran, line 52:
```suggestion
              {{ if semverCompare ">= 1.19.0-0" $kubeVersion }}
```

templates/server-ingress.yaml
```diff
@@ -48,8 +49,16 @@ spec:
...
+              service:
+                name: {{ $serviceName }}
+                port:
+                  number: {{ $servicePort }}
+              {{ else }}
               serviceName: {{ $serviceName }}
               servicePort: {{ $servicePort }}
+              {{ end }}
+
```
tvoran, line 61:
```suggestion
```

tvoran (commented):
Would you also mind adding a test in `test/unit/server-ingress.bats` for the old Ingress style? 

Something like:
```
helm template \
      --show-only templates/server-ingress.yaml \
      --set 'server.ingress.enabled=true' \                      
      --kube-version 1.18.3 \
      .
```
And then check for the apiVersion, serviceName, and servicePort.

vrumjantsev:
@tvoran sorry for the delay, updated

tvoran (approved):
Thanks!

tvoran:
@np-13 Probably next week.

Issue #542, opened 2021-06-01
Ability to set "ingressClassName" field for ingress resources - "kubernetes.io/ingress.class" annotation is going to be depricated.


jppitout:
**Is your feature request related to a problem? Please describe.**
We cannot set the `ingressClassName` through the values file only the `kubernetes.io/ingress.class` annotation which is deprecated. We get  the below warning in nginx ingress controller logs:
```
 Using the DEPRECATED annotation 'kubernetes.io/ingress.class'. The 'ingressClassName' field will be ignored.
```

**Describe the solution you'd like**
Allow ingressClassName to be set through values file.

**Describe alternatives you've considered**
N/A have no choice but to use the annotation.

**Additional context**
Many other values file allow this to be set with a `ingressClassName` key.

PR #591, opened 2021-08-14
Add server.ingress.ingressClassName

jtcressy:
Support "ingressClassName" field for ingress resources to move away from deprecated annotation
#542 

(See: [https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation))

Results of unit tests:
```shell
$ docker run -it --rm -v "${PWD}:/test" vault-helm-test bats /test/test/unit -f "ingressClassName"
1..1
ok 1 server/ingress: ingressClassName added to object spec - string
```

### DIFF
--- a/test/unit/server-ingress.bats
+++ b/test/unit/server-ingress.bats
@@ -131,6 +131,18 @@ load _helpers
   [ "${actual}" = "nginx" ]
 }
 
+@test "server/ingress: ingressClassName added to object spec - string" {
+  cd `chart_dir`
+
+  local actual=$(helm template \
+      --show-only templates/server-ingress.yaml \
+      --set 'server.ingress.enabled=true' \
+      --set server.ingress.ingressClassName=nginx \
+      . | tee /dev/stderr |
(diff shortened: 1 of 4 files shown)


jtcressy:
This might also coincide with or require #590 if this field is only available in the stable API. Requesting comment on this from someone who might be more knowledgeable.

velkovb:
ingressClassName is available in "networking.k8s.io/v1beta1" as well.

test/unit/server-ingress.bats
```diff
@@ -131,6 +131,18 @@ load _helpers
   [ "${actual}" = "nginx" ]
 }
 
+@test "server/ingress: ingressClassName added to object spec - string" {
```
tvoran, line 134: It would also be good to add the opposite test, that ingressClassName is not added to the Ingress spec when server.ingress.ingressClassName is not specified.

PR #593, opened 2021-08-17
Update values.yaml

infa-rsundik:


tvoran:
Thanks for the patch here, though I think we'll want to leave that default as blank (and the CLA would need to be signed too). Closing for now.

PR #594, opened 2021-08-17
injector: ability to set deployment update strategy

jasonhancock:
I have a cluster with 2 worker nodes. When deploying the injector with 2 replicas, any updates that require a change to the deployment will cause the new pods to get stuck in a pending state due to the affinity settings. If the `.spec.strategy.rollingUpdate.maxUnavailable` is able to be set to `1` instead of `25%`, this problem can be avoided in my small clusters.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -202,6 +202,13 @@ injector:
     # Extra annotations to attach to the injector service
     annotations: {}
 
+  # strategy for updating the deployment. This can be a multi-line string or a YAML map.
+  strategy: |
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+
 server:
   # If not set to true, Vault server will not be installed. See vault.mode in _helpers.tpl for implementation details
   enabled: true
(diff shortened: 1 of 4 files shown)


values.yaml
```diff
@@ -202,6 +202,13 @@ injector:
     # Extra annotations to attach to the injector service
     annotations: {}
 
+  # strategy for updating the deployment. This can be a multi-line string or a YAML map.
+  strategy: |
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+
```
tvoran, line 211: What do you think about defaulting to an empty value, instead of an explicit default? 

From looking through the values file, it looks like we've tended to default to YAML for options that support both YAML and multi-line string formats. Something like this?

```suggestion
  # strategy for updating the deployment. This can be a multi-line string or a YAML map.
  strategy: {}
  # strategy: |
  #   rollingUpdate:
  #     maxSurge: 25%
  #     maxUnavailable: 25%
  #   type: RollingUpdate
```

tvoran (commented):
Thanks for this! I think this will address #615 as well. Just a question about the default.

Would you mind updating values.schema.json as well?

tvoran:
Continued in #661

PR #599, opened 2021-08-20
Update csi provider to support OpenShift

paulczar:
OpenShift requires the Pod to be privileged to run as a user to be able
to write data to the host volume.  It also requires an SCC to applied
by the OpenShift administrator as noted in NOTES.

Signed-off-by: Paul Czarkowski <username.taken@gmail.com>

### DIFF
--- a/templates/NOTES.txt
+++ b/templates/NOTES.txt
@@ -12,3 +12,18 @@ Your release is named {{ .Release.Name }}. To learn more about the release, try:
   $ helm status {{ .Release.Name }}
   $ helm get manifest {{ .Release.Name }}
 
+{{ if and .Values.global.openshift .Values.csi.enabled }}
+Check that the {{ template "vault.fullname" . }}-csi-provider DaemonSet has
+the correct number of Pods running:
+
+$ kubectl -n {{ .Release.Namespace }} get daemonset {{ template "vault.fullname" . }}-csi-provider
+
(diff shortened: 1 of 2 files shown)


paulczar:
@paraenggu sorry for the very late response.  As far as I could determine these were the necessary settings,  It's possible more restrictive settings could be set, but I don't have the time to really dig into it and was more interested in just getting this working.  I grant that "not working" is more secure than "not super restrictive" but its also a lot less useful.

vmaillot:
Ahoy, any news on this PR? That would be nice to have this feature in the main line! Thanks for the feedback o/

swenson:
I just merged #767, which will allow you to specify custom `securityContext` for the CSI provider. I think this is sufficient to address this issue in a slightly more flexible way without hard-coding `privileged: true` (which I am hesitant about)?

vmaillot:
> I just merged #767, which will allow you to specify custom `securityContext` for the CSI provider. I think this is sufficient to address this issue in a slightly more flexible way without hard-coding `privileged: true` (which I am hesitant about)?

@swenson thanks a lot for this fix, it is way more flexible for sure. That should work perfectly fine for us! It is much appreciated.

vmaillot:
@swenson do you know if this feature will be in the next release? Thanks o/

swenson:
@vmaillot I'll probably make a release this week or next.

vmaillot:
> @vmaillot I'll probably make a release this week or next.

Awesome, thanks for that!

PR #601, opened 2021-08-25
Added variables to server-route.yaml to make it more flexible about ssl

leofin:
Using OpenShift.
I have some work to figure out that I can't use the default route tls configuration as `passthrough` because I wasn't using a valid certificate. Then my motivation to help other people that want to use this helm chart and have more flexibility to configure the tls using just the values.yaml.

### DIFF
--- a/templates/server-route.yaml
+++ b/templates/server-route.yaml
@@ -27,7 +27,8 @@ spec:
   port:
     targetPort: 8200
   tls:
-    termination: passthrough
+    termination: {{ .Values.server.route.tls.termination }}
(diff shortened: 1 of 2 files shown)


leofin:
@hashicorp-cla can anyone review this, please?

mbaldessari:
I supposed this has been mostly addressed via https://github.com/hashicorp/vault-helm/commit/92da5125777b87b7475f546e0a7127b9517fe735 ?

leofin:
Thank you for noticing that! Closing.

PR #603, opened 2021-08-26
feat(csi/providersDir): add configurable providersDir

eyenx:
* Adds a configurable providersDir path for the Daemonset

Background: with https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/409/ the providersDir path has been made configurable for secrets-store-csi-driver, as not every single kubernetes distribution out there uses the same path. 

This is why we should also allow vault-helm to set providersDir accordingly. Default will remain `/etc/kubernetes/secrets-store-csi-providers`

Signed-off-by: Toni Tauro <toni.tauro@adfinis.com>

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vault
-version: 0.15.0
+version: 0.15.1
 appVersion: 1.8.1
 kubeVersion: ">= 1.14.0-0"
 description: Official HashiCorp Vault Chart
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vault
-version: 0.15.1
+version: 0.15.0
 appVersion: 1.8.1
 kubeVersion: ">= 1.14.0-0"
 description: Official HashiCorp Vault Chart
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -70,7 +70,7 @@ spec:
       volumes:
         - name: providervol
           hostPath:
-            path: "/etc/kubernetes/secrets-store-csi-providers"
+            path: {{ .Values.csi.daemonSet.providersDir }}
         - name: mountpoint-dir
           hostPath:
             path: /var/lib/kubelet/pods
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -73,7 +73,7 @@ spec:
             path: {{ .Values.csi.daemonSet.providersDir }}
         - name: mountpoint-dir
           hostPath:
-            path: /var/lib/kubelet/pods
+            path: {{ .Values.csi.daemonSet.kubeletRootDir }}/pods
        {{- if .Values.csi.volumes }}
          {{- toYaml .Values.csi.volumes | nindent 8}}
        {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -758,6 +758,7 @@ csi:
     # YAML-formatted multi-line templated string map of the annotations to apply
     # to the daemonSet.
     annotations: {}
+    # Provider host path (must match the CSI provider's path)
     providersDir: "/etc/kubernetes/secrets-store-csi-providers"
 
   pod:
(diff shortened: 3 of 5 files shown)


eyenx:
@jasonodonnell Can you please review this PR?

We can't upgrade the vault integration on a cluster due to it being using differente providersDir paths.

Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vault
-version: 0.15.0
```
benashz, line 3: I believe we have some pending changes to release, so no need to bump the version until then.

Chart.yaml
```diff
@@ -1,6 +1,6 @@
...
+version: 0.15.1
```
benashz, line 3: I believe we have some pending changes to release, so no need to bump the version until then.

templates/csi-daemonset.yaml
```diff
@@ -70,7 +70,7 @@ spec:
       volumes:
         - name: providervol
           hostPath:
-            path: "/etc/kubernetes/secrets-store-csi-providers"
+            path: {{ .Values.csi.daemonSet.providersDir }}
         - name: mountpoint-dir
           hostPath:
             path: /var/lib/kubelet/pods
```
benashz, line 76: We should probably make this public as well to be in line with [secrets-store-csi-driver/values.yaml#L70](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/a6f86b798ecb3e301a66153303956c869c882c0f/manifest_staging/charts/secrets-store-csi-driver/values.yaml#L70)

values.yaml
```diff
@@ -758,6 +758,7 @@ csi:
     # YAML-formatted multi-line templated string map of the annotations to apply
     # to the daemonSet.
     annotations: {}
+    providersDir: "/etc/kubernetes/secrets-store-csi-providers"
```
benashz, line 761:
```suggestion
    # CSI provider host path
    providersDir: "/etc/kubernetes/secrets-store-csi-providers"
```

Chart.yaml
eyenx: 1933561

templates/csi-daemonset.yaml
eyenx: :+1:

eyenx: b52d1f6

eyenx:
Thanks for the review

Please re-review @benashz

benashz:
Thanks @eyenx . I think all that is missing now are some unit tests. I think the following patch suffice:

```patch
diff --git a/test/unit/csi-daemonset.bats b/test/unit/csi-daemonset.bats
index c546d0a..7803e06 100644
--- a/test/unit/csi-daemonset.bats
+++ b/test/unit/csi-daemonset.bats
@@ -315,6 +315,38 @@ load _helpers
   [ "${actual}" = "{}" ]
 }
 
+@test "csi/daemonset: csi override providersDir" {
+  cd `chart_dir`
+
+  # Test that it defines it
+  local object=$(helm template \
+      --show-only templates/csi-daemonset.yaml  \
+      --set 'csi.enabled=true' \
+      --set 'csi.daemonSet.providersDir=/alt/csi-prov-dir' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.volumes[] | select(.name == "providervol")' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+      yq -r '.hostPath.path' | tee /dev/stderr)
+  [ "${actual}" = "/alt/csi-prov-dir" ]
+}
+
+@test "csi/daemonset: csi override kubeletRootDir" {
+  cd `chart_dir`
+
+  # Test that it defines it
+  local object=$(helm template \
+      --show-only templates/csi-daemonset.yaml  \
+      --set 'csi.enabled=true' \
+      --set 'csi.daemonSet.kubeletRootDir=/alt/kubelet-root' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.volumes[] | select(.name == "mountpoint-dir")' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+      yq -r '.hostPath.path' | tee /dev/stderr)
+  [ "${actual}" = "/alt/kubelet-root/pods" ]
+}
+
 #--------------------------------------------------------------------
 # volumeMounts
```

eyenx:
thanks @benashz, added your suggested changes in b2af383

benashz (approved):
LGTM!

Thank you for your  contribution to HashiCorp!

eyenx:
My pleasure

PR #604, opened 2021-08-27
Bump Version to 1.8.2

Sevenlive:
Hey everyone,

The Vault Version 1.8.2 fixed some serious issues that could lead to panic (https://github.com/hashicorp/vault/pull/12379). Because of that, the pods start to run into a crash-loops. To fix that, the Helmfile should be updated to the newest Version 1.8.2.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: vault
-version: 0.15.0
-appVersion: 1.8.1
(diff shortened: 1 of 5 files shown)


bobthebuilderberlin:
Would be very much appreciated, since 1.8.2 contains a bug fix that we need.

tvoran:
Superseded by #612

PR #568, opened 2021-07-10
Support vault-k8s internal leader election

tvoran:
Adds extra RBAC and removes leader-elector container deployment to support https://github.com/hashicorp/vault-k8s/pull/271.

PR #607, opened 2021-09-03
Adding support for the old leader-elector

tvoran:
Adds an option to deploy the old leader-elector container that was removed in PR #568. (The new vault-k8s uses an internal mechanism for leader determination, so this is just for backwards compatibility, and can be removed in the near future.)

Note that this defaults to **not** deploying the old leader-elector container and Endpoint, unless `injector.leaderElector.useContainer` is set to `true`.


templates/injector-leader-endpoint.yaml
```diff
@@ -0,0 +1,12 @@
...
+# This is created here so it can be cleaned up easily, since if
+# the endpoint is left around the leader won't expire for about a minute.
+apiVersion: v1
+kind: Endpoints
+metadata:
+  name: {{ template "vault.fullname" . }}-agent-injector-leader
+  labels:
+    app.kubernetes.io/name: {{ include "vault.name" . }}-agent-injector
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
```
benashz, line 11: I wonder if adding an extra annotation denoting the deprecation of this container would be a good idea. Not sure if that would satisfy what @calvn requested w.r.t. facilitating debugging.

tvoran: Sure, something like `deprecated: true`?

I also added a change to the logging in https://github.com/hashicorp/vault-k8s/pull/297 to make it easier to tell which leader elector scheme is in use.

benashz: Sure, that sounds good.

tomhjp (approved):
Shall we pick a version now that we'll remove it again, so we can give advanced notice in the next changelog?

tvoran:
@tomhjp That sounds like a good idea; how about v0.18.0? https://github.com/hashicorp/vault-helm/blob/master/CHANGELOG.md

tomhjp:
@tvoran v0.18.0 sounds good to me, gives a nice amount of overlap and warning.

PR #608, opened 2021-09-03
Fix the CSI acceptance test

tvoran:
The `master` branch was renamed to `main` for
kubernetes-sigs/secrets-store-csi-driver, so the helm install fails.

### DIFF
--- a/test/acceptance/csi.bats
+++ b/test/acceptance/csi.bats
@@ -9,7 +9,7 @@ load _helpers
   kubectl create namespace acceptance
 
   # Install Secrets Store CSI driver
-  helm install secrets-store-csi-driver https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/charts/secrets-store-csi-driver-0.0.20.tgz?raw=true \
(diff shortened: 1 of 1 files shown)


test/acceptance/csi.bats
```diff
@@ -9,7 +9,8 @@ load _helpers
   kubectl create namespace acceptance
 
   # Install Secrets Store CSI driver
-  helm install secrets-store-csi-driver https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/charts/secrets-store-csi-driver-0.0.20.tgz?raw=true \
+  CSI_DRIVER_VERSION=0.0.22
```
tomhjp, line 12: May as well update to the latest version, 0.2.0, while we're here. There shouldn't be any breaking changes, but happy to save the update for another PR if there are.

tomhjp (approved):

test/acceptance/csi.bats
tvoran: Worked for me, changed to 0.2.0 in 1e3d739

PR #610, opened 2021-09-09
Add support for nodePort for `vault-active` and `vault-standby` services

jzck:
When using `ha.enabled = true` and `service.nodePort = 30001` in conjunction, the ports clash because it can only be allocated to a single service (`vault`), therefore it can't also be allocated to the `vault-active` and `vault-standby` service:

```
Error: cannot patch "vault-standby" with kind Service: Service "vault-standby" is invalid: spec.ports[0].nodePort: Invalid value: 30001: provided port is already allocated && cannot patch "vault" with kind Service: Service "vault" is invalid: spec.ports[0].nodePort: Invalid value: 30001: provided port is already allocated
```

### DIFF
--- a/templates/server-ha-active-service.yaml
+++ b/templates/server-ha-active-service.yaml
@@ -26,8 +26,8 @@ spec:
     - name: {{ include "vault.scheme" . }}
       port: {{ .Values.server.service.port }}
       targetPort: {{ .Values.server.service.targetPort }}
-      {{- if and (.Values.server.service.nodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
-      nodePort: {{ .Values.server.service.nodePort }}
+      {{- if and (.Values.server.service.activeNodePort) (eq (.Values.server.service.type | toString) "NodePort") }}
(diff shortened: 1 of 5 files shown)


jzck:
FYI, this PR is addressing the same issue as: https://github.com/hashicorp/vault-helm/pull/463.

Except here we define customizable `nodePort`s for each service instead of doing random, as suggested by https://github.com/hashicorp/vault-helm/pull/463 (cc @jasonodonnell )

samgabrail (approved):
Looks good to me.

gsanchez-uz (approved):
I changed same code for resolve this issue on my clusters

srosell-uz (approved):

gsanchez-uz (approved):

gsanchez-uz (approved):

swenson (approved):
Thanks for this.

This slightly breaks backwards compatibility I think, since we don't fall back to the `.Values.server.service.nodePort`. However, I think that would just keep the buggy behavior, which we don't want.

So, I think we should go ahead and go forward with merging this.

PR #612, opened 2021-09-16
vault-helm 0.16.0 release

tvoran:
CHANGES:

* Support for deploying a leader-elector container with the vault-k8s injector injector will be removed in version 0.18.0 of this chart since vault-k8s now uses an internal mechanism to determine leadership. To enable the deployment of the leader-elector container for use with vault-k8s 0.12.0 and earlier, set useContainer=true.
* Default to vault 1.8.2 for `server.image` and `injector.agentImage`
* Default to vault-k8s 0.13.0 for `injector.image`

Improvements:

* Make CSI provider hostPaths configurable via csi.daemonSet.providersDir and csi.daemonSet.kubeletRootDir GH-603
* Support vault-k8s internal leader election GH-568 GH-607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+## 0.16.0 (September 16th, 2021)
+
 CHANGES:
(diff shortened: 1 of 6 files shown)


benashz (approved):
Ship it!

PR #613, opened 2021-09-16
load balancer for vault active service

### LINKED ISSUE #420: Expose Vault with a Service according to Vault's reference architecture
I'm trying to implement the [Vault reference architecture](https://learn.hashicorp.com/tutorials/vault/kubernetes-reference-architecture). There are some recommendations which aren't possible to follow with this chart alone.

For [exposing the Vault service](https://learn.hashicorp.com/tutorials/vault/kubernetes-reference-architecture#exposing-the-vault-service) outside of a dedicated cluster, none of the existing Services (vault, vault-standby, vault-active or vault-ui) are suitable:

* None of the services allow you to set spec.externalTrafficPolicy which should be set to "local"
* Only vault-ui is supported with type LoadBalancer (with chosen loadBalancerIP), but I may want to run my Vault cluster with ui=off. While the service would be fine, this would be quite misleading.

Note that Ingress is also recommended against.

I'd like if the chart defined an optional Service (perhaps named vault-lb) which is:

* type: "LoadBalancer"
* configurable loadBalancerIP
* spec.externalTrafficPolicy: "local"
* selects active Vault node

Happy to make a PR if this is considered desirable.

If all these settings were configurable with values.yaml, this Service would make it possible to deprecate vault-ui.

Could add spec.externalTrafficPolicy to vault-ui, and then there's enough in values.yaml to configure it to meet the recommendations, but the name is misleading if you don't want to turn Vault's UI on. At the moment, I have to apply a manifest to define a Service as per the reference documentation after installing the chart.

TylerWanner:
provides capability to run without UI by adding a configurable LoadBalancer service to vault active
closes #420

### DIFF
--- a/templates/server-ha-active-service.yaml
+++ b/templates/server-ha-active-service.yaml
@@ -18,6 +18,18 @@ spec:
   {{- if .Values.server.service.type}}
   type: {{ .Values.server.service.type }}
   {{- end}}
+  {{- if .Values.server.service.externalTrafficPolicy }}
+  externalTrafficPolicy: {{ .Values.server.service.externalTrafficPolicy }}
+  {{- end }}
+  {{- if and (eq (.Values.server.serviceType | toString) "LoadBalancer") (.Values.server.loadBalancerSourceRanges) }}
+  loadBalancerSourceRanges:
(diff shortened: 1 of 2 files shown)


TylerWanner:
resolved by #626

PR #621, opened 2021-10-05
add staticSecretRenderInterval to injector

### LINKED ISSUE #611: Add support for templateConfig.staticSecretRenderInterval
**Description of the problem**
After the release of Vault 1.8.0, it's been added the possibility to control the render/refresh time for secrets done by the vault-agent:
https://www.vaultproject.io/docs/agent/template-config#static_secret_render_interval
https://www.vaultproject.io/docs/release-notes/1.8.0

Nevertheless, there's no way to introduce that value directly through the attribute `injector.agentDefaults`, and you have to use a different approach, by introducing an environment variable named `AGENT_INJECT_TEMPLATE_STATIC_SECRET_RENDER_INTERVAL`.

**Description of the solution**
A way to directly introduce that value as in the example:
```
injector:
  agentDefaults:
    templateConfig:
      staticSecretRenderInterval: "30s"
```

kaitoii11:
fix #611 
Add support for Static Secret Render Interval in Template Config.

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -109,6 +109,8 @@ spec:
               value: "{{ .Values.injector.agentDefaults.template }}"
             - name: AGENT_INJECT_TEMPLATE_CONFIG_EXIT_ON_RETRY_FAILURE
               value: "{{ .Values.injector.agentDefaults.templateConfig.exitOnRetryFailure }}"
+            - name: AGENT_INJECT_TEMPLATE_CONFIG_STATIC_SECRET_RENDER_INTERVAL
+              value: "{{ .Values.injector.agentDefaults.templateConfig.staticSecretRenderInterval }}"
             {{- include "vault.extraEnvironmentVars" .Values.injector | nindent 12 }}
             - name: POD_NAME
               valueFrom:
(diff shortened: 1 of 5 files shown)


templates/injector-deployment.yaml
```diff
@@ -109,6 +109,8 @@ spec:
               value: "{{ .Values.injector.agentDefaults.template }}"
             - name: AGENT_INJECT_TEMPLATE_CONFIG_EXIT_ON_RETRY_FAILURE
               value: "{{ .Values.injector.agentDefaults.templateConfig.exitOnRetryFailure }}"
+            - name: AGENT_INJECT_TEMPLATE_CONFIG_STATIC_SECRET_RENDER_INTERVAL
```
tvoran, line 112: I believe this environment variable was shortened a bit:

```suggestion
            - name: AGENT_INJECT_TEMPLATE_STATIC_SECRET_RENDER_INTERVAL
```

values.yaml
```diff
@@ -85,6 +85,7 @@ injector:
     # Default values within Agent's template_config stanza.
     templateConfig:
       exitOnRetryFailure: true
+      staticSecretRenderInterval: "5m"
```
tvoran, line 88: I think it might be good to default this to empty `""`, and only add it to the deployment if it's been set by the user. That way it's not overriding the default in Vault Agent.

(The `exitOnRetryFailure` case was a little different since we explicitly wanted a different default setting for Kubernetes vs non-Kubernetes).

tvoran:
This looks great! Would you mind adding another test to check that `AGENT_INJECT_TEMPLATE_STATIC_SECRET_RENDER_INTERVAL` isn't set on the deployment by default?

tvoran (approved):
Thanks!

PR #640, opened 2021-10-27
Move the volumeMounts section back to the right position again.

dreamblack86:
The commit:
d27121 Added webhook-certs volume mount to sidecar injector (#545)
had it fixed correctly in the past.

But by the short-term decision to take the leader elector container out then back in has "probably" led to this circumstance during copy&paste that the volumemount has moved down again.

see changes how it came to this small error.
remove leader-elector: d31f94 Support vault-k8s internal leader election (#568)
bring leader-elector back: 5a864f Adding support for the old leader-elector (#607)

### DIFF
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -137,6 +137,12 @@ spec:
             periodSeconds: 2
             successThreshold: 1
             timeoutSeconds: 5
+{{- if .Values.injector.certs.secretName }}
+          volumeMounts:
+            - name: webhook-certs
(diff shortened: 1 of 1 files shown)


tvoran:
Thanks for spotting this and for the patch, though if webhook certs are specified, then the leader elector container shouldn't be enabled anyway. Also for the next release we're planning to remove the leader-elector container deployment again.

tvoran:
Closing for now, since the leader-elector container was removed from the deployment in 0.18.0.

PR #642, opened 2021-11-04
fix missing dataStorage when ha_storage:raft is enabled

JubbaSmail:
When using Integrated Storage for HA Coordination as explained here:
https://learn.hashicorp.com/tutorials/vault/raft-ha-storage

Helm configuration need to be something like:
```
server:
  ha:
    enabled: true
    config: |
      ui = true

      listener "tcp" {
        tls_disable = 1
        address = "[::]:8200"
        cluster_address = "[::]:8201"
      }
      
      ha_storage "raft" {
        path = "/vault/ha_data"
        node_id = "HOSTNAME"       
      }

      storage "s3" {
        bucket     = "BUCKET_NAME"
      }

      api_addr = "[::]:8200"
      cluster_addr = "[::]:8201"

  dataStorage:
    enabled: true
    mountPath: "/vault/ha_data"
```

however, in the current template, dataStorage will only be added if **ha.raft.enabled** is **true**,
and since ha_storage need to be added in the **config** so dataStorage will not be created.

Removing the raft condition `(eq (.Values.server.ha.raft.enabled | toString) "true"))` will fix the issue

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -154,7 +154,7 @@ based on the mode configured.
             - name: audit
               mountPath: {{ .Values.server.auditStorage.mountPath }}
   {{ end }}
-  {{ if or (eq .mode "standalone") (and (eq .mode "ha") (eq (.Values.server.ha.raft.enabled | toString) "true"))  }}
+  {{ if or (eq .mode "standalone") (eq .mode "ha") }}
     {{ if eq (.Values.server.dataStorage.enabled | toString) "true" }}
             - name: data
               mountPath: {{ .Values.server.dataStorage.mountPath }}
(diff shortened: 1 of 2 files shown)


JubbaSmail:
@benashz can you take a look?

JubbaSmail:
@benashz can you check this PR?

tvoran:
Hi @JubbaSmail, thanks for the patch! Though I don't think this is necessary when using integrated storage, since the config in that case can be modified from the [default](https://github.com/hashicorp/vault-helm/blob/1eb93ccd77814c03b4930e5bf32c5e4a463159d2/values.yaml#L642) by setting [`server.ha.raft.config`](https://www.vaultproject.io/docs/platform/k8s/helm/configuration#config-1). So in your example, the values overrides could looks something like this:

```yaml
server:
  ha:
    enabled: true
    raft:
      enabled: true
      config: |
        ui = true

        listener "tcp" {
          tls_disable = 1
          address = "[::]:8200"
          cluster_address = "[::]:8201"
        }

        api_addr = "[::]:8200"
        cluster_addr = "[::]:8201"

        ... 
```

There's also a simple example here that uses the default config: https://www.vaultproject.io/docs/platform/k8s/helm/examples/ha-with-raft

PR #647, opened 2021-11-12
allow extraInitContainers to be a template

tareqhs:
A container can be a comlpex structure that itself needs templating to allow both additional customization and reuse of existing values in the chart. Interpret the value of `server.extraInitContainers` as a template if its value is a string, a method that already applies to affinity, tolerations, annotations, ...


tareqhs:
hi, any update on this?

PR #648, opened 2021-11-15
Update stakater vault chart

sabkat:


values.openshift.yaml
```diff
@@ -0,0 +1,18 @@
+# These overrides are appropriate defaults for deploying this chart on OpenShift
```
hanzala1234, line 1: why this file?

templates/server-route.yaml
```diff
@@ -14,20 +15,25 @@ metadata:
...
     kind: Service
     name: {{ $serviceName }}
     weight: 100
   port:
     targetPort: 8200
+  {{- if $route.tls.enabled }}
   tls:
-    termination: passthrough
+    insecureEdgeTerminationPolicy: {{ $route.tls.insecureEdgeTerminationPolicy }}
+    termination: {{ $route.tls.termination }}
```
hanzala1234, line 35: we will need to provide tls termination policy in values.yaml, seems like its  missing

templates/server-route.yaml
```diff
@@ -14,20 +15,25 @@ metadata:
     app.kubernetes.io/name: {{ include "vault.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
-    {{- with .Values.server.route.labels }}
+    {{- with $route.labels }}
```
hanzala1234, line 18: I dont see this function in helper.tpl

sabkat: Its not a function. Its a variable defined at the top of the file taking value from Values.server.route

values.openshift.yaml
sabkat: Deleting

templates/server-route.yaml
sabkat: What should I provide. The line in red ?

templates/server-route.yaml
hanzala1234: ok. thanks for explanation

PR #651, opened 2021-11-18
allow extraContainers to be set as a template string

tareqhs:
A container can be a complex structure that itself needs templating to allow both additional customization and reuse of existing values in the chart. Interpret the value of `server.extraContainers` as a template if its value is a string.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -200,7 +200,11 @@ spec:
                 {{- end }}
             {{- end }}
         {{- if .Values.server.extraContainers }}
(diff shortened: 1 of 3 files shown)


tvoran:
@tareqhs Thanks for the patch, this looks pretty straightforward. I'm curious what you think about the approach in #536?

tareqhs:
hi @tvoran the other PR is problematic for backward compatibility as some existing deployments might already have the value in yaml structure. i've go for a dual approach that allows the value to be both a template string and a yaml structure.

tareqhs:
hi @tvoran , any update here?

AdrianFarmadin:
Hi, any update?

PR #653, opened 2021-11-23
Configurable PodDisruptionBudget for Injector

sosheskaz:
This adds a section to values which can configure a PodDisruptionBudget for Injector.

I saw no reason not to configure it free-form, since it's propagating to the Kubernetes API rather than being internally significant. Please let me know if we'd prefer to change how that works.

Example:

```
helm template --show-only templates/injector-disruptionbudget.yaml --set 'injector.podDisruptionBudget.maxUnavailable=1' .
```

```yaml
---
# Source: vault/templates/injector-disruptionbudget.yaml
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: RELEASE-NAME-vault-agent-injector
  namespace: inst-system
  labels:
    helm.sh/chart: vault-0.18.0
    app.kubernetes.io/name: vault-agent-injector
    component: webhook
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: vault-agent-injector
      component: webhook
  maxUnavailable: 1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+Features:
+* Added configurable podDisruptionBudget for injector [GH-653](https://github.com/hashicorp/vault-helm/pull/653)
+
 ## 0.18.0 (November 17th, 2021)
 
 CHANGES:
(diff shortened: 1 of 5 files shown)


tvoran (approved):
This looks great! Thanks so much.

PR #654, opened 2021-11-23
Fix spelling error in server disruptionbudget test

sosheskaz:
Fix a spelling error in the unit test which set a variable which does not exist.

There is no functional change, and no change to the result of the test.


tvoran (approved):
Good spot!

PR #655, opened 2021-11-26
Add support for globally disabling RBAC setup

mprasil:
In some more restrictive environments the Vault admins might not have
access required to let Helm create the RBAC configuration and instead
rely on that being managed some other way.

This change lets users disable such RBAC configuration globally with
`global.rbac=false` with the assumption that appropriate RBAC
configuration is set some other way prior to chart deployment.


mprasil:
Folks any update on this?

mprasil:
Closing this as it got no traction and we had to find another approach for this.

PR #657, opened 2021-12-02
Run CI tests in github workflows 

tvoran:
Ports the bats unit, chart-verifier, and bats acceptance tests to use github workflows and actions. The acceptance tests run using kind, and run for multiple k8s versions, on pushes to the main branch.

Example run of the kind acceptance tests can be seen here: https://github.com/hashicorp/vault-helm/actions/runs/1537925214

- Adds a `SKIP_CSI` env check in the CSI acceptance test, set in the workflow if K8s version is less than 1.16
- Adds `kubeAdmConfigPatches` to the kind config to allow testing the CSI provider on K8s versions prior to 1.21
- Updates the Secrets Store CSI driver to 1.0.0 in tests
- Makes the HA Vault tests more robust by waiting for all consul client pods to be Ready, and waits with a timeout for Vault to start responding as sealed (since the tests on GitHub runners were often failing at that point)


.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,34 @@
...
+    branches:
+      - main
+  workflow_dispatch: {}
+
+jobs:
+  kind:
+    strategy:
+      fail-fast: false
+      matrix:
+        kind-k8s-version: [1.14.10, 1.19.11, 1.22.4]
```
tvoran, line 14: The thinking here is 1.14 is the minimum version, 1.22 is the max, and I wanted to include an in-between version that was pre-1.21 but more recent than 1.14.

tomhjp: How about 1.14 and the three most recent versions? Then it includes our oldest version + roughly approximates the support windows for major managed Kubernetes providers [1](https://cloud.google.com/kubernetes-engine/docs/release-notes) [2](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html) [3](https://docs.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#kubernetes-version-support-policy). Kubernetes releases every 3 months and each version is [supported for 12 months](https://kubernetes.io/blog/2020/08/31/kubernetes-1-19-feature-one-year-support/), so maybe even the 4 most recent versions would make sense?

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup test tools
+        uses: ./.github/workflows/setup-test-tools
```
tomhjp, line 19: This is really neat!

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Create K8s Kind Cluster
+        uses: helm/kind-action@v1.2.0
```
tomhjp, line 22: This is great, will be useful for converting the CSI provider CI as well.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    env:
+      CHART_VERIFIER_VERSION: '1.2.1'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup test tools
+        uses: ./.github/workflows/setup-test-tools
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.17.2'
```
tomhjp, line 23:
```suggestion
          go-version: '1.17.4'
```

test/acceptance/csi.bats
```diff
@@ -2,18 +2,27 @@
...
+
   cd `chart_dir`
   
   kubectl delete namespace acceptance --ignore-not-found=true
   kubectl create namespace acceptance
 
   # Install Secrets Store CSI driver
-  CSI_DRIVER_VERSION=0.2.0
-  helm install secrets-store-csi-driver https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/v${CSI_DRIVER_VERSION}/charts/secrets-store-csi-driver-${CSI_DRIVER_VERSION}.tgz?raw=true \
+  CSI_DRIVER_VERSION=1.0.0
```
tomhjp, line 20: Thanks for updating this

test/kind/config.yaml
```diff
@@ -5,3 +5,16 @@ nodes:
...
+# prior to 1.21
+kubeadmConfigPatches:
+  - |
+    apiVersion: kubeadm.k8s.io/v1beta2
+    kind: ClusterConfiguration
+    metadata:
+      name: config
+    apiServer:
+      extraArgs:
+        "service-account-issuer": "kubernetes/serviceaccount"
```
tomhjp, line 18:
```suggestion
        "service-account-issuer": "https://kubernetes.default.svc.cluster.local"
```

I would err towards providing a non-legacy value unless I'm missing something.

test/kind/config.yaml
```diff
@@ -5,3 +5,16 @@ nodes:
...
+        "service-account-signing-key-file": "/etc/kubernetes/pki/sa.key"
+        "service-account-api-audiences": "https://kubernetes.default.svc"
```
tomhjp, line 20:
```suggestion
        "service-account-api-audiences": "https://kubernetes.default.svc.cluster.local"
```

tomhjp (approved):
Awesome!

test/kind/config.yaml
tvoran: Yep, agreed. Done in 01af2f52355bd1f3960e19e1ed6a32e7975fcd72.

tomhjp (approved):

.github/workflows/acceptance.yaml
tvoran: Yeah, that makes sense. I went with 1.14 and the 4 most recent minor versions. It turns out kind node images aren't published for every patch release, so I used the latest patches I could find for each in e29aae9.

PR #659, opened 2021-12-09
Make terminationGracePeriodSeconds configurable

### LINKED ISSUE #522: Configurable terminationGracePeriodSeconds for Vault Server
During GKE automatic node pool upgrades, we are seeing an issue where the upgrade happens too quickly, and the vault-active service goes down for several minutes because too many nodes are being brought offline. In the GCP documentation,
https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades#node_pool_upgrades
```
Configurations that can slow the rate of node upgrades include:

A high value of terminationGracePeriodSeconds in a Pod's configuration.
```
The first thing that is suggested is raising the terminationGracePeriodSeconds. Right now it is set to 10 seconds and does not appear to be configurable:
https://github.com/hashicorp/vault-helm/blob/f226051e53227521beded82442baf97ea28d666f/templates/server-statefulset.yaml#L44
Ten seconds seems quick, it is not necessary for it to be so fast in our situation, and may increase reliability.

Just a simple configuration option in values.yaml for the `server.terminationGracePeriodSeconds`

None so far, but I am open to anything.

**Additional context**
Just putting it in the Backlog, seems useful.

mikutas:
for server pod

close #522

### DIFF
--- a/values.schema.json
+++ b/values.schema.json
@@ -14,6 +14,12 @@
                                 "string"
                             ]
                         },
+                        "kubeletRootDir": {
+                            "type": "string"
+                        },
+                        "providersDir": {
+                            "type": "string"
+                        },
                         "updateStrategy": {
                             "type": "object",
                             "properties": {
@@ -24,12 +30,6 @@
                                     "type": "string"
                                 }
                             }
-                        },
-                        "providersDir": {
-                          "type": "string"
-                        },
-                        "kubeletRootDir": {
-                          "type": "string"
                         }
                     }
                 },
@@ -600,6 +600,9 @@
                         "labels": {
                             "type": "object"
                         },
+                        "pathType": {
+                            "type": "string"
+                        },
                         "tls": {
                             "type": "array"
                         }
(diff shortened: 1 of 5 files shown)


tvoran (commented):
Thanks for the patch! Would you mind adding tests for this parameter in `test/unit/server-statefulset.bats`?

tvoran (commented):
Can you add one more test that sets terminationGracePeriodSeconds? And would you also mind adding terminationGracePeriodSeconds to [values.schema.json](https://github.com/hashicorp/vault-helm/blob/0375b184b33393bb013431ff3717296e30bff713/values.schema.json#L781)?

values.schema.json
```diff
@@ -14,6 +14,12 @@
                                 "string"
                             ]
                         },
+                        "kubeletRootDir": {
+                            "type": "string"
+                        },
+                        "providersDir": {
+                            "type": "string"
+                        },
```
tvoran, line 22: Thanks for updating all these missing parameters!

tvoran (approved):
Look great, thanks again!

Issue #490, opened 2021-04-04
OpenShift route has serviceport and tls hardcoded, and tls is incorrect if endpoint is not secure


rgordill:
**Describe the bug**
service-route.yaml has servicePort hardcoded (8200), and should be equivalent of service-ingress.yaml. Additionally, passthrough is selected for tls termination, but by default 8200 is not a secure port, so same as ingress should apply.

**To Reproduce**
1. Install chart in OpenShift with server.route.enabled=true and ui.enabled=true
2. Try to access the url

When route tls is deleted, the ui is accessed without any issues.

**Expected behavior**
Route configuration consistent with ingress.

**Environment**
* Kubernetes version: 
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): OpenShift 4.7
  * Other configuration options or runtime services (istio, etc.): N/A
* vault-helm version: 0.10.0

Chart values:

```yaml
csi:
  daemonSet:
    annotations: {}
    updateStrategy:
      maxUnavailable: ""
      type: RollingUpdate
  debug: false
  enabled: false
  image:
    pullPolicy: IfNotPresent
    repository: hashicorp/vault-csi-provider
    tag: 0.1.0
  livenessProbe:
    failureThreshold: 2
    initialDelaySeconds: 5
    periodSeconds: 5
    successThreshold: 1
    timeoutSeconds: 3
  pod:
    annotations: {}
  readinessProbe:
    failureThreshold: 2
    initialDelaySeconds: 5
    periodSeconds: 5
    successThreshold: 1
    timeoutSeconds: 3
  resources: {}
  serviceAccount:
    annotations: {}
  volumeMounts: null
  volumes: null
global:
  enabled: true
  imagePullSecrets: []
  openshift: true
  psp:
    annotations: |
      seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default,runtime/default
      apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
      seccomp.security.alpha.kubernetes.io/defaultProfileName:  runtime/default
      apparmor.security.beta.kubernetes.io/defaultProfileName:  runtime/default
    enable: false
  tlsDisable: true
injector:
  affinity: |
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              app.kubernetes.io/name: {{ template "vault.name" . }}-agent-injector
[...]

PR #671, opened 2022-01-07
Allow the route termination to be customized

mbaldessari:
This covers, in part, issue #490. Namely, the route currently hardcodes
the termination mode to "passhtrough". Let's parametrize this so a user
can customize it. The use-case here being that we can deploy the vault
in http mode and expose it via https externally and by choosing 'edge'
as the termination type we let the TLS be taken care of by the route
object.

The default remains 'passthrough' in order to not disrupt any existing
setup.

### DIFF
--- a/templates/server-route.yaml
+++ b/templates/server-route.yaml
@@ -27,7 +27,7 @@ spec:
   port:
     targetPort: 8200
   tls:
-    termination: passthrough
+    termination: {{ .Values.server.route.termination }}
 {{- end }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


tvoran:
Thanks for the patch! I'm thinking we might as well allow users to set the whole `tls` block as YAML, like the approach taken in #396. What do you think?

mbaldessari:
Thanks Theron! Absolutely. #396 covers more, so let's go for that one I'll close this one out

Issue #665, opened 2021-12-22
Test Helm Hook Pod should be more configurable


Iristyle:
Depending on how Vault TLS certificates are created, it's possible for the [server-test Pod](https://github.com/hashicorp/vault-helm/blob/main/templates/tests/server-test.yaml) to not have access to the CA cert, preventing the tests from passing. This is causing CI failures for us because we wait for the Vault rollout to complete successfully. In production, we may see failing pods that frankly, we don't care about (since we have other checks to verify Vault is operational)

There are a number of features which would make the server-test Pod more useful:

* Ability to completely omit it from being generated
* Ability to specify additional environment variables (like `VAULT_CACERT` / `VAULT_CAPATH`) - see https://www.vaultproject.io/docs/commands#environment-variables
* Ability to mount extra volumes (i.e. with the CA cert)

Currently, the only workaround is a bit hacky and involves using Kustomize to delete the Pod resource. Unfortunately we only have this option in CI, but not in production, because we also use the KOTS tool from Replicated.

mshaposhnik:
So, any option to disable that pod at all?

PR #673, opened 2022-01-07
Add volumes and env vars to helm hook test pod

Iristyle:
- Fix test typo
 - Add basic server-test Pod tests 
     - This covers all existing functionality that matches what's
        present in server-statefulset.bats
 - Fix server-test helm hook Pod rendering 
     - Properly adhere to the global.enabled flag and the presence of
   the injector.externalVaultAddr setting, the same way that
   the servers StatefulSet behaves
 - Uses the same extraEnvironmentVars, volumes and volumeMounts set on
   the server statefulset to configure the Vault server test pod used by
   the helm test hook
 - This is necessary in situations where TLS is configured, but the
   certificates are not affiliated with the k8s CA / part of k8s PKI

 - Fixes GH-665

### DIFF
--- a/templates/tests/server-test.yaml
+++ b/templates/tests/server-test.yaml
@@ -1,4 +1,6 @@
-{{- if .Values.server.enabled }}
+{{ template "vault.mode" . }}
+{{- if ne .mode "external" }}
+{{- if and (ne .mode "") (eq (.Values.global.enabled | toString) "true") }}
 apiVersion: v1
 kind: Pod
 metadata:
@@ -37,4 +39,5 @@ spec:
           exit 0
 
   restartPolicy: Never
+{{- end }}
 {{- end }}
--- a/templates/tests/server-test.yaml
+++ b/templates/tests/server-test.yaml
@@ -17,6 +17,7 @@ spec:
       env:
         - name: VAULT_ADDR
           value: {{ include "vault.scheme" . }}://{{ template "vault.fullname" . }}.{{ .Release.Namespace }}.svc:{{ .Values.server.service.port }}
+        {{- include "vault.extraEnvironmentVars" .Values.server | nindent 8 }}
       command:
         - /bin/sh
         - -c
@@ -37,7 +38,14 @@ spec:
           fi
 
           exit 0
-
+      volumeMounts:
+        {{- if .Values.server.volumeMounts }}
+          {{- toYaml .Values.server.volumeMounts | nindent 8}}
+        {{- end }}
+  volumes:
+    {{- if .Values.server.volumes }}
+      {{- toYaml .Values.server.volumes | nindent 4}}
+    {{- end }}
   restartPolicy: Never
 {{- end }}
 {{- end }}
--- a/test/unit/server-test.bats
+++ b/test/unit/server-test.bats
@@ -66,6 +66,28 @@ load _helpers
   [ "${actual}" = "true" ]
 }
 
+@test "server/standalone-server-test-Pod: disable with global.enabled" {
+  cd `chart_dir`
+  local actual=$( (helm template \
+      --show-only templates/tests/server-test.yaml  \
+      --set 'global.enabled=false' \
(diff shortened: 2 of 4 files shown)


templates/tests/server-test.yaml
```diff
@@ -35,6 +37,9 @@ spec:
           fi
 
           exit 0
-
+      volumeMounts:
+        {{ template "vault.mounts" . }}
```
Iristyle, line 41: This drags in the `audit` and `data` volumes which doesn't make sense... shooting for something simpler

Iristyle:
I've verified this fixes the TLS issues in the test chart. Logs now look like:

```
Checking for sealed info in 'vault status' output                                                                                 
Attempt 0...                                                                                                                      
Error checking seal status: Get "https://cd4pe-vault.default.svc:8200/v1/sys/seal-status": dial tcp 10.106.187.188:8200: connect: 
Attempt 1...                                                                                                                      
Error checking seal status: Get "https://cd4pe-vault.default.svc:8200/v1/sys/seal-status": dial tcp 10.106.187.188:8200: connect: 
Attempt 2...                                                                                                                      
Error checking seal status: Get "https://cd4pe-vault.default.svc:8200/v1/sys/seal-status": dial tcp 10.106.187.188:8200: connect: 
Attempt 3...                                                                                                                      
Error checking seal status: Get "https://cd4pe-vault.default.svc:8200/v1/sys/seal-status": dial tcp 10.106.187.188:8200: connect: 
Attempt 4...                                                                                                                      
sealed: true
```

tvoran (commented):
Would you mind adding unit tests for these options? I think it's probably time we added a `test/unit/server-test.bats`. Should be able to use server-statefulset.bats for examples of testing `extraEnvironmentVars`, `volumeMounts`, and `volumes`.

templates/tests/server-test.yaml
```diff
@@ -35,6 +36,13 @@ spec:
...
           exit 0
-
+      volumeMounts:
+        {{- if .Values.server.volumeMounts }}
+          {{- toYaml .Values.server.volumeMounts | nindent 12}}
+        {{- end }}
+  volumes:
+    {{- if .Values.server.volumes }}
+      {{- toYaml .Values.server.volumes | nindent 8}}
+    {{- end }}
```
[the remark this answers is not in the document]
Iristyle: I can re-render and check... but I also did verify that the current indentation works in a live deployment to map the volumes in question.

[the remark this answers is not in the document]
Iristyle: You were right on the indentation rendering wrong -- verified with

> helm template --show-only templates/tests/server-test.yaml --set 'server.volumes[0].name=plugins' --set 'server.volumes[0].emptyDir=\{\}' --set 'server.volumeMounts[0].name=plugins' --set 'server.volumeMounts[0].mountPath=/usr/local/libexec/vault' --set 'server.volumeMounts[0].readOnly=true' .

Iristyle:
Follow-on PR at #678 brings some more feature parity to the test pod

test/unit/server-test.bats
```diff
@@ -0,0 +1,231 @@
...
+
+@test "server/standalone-server-test-Pod: enable with server.standalone.enabled true" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/tests/server-test.yaml  \
+      --set 'server.standalone.enabled=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
tvoran, line 57: Would you mind adding one more test like this, but for "server.ha.enabled true"?

tvoran (commented):
Thanks for adding all the tests. Just one more suggestion and I think we'll be good to go.

test/unit/server-test.bats
Iristyle: Done

tvoran (approved):
Thanks!

PR #678, opened 2022-01-13
Adds missing Helm hook test pod features

Iristyle:
Builds on #673 

- imagePullSecrets
- logging config
- securityContext / OpenShift support
- annotations


Iristyle:
Removing draft b/c I'm done with this one... only put it in draft b/c #673 is intended to be merged prior

PR #679, opened 2022-01-14
add namespace support for openshift route

jacobmammoliti:
This change adds the `namespace` key in the OpenShift route metadata so that the Route is deployed to the same namespace as the Vault services/endpoints . Before this change, the route was always put in the default namespace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Features:
 * Added configurable update strategy for injector [GH-661](https://github.com/hashicorp/vault-helm/pull/661)
 * csi: ability to set priorityClassName for CSI daemonset pods [GH-670](https://github.com/hashicorp/vault-helm/pull/670)
 
+Improvements:
+* Set the namespace on the OpenShift Route [GH-679](https://github.com/hashicorp/vault-helm/pull/679)
+
 ## 0.18.0 (November 17th, 2021)
 
 CHANGES:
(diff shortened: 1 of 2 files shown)


tvoran:
Hi @jacobmammoliti, this looks fine, though I'm curious how you're deploying the chart in OpenShift? Since there was no namespace specified in the Route, it should have landed in the same namespace as the chart.

jacobmammoliti:
Hey @tvoran, here is an example below using `helm template`:

Snippet of my `values.yaml` where I enable the route object: 
```
...
server:
  route:
    enabled: true
...
```

Running `helm template vault hashicorp/vault --values values.yaml -n vault` renders this route manifest:
```YAML
# Source: vault/templates/server-route.yaml
kind: Route
apiVersion: route.openshift.io/v1
metadata:
  name: vault
  labels:
    helm.sh/chart: vault-0.17.1
    app.kubernetes.io/name: vault
    app.kubernetes.io/instance: vault
spec:
  host: chart-example.local
  to:
    kind: Service
    name: vault-active
    weight: 100
  port:
    targetPort: 8200
  tls:
    termination: passthrough
```

The Ingress object in our chart behaves the same way. By default, it would get deployed to the default namespace but we have [this](https://github.com/hashicorp/vault-helm/blob/main/templates/server-ingress.yaml#L23) to have deploy to the namespace we are deploying everything else in.

tvoran:
> By default, it would get deployed to the default namespace
Again, I don't believe that's necessarily true. If the object sent to K8s doesn't specify a namespace, it'll land in the namespace specified by the tool. i.e.

```
kubectl apply -n <namespace> ....
helm install -n <namespace> ....
```

And I did double-check this on OpenShift. That's why I was curious how/if you were seeing a chart deployment where the Route ended up in a different namespace than the rest of the chart.

tvoran (approved):

PR #680, opened 2022-01-14
Add labels to deployment, statefulset and service accounts

Iristyle:


Iristyle:
For some reason only the CLA is running and not the normal CircleCI checks... ran all unit tests locally successfully

bmeverett:
Any update on this? I am looking for this addition as well

MeNsaaH:
Same here. Any traction here?

Issue #682, opened 2022-01-15
Role, rolebinding and secret for injector-leader-elector are missing namespace


thechristschn:
**Describe the bug**
The role, rolebinding and secret resources used by the injector-leader-elector are always deployed to the current namespace instead of a given namespace.

This affects role `vault-agent-injector-leader-elector-role`, rolebinding `vault-agent-injector-leader-elector-binding` and secret `vault-injector-certs`. These resources are only deployed if `injector.replicas` is set to more than 1.

While the secret doesn't seem to be important because it will be created in the correct namespace if missing, the role and rolebinding are more important, because the serviceAccount doesn't get the needed permissions to elect a leader. Or maybe to even create the needed certificates.

**To Reproduce**
1. Install chart to a specific namespace with `injector.replicas: 2`
2. Check the logs of the injector. It goes into a `CrashLoopBackOff` because permissions are missing.

```
Using internal leader elector logic for webhook certificate management
E0115 13:07:49.773888       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.22.2/tools/cache/reflector.go:167: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:vault-injector:vault-injector-agent-injector" cannot list resource "secrets" in API group "" in the namespace "vault-injector"
```

**Expected behavior**
All namespaced resources should be deployed to the given namespace and the leader-elector/sidecar-injector works.

**Environment**
* Kubernetes version: 1.21
* vault-helm version: 0.18.0

Chart values:

```yaml
# all default except from:
injector.replicas: 2
```

tvoran:
Hi @thechristschn, I'm not seeing the same behavior you're describing when doing a helm install to a non-default namespace (with injector.replicas > 1); can you share all the values you set on a chart deployment, and the helm command used to deploy?

thechristschn:
Hello @tvoran,

sorry, I missed a essential point on how to reproduce this bug, because I oversimplified the way we deploy the injector (and didn't knew it changes the behavior).
We render the helm chart with `helm template` and apply it afterwards with kubectl. We have additional processing on the helm charts, so we can't just use `helm install`.

So the commands to reproduce the issue:
```
kubectl create ns vault
helm repo add hashicorp https://helm.releases.hashicorp.com
helm template vault hashicorp/vault --set injector.replicas=2 -n vault | kubectl apply -f -
```
[...]

PR #683, opened 2022-01-15
Add namespace to injector-leader-elector role, rolebinding and secret

thechristschn:
This adds namespaces to namespaced resources needed by the injector-leader-elector.

This fixes https://github.com/hashicorp/vault-helm/issues/682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGES:
 
 Improvements:
 * CSI: Set `extraLabels` for daemonset, pods, and service account [GH-690](https://github.com/hashicorp/vault-helm/pull/690)
+* Add namespace to injector-leader-elector role, rolebinding and secret [GH-683](https://github.com/hashicorp/vault-helm/pull/683)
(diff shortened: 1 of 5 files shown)


tvoran (commented):
Thanks! Would you mind adding a unit test for each template here? Just something that checks that the namespace was set correctly on the object, similar to this test: https://github.com/hashicorp/vault-helm/blob/f7ab37fd50e365735916ec71b0d5ef27e564c264/test/unit/injector-mutating-webhook.bats#L34-L43

thechristschn:
> Thanks! Would you mind adding a unit test for each template here? Just something that checks that the namespace was set correctly on the object, similar to this test:
> 
> https://github.com/hashicorp/vault-helm/blob/f7ab37fd50e365735916ec71b0d5ef27e564c264/test/unit/injector-mutating-webhook.bats#L34-L43

Hello @tvoran,
I added the tests.

tvoran (approved):
Looks great, thanks!

PR #684, opened 2022-01-18
Fix values schema to support config in YAML

gw0:
In newest Helm, if you provide `.Values.server.standalone.config` as a YAML object the schema validation fails with:
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
vault:
- server.standalone.config: Invalid type. Expected: string, given: object
```

### DIFF
--- a/values.schema.json
+++ b/values.schema.json
@@ -522,7 +522,10 @@
                             ]
                         },
                         "config": {
-                            "type": "string"
+                            "type": [
+                                "string",
+                                "object"
+                            ]
                         },
                         "disruptionBudget": {
                             "type": "object",
@@ -545,7 +548,10 @@
                             "type": "object",
                             "properties": {
                                 "config": {
-                                    "type": "string"
+                                    "type": [
+                                        "string",
+                                        "object"
+                                    ]
                                 },
                                 "enabled": {
                                     "type": "boolean"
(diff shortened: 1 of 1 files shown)


tomhjp:
Hi @gw0, please could you provide an example of what setting it to a yaml object would look like? Ideally by adding a unit test to make sure we maintain that compatibility in the future too.

gw0:
Converting the configuration from HCL to YAML is pretty straightforward. Basicaly, instead of providing an un-diffable huge string (the helm diff plugin just shows that the whole string changed), you just provide an equivalent YAML code (the helm diff plugin can show what exactly changed). The following used to work before the introduction of values schema.

Instead of a huge HCL-formatted configuration string inside YAML:
```hcl
server:
  standalone:
    enabled: true
    config: |
      disable_mlock = true
      log_format = "json"
      listener "tcp" {
        address = "[::]:8200"
        cluster_address = "[::]:8201"
      }
      storage "file" {
        path = "/vault/data"
      }
```

You provide an equivalent configuration in YAML (note that HCL named blocks are mapped to two-layer YAML dicts):
```yaml
server:
  standalone:
    enabled: true
    config:
      disable_mlock: true
      log_format: "json"
      listener:
        tcp:
          address: "[::]:8200"
          cluster_address: "[::]:8201"
      storage:
        file:
          path: "/vault/data"
```

tomhjp (approved):
Sorry for the delay, but this LGTM. Thanks for the contribution!

PR #688, opened 2022-02-03
fix typo vaules.yaml

rahymov:


swenson:
Okay, it looks like this typo was fixed in a previous PR. Thanks though!

PR #690, opened 2022-02-23
Add extraLabels for CSI DaemonSet

michaeljs1990:
In our environment, we have some required labels in order to be able to deploy some resources including daemonsets. You support this for the server and agent already this just adds it for the csi provider as well.

### DIFF
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -8,6 +8,9 @@ metadata:
     app.kubernetes.io/name: {{ include "vault.name" . }}-csi-provider
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{- if  .Values.csi.extraLabels -}}
+      {{- toYaml .Values.csi.extraLabels | nindent 8 -}}
+    {{- end -}}
   {{ template "csi.daemonSet.annotations" . }}
 spec:
   updateStrategy:
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -8,6 +8,9 @@ metadata:
     app.kubernetes.io/name: {{ include "vault.name" . }}-csi-provider
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{- if  .Values.csi.daemonSet.extraLabels -}}
+      {{- toYaml .Values.csi.daemonSet.extraLabels | nindent 4 -}}
+    {{- end -}}
   {{ template "csi.daemonSet.annotations" . }}
 spec:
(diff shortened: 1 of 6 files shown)


tomhjp (commented):
Thanks! LGTM, just two requests:

* Would it make sense to add `annotations` at the same time?
* Please could you add some tests exercising the new options as well? There are some [examples](https://github.com/hashicorp/vault-helm/blob/e629dc9d65355a1329ae4535be423d03489317c4/test/unit/server-statefulset.bats#L1023-L1064) from the server tests that you can copy heavily from, and they should go into [this file](https://github.com/hashicorp/vault-helm/blob/e629dc9d65355a1329ae4535be423d03489317c4/test/unit/csi-daemonset.bats) for CSI.

michaeljs1990:
> Thanks! LGTM, just two requests:
> 
Thanks for the quick review!

> * Would it make sense to add `annotations` at the same time?

I was looking through the template files and although no parent level annotations exist under the csi key they do for `csi.pod.annotations`, `csi.daemonSet.annotations`, and `csi.serviceAccount.annotations`. It looks like in the other sections such as `injector.annotations` is really just for the pod so it could be `injector.pod.annotations`. It looks like the style changed a bit between how the yaml is defined here so possible the correct way to add extraLabels under csi should be for it to live in `csi.daemonSet.extraLabels`.

I'm not sure if that is the intended way it is meant to be done going forward as it is a little easier to figure out where the annotations and labels are going to end up.

Do you think this makes sense to move under `csi.daemonSet.extraLabels`? I can keep the change isolated to just CSI changes for this diff and then in a follow-up diff try to make it work the same under `injector`, `server`, and `ui` but not breaking the old way to define it but instead adding a note about what one should be used going forward.

> * Please could you add some tests exercising the new options as well? There are some [examples](https://github.com/hashicorp/vault-helm/blob/e629dc9d65355a1329ae4535be423d03489317c4/test/unit/server-statefulset.bats#L1023-L1064) from the server tests that you can copy heavily from, and they should go into [this file](https://github.com/hashicorp/vault-helm/blob/e629dc9d65355a1329ae4535be423d03489317c4/test/unit/csi-daemonset.bats) for CSI.

Will do 👍

michaeljs1990:
Ok, I just went ahead and assumed that the new format was CSI was what was desired now. If this isn't correct let me know 😄 This adds the ability to add extra labels for the daemon set, pods, and service account.

Some unit tests were added as well and all passed. Some random ones failed for me that are unrelated to my change as I ran them off main and they still failed.

```
UNIT_TESTS_FILTER="extraLabels" make test-unit
```

used for testing.

tomhjp (approved):
Oh good point, I forgot about that splitting of annotations. I think how you've got it now looks good to me. Thanks for looking at this so holistically. I think what you mention for the other `extraLabels` configurations makes sense too if you do fancy looking at that.

PR #692, opened 2022-02-24
Mutating Webhook Kubernetes 1.22

RemcoBuddelmeijer:
Mutating Webhook Kubernetes 1.22 consistency pull request to ensure no previous behaviour is changed.

Changes to be done:
1. values.yaml hold default values from MutatingWebhookConfiguration/v1beta1
2. injector-mutating-webhook.yaml sets the default values based on the values.yaml

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -372,13 +372,13 @@ Sets extra injector service annotations
 Sets extra injector webhook annotations
 */}}
 {{- define "injector.webhookAnnotations" -}}
-  {{- if .Values.injector.webhookAnnotations }}
+  {{- if .Values.injector.webhook.webhookAnnotations }}
   annotations:
-    {{- $tp := typeOf .Values.injector.webhookAnnotations }}
+    {{- $tp := typeOf .Values.injector.webhook.webhookAnnotations }}
     {{- if eq $tp "string" }}
-      {{- tpl .Values.injector.webhookAnnotations . | nindent 4 }}
+      {{- tpl .Values.injector.webhook.webhookAnnotations . | nindent 4 }}
     {{- else }}
-      {{- toYaml .Values.injector.webhookAnnotations | nindent 4 }}
+      {{- toYaml .Values.injector.webhook.webhookAnnotations | nindent 4 }}
     {{- end }}
   {{- end }}
 {{- end -}}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -372,13 +372,13 @@ Sets extra injector service annotations
 Sets extra injector webhook annotations
 */}}
 {{- define "injector.webhookAnnotations" -}}
-  {{- if .Values.injector.webhook.webhookAnnotations }}
+  {{- if or (((.Values.injector.webhook)).annotations) (.Values.injector.webhookAnnotations)  }}
   annotations:
-    {{- $tp := typeOf .Values.injector.webhook.webhookAnnotations }}
+    {{- $tp := typeOf (or (((.Values.injector.webhook)).annotations) (.Values.injector.webhookAnnotations)) }}
     {{- if eq $tp "string" }}
-      {{- tpl .Values.injector.webhook.webhookAnnotations . | nindent 4 }}
+      {{- tpl (((.Values.injector.webhook)).annotations | default .Values.injector.webhookAnnotations) . | nindent 4 }}
     {{- else }}
-      {{- toYaml .Values.injector.webhook.webhookAnnotations | nindent 4 }}
+      {{- toYaml (((.Values.injector.webhook)).annotations | default .Values.injector.webhookAnnotations) | nindent 4 }}
     {{- end }}
   {{- end }}
 {{- end -}}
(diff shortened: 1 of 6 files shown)


values.yaml
```diff
@@ -90,34 +90,56 @@ injector:
...
-
-  # Configures failurePolicy of the webhook. The "unspecified" default behaviour deoends on the
-  # API Version of the WebHook.
-  # To block pod creation while webhook is unavailable, set the policy to `Fail` below.
-  # See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy
-  #
-  failurePolicy: Ignore
-
-  # Extra annotations to attach to the webhook
-  webhookAnnotations: {}
```
tomhjp, line 120: While I like the grouping under `webhook`, we try to maintain backwards compatibility for previously written values files. Please could we keep these settings in place? Up to you whether you'd like to duplicate options between here and `webhook`, and comment these ones as deprecated, or just use a flat structure.

tomhjp (commented):
Thanks for raising this! Would love to get this merged. In addition to the backwards compatibility query, please could we add:

* Updates to the schema - you can use the instructions [here](https://github.com/hashicorp/vault-helm/blob/96b8c98b3b7c6db1a7eb59acc85650fc6322cc3e/test/README.md#generating-the-values-json-schema) to help generate the new bits of schema, although it does require a bit of hand-editing to merge them in due to the fields that accept multiple types.
* A simple unit test for each new setting

Happy to help with either of those if you don't have the bandwidth.

values.yaml
RemcoBuddelmeijer: Good point! I will go ahead and duplicate the options and deprecate the old ones.

RemcoBuddelmeijer: I have gone ahead and edited the values.yaml, values.schema.json and the webhook deployment file. Next commit will contain the small unit test and the fix for the _helpers.tpl

RemcoBuddelmeijer:
Processed all the review comments and added unit tests!

RemcoBuddelmeijer:
Hey @tomhjp, 
Is there any chance that you can look at this in the near future? Thank you!

templates/_helpers.tpl
```diff
@@ -372,13 +372,13 @@ Sets extra injector service annotations
 Sets extra injector webhook annotations
 */}}
 {{- define "injector.webhookAnnotations" -}}
-  {{- if .Values.injector.webhookAnnotations }}
+  {{- if or (((.Values.injector.webhook)).annotations) (.Values.injector.webhookAnnotations)  }}
```
tomhjp, line 375: This double parenthesis pattern is new to me - it looks like it's behaving a bit like a null coalescing operator?

test/unit/injector-mutating-webhook.bats
```diff
@@ -53,103 +53,274 @@ load _helpers
...
 }
 
-@test "injector/MutatingWebhookConfiguration: can set namespaceSelector" {
+@test "injector/MutatingWebhookConfiguration: can set failurePolicy (deprecated)" {
   cd `chart_dir`
   local actual=$(helm template \
       --show-only templates/injector-mutating-webhook.yaml  \
       --set 'injector.enabled=true' \
-      --set 'injector.namespaceSelector.matchLabels.injector=true' \
+      --set 'injector.webhook=null' \
```
tomhjp, line 73: For users upgrading from older versions of the chart, `injector.webhook` typically won't be null, and they could get a behaviour change if they had set `injector.failurePolicy` to `Fail` but don't update anything under `injector.webhook`. Providing a non-null default value is the right thing, but I think we might just need to switch the precedence of the deprecated vs new field, i.e. change from:

```
failurePolicy: {{ ((.Values.injector.webhook)).failurePolicy | default .Values.injector.failurePolicy }}
```

to

```
failurePolicy: {{ .Values.injector.failurePolicy | default ((.Values.injector.webhook)).failurePolicy }}
```

And similar for the other values too. A good test would be that all the new tests you've written still pass without having to use `--set injector.webhook=null` at all.

tomhjp (commented):
Thanks for adding all the excellent tests! I just have one minor concern about the default upgrade path, but otherwise LGTM

templates/_helpers.tpl
RemcoBuddelmeijer: Yes, you are 100% correct! I have put this in place to support the new fields without them being required. Making use of the Go text processing functionality.

test/unit/injector-mutating-webhook.bats
RemcoBuddelmeijer: I actually didn't think this would happen. Thank you for letting me know.

My thought process was that: when upgrading, users would copy over their old values to the new fields. Follow the deprecation notes. And make use of the Helm chart as intended.
The other thought process I followed was that users would completely ignore the values.yaml update. Which followed into the new fields needing to be checked using the "`null coalescing operator`". Now if we change this behaviour, this operator might not be necessary anymore as it's not the preferred default.

Won't preferring the old fields over the new ones give any kind of confusion?

tomhjp: Yeah I think that seems fair. The disruption is minor and the long term direction is better this way round 👍 thanks for the discussion

tomhjp (approved):
Thanks for the contribution!

PR #700, opened 2022-03-02
change helm trigger branch from master to main

alvin-huang:
The default branch on `hashicorp/helm-charts` was renamed to `main` and this reflects that change.


tvoran (approved):
Thanks!

PR #703, opened 2022-03-11
VAULT-571 Matching documented behavior and consul

swenson:
Consul's helm template defaults most of the enabled to the special value
`"-"`, which means to inherit from global. This is what is implied
should happen in Vault as well according to the documentation for the
helm chart:

> [global.enabled] The master enabled/disabled configuration. If this is
> true, most components will be installed by default. If this is false,
> no components will be installed by default and manually opting-in is
> required, such as by setting server.enabled to true.

(https://www.vaultproject.io/docs/platform/k8s/helm/configuration#enabled)


test/docker/Test.dockerfile
```diff
@@ -22,15 +22,17 @@ RUN apk update && apk add --no-cache --virtual .build-deps \
...
+    gcompat
 
 # yq
 RUN pip install yq
 
 # gcloud
-RUN curl -OL https://dl.google.com/dl/cloudsdk/channels/rapid/install_google_cloud_sdk.bash && \
-    bash install_google_cloud_sdk.bash --disable-prompts --install-dir='/root/' && \
-    ln -s /root/google-cloud-sdk/bin/gcloud /usr/local/bin/gcloud
+#RUN curl -OL https://dl.google.com/dl/cloudsdk/channels/rapid/install_google_cloud_sdk.bash && \
```
swenson, line 32: Sorry, ignore this. I was trying to fix the Dockerfile and this slipped in. I'll remove it for the final PR.

templates/csi-clusterrole.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and (eq (.Values.csi.enabled | toString) "true" ) (eq (.Values.global.enabled | toString) "true") }}
+{{- if eq (include "vault.csiEnabled" . | toString) "true" -}}
```
swenson, line 1: Sadly the template `define` implicitly converts booleans to strings, so we can't just do `if (include "vault.csiEnabled" .)`

values.yaml
```diff
@@ -577,6 +579,7 @@ server:
   # the "file" backend.  This mode is not highly available and should not be scaled
   # past a single replica.
   standalone:
+    # @default: global.enabled
     enabled: "-"
```
tvoran, line 583: Should this default to server.enabled?

templates/_helpers.tpl
```diff
@@ -31,6 +31,55 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{/*
+Compute if the csi driver is enabled.
+*/}}
+{{- define "vault.csiEnabled" -}}
+{{- $_ := set . "csiEnabled" (or
+  (and (ne (.Values.csi.enabled | toString) "-") (eq (.Values.csi.enabled | toString) "true"))
```
tomhjp, line 39: Is the `(ne (.Values.csi.enabled | toString) "-")` part redundant? I'm reading it as `(A != B) && (A == C)`, which seems equivalent to `A == C` as a test, but I could be reading it wrong.

swenson: I think you're right. I'll simplify this.

values.yaml
```diff
@@ -493,7 +494,8 @@ server:
 
   # Enables a headless service to be used by the Vault Statefulset
   service:
-    enabled: true
+    # @default: global.enabled
+    enabled: "-"
```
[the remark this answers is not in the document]
swenson: I'm not sure. On the one hand, it doesn't make sense to have it default `true` if the server would be disabled with `global.enabled = false`. But on the other hand, it feels a little awkward to have `-` have multiple meanings (default to `global.enabled` in previous cases, but here default to `server.enabled` (which then defaults to `global.enabled`)).

I'm open to suggestion, for sure.

test/unit/schema.bats
```diff
@@ -1,46 +0,0 @@
-#!/usr/bin/env bats
```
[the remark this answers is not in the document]
swenson: I'll re-add this.

values.yaml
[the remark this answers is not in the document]
tvoran: In my mind, `global.enabled` is a toggle for each of the major components deployed by this chart, specifically affecting `injector.enabled`, `server.enabled`, and `csi.enabled`. Since this is essentially three charts in one, if `global.enabled` directly affects other options inside `server`, that feels like a leaky abstraction in my mind.

And likewise, `injector.enabled`, `server.enabled`, and `csi.enabled` are toggles for everything in their respective components. So to me it seems fine to be to have `server.service.enabled` default to `true`, since that would only be honored if the server component is enabled.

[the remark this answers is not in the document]
swenson: Agreed.

templates/_helpers.tpl
```diff
@@ -31,6 +31,55 @@ Expand the name of the chart.
...
+*/}}
+{{- define "vault.serverServiceEnabled" -}}
+{{- $_ := set . "serverServiceEnabled" (and
+      (or
+        (eq (.Values.server.enabled | toString) "true")
+        (and (eq (.Values.server.enabled | toString) "-") (eq (.Values.global.enabled | toString) "true")))
+      (or
+        (eq (.Values.server.service.enabled | toString) "true")
+        (and (eq (.Values.server.service.enabled | toString) "-") (eq (.Values.global.enabled | toString) "true")))) -}}
+{{- end -}}
```
tvoran, line 72: Can this leverage the serverEnabled template? Something along the lines of:

```suggestion
{{- define "vault.serverServiceEnabled" -}}
  {{- template "vault.serverEnabled" . -}}
  {{- $_ := set . "serverServiceEnabled" (and
      (.serverEnabled)
      (eq (.Values.server.service.enabled | toString) "true")) -}}
{{- end -}}
```

templates/ui-service.yaml
```diff
@@ -1,7 +1,10 @@
 {{ template "vault.mode" . }}
+# {{ .mode }}
```
tvoran, line 2: Still needed?

```suggestion
```

templates/ui-service.yaml
```diff
@@ -1,7 +1,10 @@
...
 {{- if ne .mode "external" }}
-{{- if and (ne .mode "") (eq (.Values.global.enabled | toString) "true") }}
-{{- if eq (.Values.ui.enabled | toString) "true" }}
+{{- template "vault.uiEnabled" . -}}
+# {{ .uiEnabled }}
```
tvoran, line 5: Still needed?

```suggestion
```

tvoran (commented):
This is looking great! Just a few more suggestions.

values.schema.json
```diff
@@ -747,7 +756,10 @@
                             ]
                         },
                         "enabled": {
-                            "type": "boolean"
+                            "type": [
+                                "boolean",
+                                "string"
+                            ]
```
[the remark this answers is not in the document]
swenson: Thanks for catching that!

values.yaml
```diff
@@ -577,6 +578,7 @@ server:
   # the "file" backend.  This mode is not highly available and should not be scaled
   # past a single replica.
   standalone:
+    # @default: global.enabled
```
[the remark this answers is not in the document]
swenson: :+1:

templates/ui-service.yaml
swenson: Thanks for catching these. I was using them in debugging :)

swenson:
Thanks @tvoran -- I made updates based on your comments.

templates/_helpers.tpl
```diff
@@ -31,6 +31,54 @@ Expand the name of the chart.
...
+{{/*
+Compute if the server service is enabled.
+*/}}
+{{- define "vault.serverServiceEnabled" -}}
+{{- template "vault.serverEnabled" . -}}
+{{- $_ := set . "serverServiceEnabled" (and
+      .serverEnabled
+      (or
+        (eq (.Values.server.service.enabled | toString) "true")
+        (and (eq (.Values.server.service.enabled | toString) "-") (eq (.Values.global.enabled | toString) "true")))) -}}
```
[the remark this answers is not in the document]
swenson: Ah, I think you're right. I'll simplify this.

CHANGELOG.md
```diff
@@ -1,12 +1,15 @@
...
 CHANGES:
+* `global.enabled` now works as documented, that is, setting `global.enabled` to false will disable everything, with individual components able to turned on individually
+* Default value of `-` used for csi, injector, server, and ui to indicate that they follow `global.enabled`.
 * Vault default image to 1.9.3
 * CSI provider default image to 1.0.0
 
 Improvements:
 * CSI: Set `extraLabels` for daemonset, pods, and service account [GH-690](https://github.com/hashicorp/vault-helm/pull/690)
 * Add namespace to injector-leader-elector role, rolebinding and secret [GH-683](https://github.com/hashicorp/vault-helm/pull/683)
+* Added helper functions for checking if csi, injector, server, server service, and ui are enabled in [tmeplates/_helpers.tpl](templates/_helpers.tpl) and made corresponding changes to use them
```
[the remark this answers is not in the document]
swenson: :+1:

tvoran (approved):
Thanks!

swenson:
Thanks everyone!

Issue #629, opened 2021-10-20
Raft cluster_addr env variable is hardcoded to use https


jmuleiro:
**Is your feature request related to a problem? Please describe.**
Feature. I've been trying to setup raft storage. After deleting one of the peers manually because of some debugging, I'm unable to re-add the node to the cluster. I've noticed that the environment variable `VAULT_CLUSTER_ADDR` is hardcoded to use https as seen in the `server-statefulset.yaml` template.

**Describe the solution you'd like**
Even though I've specified the cluster address and the leader API's address within the retry_join blocks, the pods still use a semi-hardcoded environment variable to try and reach the other nodes. I believe this may lead to not being able to re-join the cluster once a peer has been removed. If not using https, which is my case, we should be able to specify.

**Describe alternatives you've considered**
Tried hardcoding the environment variable a couple times but the way it is configured within the template does not allow for any changes when it comes to the https protocol.

**Additional context**
My **high availability** configuration looks like this, currently:
```
  ha:
    enabled: true
    replicas: 2
    raft:
      enabled: true
      setNodeId: true
      config: |
        ui = true
        api_addr = "http://POD_IP:8200"
        cluster_addr = "http://POD_IP:8201"
        disable_mlock = true
        
        listener "tcp" {
          tls_disable = 1
          address = "0.0.0.0:8200"
          cluster_address = "0.0.0.0:8201"
        }

        storage "raft" {
          path = "/vault/data"
          node_id = "vault-0"
          retry_join {
            leader_api_addr = "http://vault-0.vault-internal:8200"
          }

          retry_join {
            leader_api_addr = "http://vault-1.vault-internal:8200"
          }

          autopilot {
            cleanup_dead_servers = "true"
            last_contact_threshold = "200ms"
            last_contact_failure_threshold = "10m"
            max_trailing_logs = 250000
            min_quorum = 2
            server_stabilization_time = "10s"
          }
        }
        
        service_registration "kubernetes" {}

        seal "gcpckms" {
          project = "REDACTED"
          region = "global"
          key_ring = "REDACTED"
          crypto_key = "REDACTED"
        }
```
The `api_addr` variable is picked up and used to configure the pods' VAULT_API_ADDR variable, contrary to `cluster_addr` which is ignored.

sharkannon:
[...]

PR #709, opened 2022-03-26
Issue 629: updated to allow customization of the CLUSTER_ADDR the same…

sharkannon:
This is to hopefully fix Issue #629, allow the customization of VAULT_ADDR in various scenarios.  Copied the layout of VAULT_API_ADDR so that people can just copy the same style of formatting if required.


templates/server-statefulset.yaml
```diff
@@ -110,7 +110,11 @@ spec:
                 fieldRef:
                   fieldPath: metadata.name
             - name: VAULT_CLUSTER_ADDR
+              {{- if .Values.server.ha.clusterAddr }}
```
jmuleiro, line 113: This is correct, but maybe use helm's **default** instead of if/else ? since you're assigning `.Values.server.ha.clusterAddr` to `value` anyway, maybe try, instead of the if/else block:
```
# ...
- name: VAULT_CLUSTER_ADDR
  value: {{ .Values.server.ha.clusterAddr | default (printf "https://$(HOSTNAME).%s-internal:8201" (template "vault.fullname" .)) }}
# ...
```
Just a suggestion, let me know :)

[Documentation reference](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/)

values.yaml
```diff
@@ -680,6 +680,11 @@ server:
     # If set to null, this will be set to the Pod IP Address
     apiAddr: null
 
+    # Set the cluster_addr confuguration for Vault HA
+    # See https://www.vaultproject.io/docs/configuration#cluster_addr
+    # If set to nul, this will be set to https://$(HOSTNAME).{{ template "vault.fullname" . }}-internal:8201
```
jmuleiro, line 685:
```suggestion
    # If set to null, this will be set to https://$(HOSTNAME).{{ template "vault.fullname" . }}-internal:8201
```

jmuleiro (commented):
Looks great, just left a suggestion :)

templates/server-statefulset.yaml
sharkannon: Yea, makes sense, I was just trying to keep the syntax consistent with how others were doing things,

sharkannon: though, not sure why you have the`(template "vault.fullname" .)` in there.. typo?

sharkannon: oh, you're replacing with template fullname.. that isn't required as $(HOSTNAME) is an environment variable setting on the host itself.. it resolves correctly as a variable.

sharkannon: Forget it, I know what you're doing. :)

sharkannon: The default function won't parse the "template" function, so this won't work unfortunately.

jmuleiro: My bad, in that snippet I just followed the documentation blindly, as you stated it might not work then. Helm charts can be a pain.

sharkannon: All good, when you said it, I figured i'd try, didn't work though.. made me sad :(

MeijerM1 (approved):

sharkannon:
@tvoran I was wondering if you'd be able to review this PR.. you reviewed my other PR the other day and merged it.  I'm personally in need of this change, as I've had to use my forked version in our prod infrastructure.

calvn:
@sharkannon thanks for the contribution! Can you also add an entry to CHANGELOG.md around this?

sharkannon:
@calvn Should be good now, please let me know.

calvn (approved):
Changes look good, and unit tests passed locally on my end. Thanks for the contribution!

PR #712, opened 2022-04-04
Remove unncessary template calls

Iristyle:
- As part of VAULT-571 / #703 in 7109159, a new vault.serverEnabled
   template was added (and included in vault.mode)

   Various templates were updated accordingly, but those that were
   already calling vault.mode had an additonal call to
   vault.serverEnabled made which was unnecessary

   Remove those


tvoran (approved):
Well spotted, thanks!

PR #719, opened 2022-04-20
Updating chart to use 1.10.0 and 0.15.0 images

rgarcia6520:
Updating chart, values and other templates to use the following image versions:

vault: 1.10.0

vault-k8s: 0.15.0


tvoran:
Thanks for the patch here, though this has been superseded by the 0.20.1 release: https://github.com/hashicorp/vault-helm/blob/main/CHANGELOG.md#0201-may-25th-2022

@jseiser For MFA I'd recommend starting with Vault 1.10.3, since there were some bugs in that area with the earlier 1.10 releases: https://www.vaultproject.io/docs/upgrading/upgrade-to-1.10.x#login-mfa-not-enforced-after-restart

PR #722, opened 2022-04-27
Restore Secrets Store CSI driver to 1.0.0

swenson:
1.0.1+ seems to only support Kubernetes 1.19+, so we break support for k8s 1.16 if we upgrade


tomhjp (approved):
Good catch - I think we can probably roll forwards again soon, but we can take our time with that, making sure it's documented etc.

swenson:
Yep. I saw the acceptance tests fail, so wanted to make sure to fix them.

Thanks!

PR #736, opened 2022-05-23
Set default object selector for webhooks to exclude injector itself

swenson:
If `injector.failurePolicy` is set to `Fail`, there is a race condition
where if the mutating webhook config is setup before the injector, then
the injector can fail to start because it tries to inject itself.

We can work around this by ignoring the injector pod in in the webhook
by default.

Thanks to @joeyslalom for the object selector to exclude the pod.

Fixes https://github.com/hashicorp/vault-k8s/issues/258


tvoran (approved):

swenson:
Thanks!

PR #737, opened 2022-05-23
Restore missing 'vault' service account

swenson:
Our tutorials rely on this service account being present even if we are
using an external Vault.

The `values.yaml` also states that external Vaults are expected to use
this service account.

For example,
https://learn.hashicorp.com/tutorials/vault/kubernetes-external-vault?in=vault/kubernetes#install-the-vault-helm-chart-configured-to-address-an-external-vault

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -59,6 +59,32 @@ Compute if the server is enabled.
 {{- end -}}
 
 {{/*
+Compute if the server auth delegator serviceaccount is enabled.
+*/}}
+{{- define "vault.serverServiceAccountEnabled" -}}
+{{- $_ := set . "serverServiceAccountEnabled"
+  (and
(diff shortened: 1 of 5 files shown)


templates/_helpers.tpl
```diff
@@ -58,6 +58,32 @@ Compute if the server is enabled.
...
 
+{{/*
+Compute if the server auth delegator serviceaccount is enabled.
+*/}}
+{{- define "vault.serverServiceAccountEnabled" -}}
+{{- $_ := set . "serverServiceAccountEnabled"
+  (and
+    (eq (.Values.server.serviceAccount.create | toString) "true" )
+    (or
+      (eq (.Values.server.enabled | toString) "true")
```
swenson, line 69: Trying to figure out how `server.enabled`, `global.enabled`, `serviceAccount.create`, and `authDelegator.enabled` all **should** interact is a good way to get a headache.

This is what I came up with.

PR #744, opened 2022-06-01
Start testing against Kubernetes 1.24

swenson:


.github/workflows/acceptance.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kind-k8s-version: [1.16.15, 1.20.15, 1.21.10, 1.22.7, 1.23.4]
+        kind-k8s-version: [1.16.15, 1.20.15, 1.21.10, 1.22.7, 1.23.4, 1.24.1]
```
tvoran, line 14: Might as well update the other patch versions too:
```suggestion
        kind-k8s-version: [1.16.15, 1.20.15, 1.21.12, 1.22.9, 1.23.6, 1.24.1]
```

.github/workflows/acceptance.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
         with:
           config: test/kind/config.yaml
           node_image: kindest/node:v${{ matrix.kind-k8s-version }}
+          version: v0.14.0
 
       # Skip CSI tests if K8s version < 1.16.x
```
tvoran, line 28: If you want, we could also remove the < 1.16.x check step here (and the `check_skip_csi` function and calls in test/acceptance/csi.bats.

tvoran (approved):

.github/workflows/acceptance.yaml
swenson: :+1:

swenson:
Thanks!

PR #745, opened 2022-06-06
Set VAULT_ADDR env var for CSI Provider pods

tomhjp:
Makes the setting of Vault address for CSI more consistent with the injector.

This won't have any effect until we update the CSI Provider `-vault-addr` flag to have an empty default, as flags take precedence over environment variables: https://github.com/hashicorp/vault-csi-provider/pull/160. Note this also means users can continue to set the `-vault-addr` flag via `csi.extraArgs` without being affected by this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,13 @@
 ## Unreleased
 
 CHANGES:
-* Start testing against Kubernetes 1.24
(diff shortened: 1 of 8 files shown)


swenson (approved):
LGTM

tvoran (approved):

PR #749, opened 2022-06-21
Add Vault Agent sidecar to CSI Provider

tomhjp:
Adds Agent as a sidecar for the CSI Provider to:

* Cache k8s auth login leases
* Cache secret leases
* Automatically renew renewable leases in the background

It depends on https://github.com/hashicorp/vault-csi-provider/pull/163 to work properly, because otherwise the initial Kubernetes JWT used to log in will be different every time, so we'll never get cache hits.

### DIFF
--- a/test/unit/csi-agent-configmap.bats
+++ b/test/unit/csi-agent-configmap.bats
@@ -11,16 +11,6 @@ load _helpers
   [ "${actual}" = "false" ]
 }
 
-@test "csi/Agent-ConfigMap: enabled with csi.enabled" {
-  cd `chart_dir`
-  local actual=$(helm template \
-      --show-only templates/csi-agent-configmap.yaml  \
-      --set 'csi.enabled=true' \
-      . | tee /dev/stderr |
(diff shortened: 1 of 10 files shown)


test/unit/csi-agent-configmap.bats
```diff
@@ -0,0 +1,55 @@
...
+
+@test "csi/Agent-ConfigMap: name" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/csi-agent-configmap.yaml \
+      --set "csi.enabled=true" \
+      . | tee /dev/stderr |
+      yq -r '.metadata.name' | tee /dev/stderr)
+  [ "${actual}" = "release-name-vault-csi-provider-agent-config" ]
+}
```
tvoran, line 32: Seems like these two could be combined if we wanted.

tomhjp: Good point, the second is just a superset of the first really.

tvoran (approved):

Issue #663, opened 2021-12-16
Drop unneeded linux capabilities


OOTS:
**Is your feature request related to a problem? Please describe.**
In our company, we want to enforce a policy where all applications running in K8s ought to drop as many Linux capabilities (in a container's security context) as possible. I would like to do that for Vault, but the [server-statefulset.yaml file](https://github.com/hashicorp/vault-helm/blob/main/templates/server-statefulset.yaml#L73) does not offer the option to do so.

**Describe the solution you'd like**
Ideally, the following piece of configuration in the server-statefulset.yaml file:
```
{{- if not .Values.global.openshift }}
securityContext:
  allowPrivilegeEscalation: false
{{- end }}
```
should be changed to:
```
securityContext:
  capabilities:
    drop: ["ALL"]
    {{- if not .Values.global.openshift }}
    allowPrivilegeEscalation: false
    {{- end }}
```
This would comply with the principle of least privilege, and harden Vault a little bit more.

I don't fully understand the situation with mlock, but from what I get, mlock is never used on K8s. If I'm mistaken, one should probably retain the `IPC_LOCK` capability.

**Describe alternatives you've considered**
Alternatively, the block shown above could be changed to:
```
{{- if or .Values.server.securityContext (not .Values.global.openshift) }}
securityContext:
  {{- if not .Values.global.openshift }}
  allowPrivilegeEscalation: false
  {{- end }}
  {{- if .Values.server.securityContext }}
    {{ tpl .Values.server.securityContext . | nindent $whatever }}
  {{- end }}
{{- end }}
```
This would allow more flexibility.

**Additional context**
References:
* [K8s Documentation for Security Contexts/Capabilities](https://kubernetes.io/docs/tasks/configure-pod-container/security-context)/#set-capabilities-for-a-container

I can send a PR if this change would be welcome.

langecode:
Also could be nice if the security context structure was part of the values not "hard coded" inside of the templates (as also shown in the alternative here). This way it would be possible to adjust when deploying the chart - currently we have a policy in place requiring a drop of all privileges (as recommended in the restricted pod security standard: https://kubernetes.io/docs/concepts/security/pod-security-standards/

Not being able to adjust this forces us to do an exemption for this workload.

OOTS:
By the way, the analogous thing should be done with:
- the Vault Agent Injector (the pods implementing the `MutatingWebhook`)
- the Init Containers and sidecars injected by the Vault Agent Injector
[...]

PR #750, opened 2022-06-24
Feat/adding container security context

ChrisFraun:
It should partially related to this issue: https://github.com/hashicorp/vault-helm/issues/663 

What: Deployments can have security settings in their manifest on two levels: pod and container. However, there are some capabilities only configurable in one of the respective levels(https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#securitycontext-v1-core). This PR sets a default configuration for container securityContext, which drops all POSIX capabilities, and for podSecurityContext, which adds fsGroup and supplementalGroups. These are and should be standard settings in the context of Kubernetes. It also adds the possibility of running vault-injector in a Kubernetes environment without PSP (to be removed in v1.25 https://kubernetes.io/docs/concepts/security/pod-security-policy/), but with OpenPolicyAgent (possibly the PSP substitute) with the same capabilities as a **restricted** PSP instead.

Why: This missing configuration restricts vault-injector from being used with the simple upstream helm chart without modifications/unnecessary maintenance. This especially applies to the restricted policy use in OPA. 

This PR exports the respective settings to the values.yaml and is defaulting them there since the securityContext variables are hardcoded so far in the deployment, as well.
The other possibility would be to not use default values for the podSecurityContext and (container) securityContext variables in the values.yaml and let them e.g. empty to let the user decide themselves. This I would not recommend.
Benefit would be a more secure deployment in the future.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -474,16 +474,14 @@ Sets extra injector service annotations
 securityContext for the injector pod level.
 */}}
 {{- define "injector.securityContext.pod" -}}
-  {{- if not .Values.global.openshift }}
-    {{- if or (.Values.injector.uid) (.Values.injector.gid) }}
+  {{- if or (.Values.injector.uid) (.Values.injector.gid) }}
       securityContext:
         runAsNonRoot: true
         runAsGroup: {{ .Values.injector.gid | default 1000 }}
         runAsUser: {{ .Values.injector.uid | default 100 }}
-    {{- else }}
+  {{- else }}
       securityContext:
         {{- toYaml .Values.injector.securityContext.pod | nindent 8 }}
-    {{- end }}
   {{- end }}
 {{- end -}}
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -479,7 +479,7 @@ securityContext for the injector pod level.
         runAsNonRoot: true
         runAsGroup: {{ .Values.injector.gid | default 1000 }}
         runAsUser: {{ .Values.injector.uid | default 100 }}
-  {{- else }}
+  {{- else if .Values.injector.securityContext.pod }}
       securityContext:
         {{- toYaml .Values.injector.securityContext.pod | nindent 8 }}
   {{- end }}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -471,6 +471,35 @@ Sets extra injector service annotations
 {{- end -}}
 
 {{/*
+securityContext for the injector pod level.
+*/}}
+{{- define "injector.securityContext.pod" -}}
+{{- if or (.Values.injector.uid) (.Values.injector.gid) }}
+{{- if not .Values.global.openshift }}
+      securityContext:
+        runAsNonRoot: true
+        runAsGroup: {{ .Values.injector.gid | default 1000 }}
+        runAsUser: {{ .Values.injector.uid | default 100 }}
+{{- else }}
+      securityContext:
+        {{- toYaml .Values.injector.securityContext.pod | nindent 8 }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
+{{/*
+securityContext for the injector container level.
+*/}}
+{{- define "injector.securityContext.container" -}}
+{{- if .Values.injector.securityContext.container}}
+{{- if not .Values.global.openshift }}
+          securityContext:
+            {{- toYaml .Values.injector.securityContext.container | nindent 10 }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
+{{/*
 Sets extra injector webhook annotations
 */}}
 {{- define "injector.webhookAnnotations" -}}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -489,11 +489,9 @@ securityContext for the injector pod level.
 securityContext for the injector container level.
 */}}
 {{- define "injector.securityContext.container" -}}
-  {{- if not .Values.global.openshift }}
-    {{- if .Values.injector.securityContext.container}}
+  {{- if .Values.injector.securityContext.container}}
           securityContext:
             {{- toYaml .Values.injector.securityContext.container | nindent 12 }}
-    {{- end }}
   {{- end }}
 {{- end -}} 
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -494,7 +494,7 @@ securityContext for the injector container level.
 {{- if .Values.injector.securityContext.container}}
 {{- if not .Values.global.openshift }}
           securityContext:
-            {{- toYaml .Values.injector.securityContext.container | nindent 10 }}
+            {{- toYaml .Values.injector.securityContext.container | nindent 12 }}
 {{- end }}
 {{- end }}
 {{- end -}}
--- a/templates/injector-deployment.yaml
+++ b/templates/injector-deployment.yaml
@@ -40,13 +40,7 @@ spec:
       serviceAccountName: "{{ template "vault.fullname" . }}-agent-injector"
       {{- if not .Values.global.openshift }}
       hostNetwork: {{ .Values.injector.hostNetwork }}
-      securityContext:
-        runAsNonRoot: true
-        runAsGroup: {{ .Values.injector.gid | default 1000 }}
-        runAsUser: {{ .Values.injector.uid | default 100 }}
-        fsGroup: {{ .Values.injector.uid | default 100 }}
-        supplementalGroups:
-          - {{ .Values.injector.uid | default 100 }}
+      securityContext: {{ .Values.podSecurityContext }}
       {{- end }}
       containers:
         - name: sidecar-injector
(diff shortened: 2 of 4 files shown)


ChrisFraun:
any progress/impression on that?

tvoran (commented):
Hi @ChrisFraun, thanks for the work here! I left some suggestions that I think will help maintain backward compatibility with the old `injector.uid` and `injector.gid` options, while allowing the specification of the two securityContext's as complete objects (similar to an earlier version of this PR).

ChrisFraun:
Hi @tvoran,
tried to follow up on your suggestions.
Can you check it, please?

tvoran (commented):
Thanks for the updates! I left some more suggestions, and would you mind adding tests for these options in test/unit/injector-deployment.bats?

values.yaml
```diff
@@ -202,6 +202,22 @@ injector:
...
 
+  # Default pod and container security context for vault-injector
+  securityContext:
+    pod:
+      runAsNonRoot: true
+      runAsGroup: 1000
+      runAsUser: 100
+      fsGroup: 100
+      supplementalGroups:
+        - 100
```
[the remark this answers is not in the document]
ChrisFraun: I think so, yes

templates/injector-deployment.yaml
```diff
@@ -40,20 +42,12 @@ spec:
...
-      securityContext:
-        runAsNonRoot: true
-        runAsGroup: {{ .Values.injector.gid | default 1000 }}
-        runAsUser: {{ .Values.injector.uid | default 100 }}
       {{- end }}
       containers:
         - name: sidecar-injector
           {{ template "injector.resources" . }}
           image: "{{ .Values.injector.image.repository }}:{{ .Values.injector.image.tag }}"
           imagePullPolicy: "{{ .Values.injector.image.pullPolicy }}"
```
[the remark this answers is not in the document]
ChrisFraun: Makes sense to have it not inside the template, otherwise people would miss it maybe...

ChrisFraun:
Hi @tvoran, thanks a lot for your input! :) 
I merged your changes and modified the test/unit/injector-deployment.bats.
Can you check again?

templates/_helpers.tpl
```diff
@@ -470,6 +470,31 @@ Sets extra injector service annotations
...
+{{/*
+securityContext for the injector pod level.
+*/}}
+{{- define "injector.securityContext.pod" -}}
+  {{- if or (.Values.injector.uid) (.Values.injector.gid) }}
+      securityContext:
+        runAsNonRoot: true
+        runAsGroup: {{ .Values.injector.gid | default 1000 }}
+        runAsUser: {{ .Values.injector.uid | default 100 }}
+  {{- else }}
```
tvoran, line 482: Might as well check if this securityContext is set, similar to `securityContext.container`.

```suggestion
  {{- else if .Values.injector.securityContext.pod }}
```

test/unit/injector-deployment.bats
```diff
@@ -363,6 +363,87 @@ load _helpers
...
+@test "injector/deployment: default pod securityContext" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/injector-deployment.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.securityContext' | tee /dev/stderr)
+  [ "${actual}" = "null" ]
+}
+
+@test "injector/deployment: custom pod securityContext" {
```
tvoran, line 378: Can you change these tests to check something besides the default values? Doesn't need to be too exhaustive, just overriding a couple of the options would suffice.

test/unit/injector-deployment.bats
```diff
@@ -363,6 +363,87 @@ load _helpers
...
+@test "injector/deployment: default container securityContext sidecar-injector" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/injector-deployment.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].securityContext' | tee /dev/stderr)
+  [ "${actual}" = "null" ]
+}
+
+@test "injector/deployment: custom container securityContext sidecar-injector" {
```
tvoran, line 428: This should be testing values that aren't already the default.

test/unit/injector-deployment.bats
```diff
@@ -363,6 +363,87 @@ load _helpers
...
+  [ "${actual}" = "1000" ]
+}
+
+@test "injector/deployment: default container securityContext sidecar-injector" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/injector-deployment.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].securityContext' | tee /dev/stderr)
+  [ "${actual}" = "null" ]
```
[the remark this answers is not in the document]
ChrisFraun: added then`!`

test/unit/injector-deployment.bats
ChrisFraun: added not default values

test/unit/injector-deployment.bats
ChrisFraun: added some other values

test/unit/injector-deployment.bats
```diff
@@ -363,6 +363,87 @@ load _helpers
...
+
+@test "injector/deployment: default pod securityContext" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/injector-deployment.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.securityContext' | tee /dev/stderr)
+  [ "${actual}" = "null" ]
+}
+
```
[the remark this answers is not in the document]
ChrisFraun: `@test "injector/deployment: backward pod securityContext" {
  cd `chart_dir`
  local object=$(helm template \
      --show-only templates/injector-deployment.yaml  \
      --set 'injector.uid=200' \
      --set 'injector.gid=4000' \
      . | tee /dev/stderr |
      yq -r '.spec.template.spec.securityContext' | tee /dev/stderr)

  local value=$(echo $actual | yq -r .runAsUser | tee /dev/stderr)
  [ "${value}" = "200" ]

  local value=$(echo $actual | yq -r .runAsGroup | tee /dev/stderr)
  [ "${value}" = "4000" ]
}`

added this snippet

ChrisFraun:
Hi @tvoran, thank you again for the quick response and the input.
I tried to tend to you comments.
Please feel free to check again!

test/unit/injector-deployment.bats
```diff
@@ -363,6 +363,121 @@ load _helpers
...
+  [ "${value}" = "1000" ]
+
+  local value=$(echo $actual | yq -r .runAsNonRoot | tee /dev/stderr)
+  [ "${value}" = "true" ]
+
+  local value=$(echo $actual | yq -r .runAsUser | tee /dev/stderr)
+  [ "${value}" = "100" ]
+}
+
+@test "injector/deployment: custom pod securityContext" {
```
tvoran, line 407: Looks like this test will fail because the checks are for the wrong values, i.e. `[ "${actual}" = "1000" ]` when `runAsGroup` was set to 1001.

test/unit/injector-deployment.bats
```diff
@@ -363,6 +363,121 @@ load _helpers
...
+  local value=$(echo $actual | yq -r .capabilities.drop | tee /dev/stderr)
+  [ "${value}" = "ALL" ]
+}
+
+@test "injector/deployment: custom container securityContext sidecar-injector" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/injector-deployment.yaml  \
+      --set 'injector.enabled=true' \
+      --set 'injector.securityContext.container.privileged=false' \
```
tvoran, line 468: Looks like this is missing the chart location argument:

```suggestion
      --set 'injector.securityContext.container.privileged=false' \
      . | tee /dev/stderr |
```

tvoran (commented):
Thanks, though the tests aren't passing for me. Note that this repo doesn't run the unit tests on PRs from forks. Are you able to run the tests yourself? There are some instructions here: https://github.com/hashicorp/vault-helm/blob/main/CONTRIBUTING.md#testing

ChrisFraun:
Hi @tvoran, sorry for the sloppyness, I fixed the unit test and get `540 tests, 0 failures`

the only missing thing is one test on the L 459, which I could not get to work. Maybe because of escaping characters?
`
  local value=$(echo $actual | yq -r .capabilities.drop | tee /dev/stderr)
  [ "${value}" = "[" ALL "]"]
 `

tvoran:
No worries! I think for that test from L 459, since the value to check is the first element in a list, it can be referenced like this:

```bash
local value=$(echo $actual | yq -r .capabilities.drop[0] | tee /dev/stderr)
```

ChrisFraun:
`540 tests, 0 failures`
I think we are good to go. Many thanks for the advice @tvoran.

tvoran (approved):

peteroruba:
Really cool this got merged

PR #752, opened 2022-07-06
 Added support to configure default vault namespace on the agent config 

bradfordwagner:
follow up to: https://github.com/hashicorp/vault-k8s/pull/345


templates/injector-deployment.yaml
```diff
@@ -118,6 +118,8 @@ spec:
             - name: AGENT_INJECT_TEMPLATE_STATIC_SECRET_RENDER_INTERVAL
               value: "{{ .Values.injector.agentDefaults.templateConfig.staticSecretRenderInterval }}"
             {{- end }}
+            - name: AGENT_INJECT_VAULT_NAMESPACE
+              value: "{{ .Values.injector.agentDefaults.namespace }}"
```
tvoran, line 122: I think this var should only be set if `.Values.injector.agentDefaults.namespace` isn't empty, similar to how `AGENT_INJECT_TEMPLATE_STATIC_SECRET_RENDER_INTERVAL` is handled above.

tvoran (commented):
This looks like it's on the right track, though I'd like to also see tests in test/unit/injector-deployment.bats for the new option.

It also occurs to me that this PR may not strictly be necessary since the env var can already be set via the existing [injector.extraEnvironmentVars](https://www.vaultproject.io/docs/platform/k8s/helm/configuration#extraenvironmentvars) option. For example:

```
# values.yaml
injector:
  extraEnvironmentVars:
    AGENT_INJECT_VAULT_NAMESPACE: "admin"
```

Or on the cli:
```
helm install vault hashicorp/vault \
  --set injector.extraEnvironmentVars.AGENT_INJECT_VAULT_NAMESPACE="admin" 
```

PR #759, opened 2022-07-19
feat: add fsGroupChangePolicy

johanneswuerbach:
Kubernetes is chown-ing the entire data volume by default, which can be rather slow.

Make this configurable which is possible since v1.23 (GA) and v1.20 (beta) https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods


swenson:
I like the idea of this. We recently merged https://github.com/hashicorp/vault-helm/pull/767, which allows you to specify any additional `securityContext` you like, which could include `fsGroupChangePolicy`, so I think this PR is no longer necessary to specify that?

johanneswuerbach:
Yes, #767 also fixes this 🎉

PR #763, opened 2022-08-02
Allow loadBalancerClass to be specified for the main vault service

bbetter173:
This allows the `loadBalancerClass` field of the main Vault service to be configured.

### DIFF
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -22,6 +22,9 @@ spec:
   {{- if .Values.server.service.clusterIP }}
   clusterIP: {{ .Values.server.service.clusterIP }}
   {{- end }}
+  {{- if (and (eq .Values.server.service.type "LoadBalancer") .Values.server.service.loadBalancerClass) }}
+  loadBalancerClass: {{ .Values.server.service.loadBalancerClass }}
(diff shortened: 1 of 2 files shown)


timclifford:
```
  service:
    enabled: true

    type: LoadBalancer

    # loadBalancerClass: service.k8s.aws/nlb

    externalTrafficPolicy: Local
    port: 8200
    targetPort: 8200
    
    annotations:  
      service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
      service.beta.kubernetes.io/aws-load-balancer-type: external
      service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "8200"
```
When defining the above using v1.27 k8s and aws-load-balancer-controller v2.5.2 which seems to inject the loadBalancerClass field in even though its not defined in the vault service template.

image

There is an issue then when you try to rollout an update.
```
Service "vault" is invalid: spec.loadBalancerClass: Invalid value: "null": may not change once set
```
In this case you need to tear down your vault services in order to do a release which isn't ideal.
 
This PR will resolve that since you can define loadBalancerClass to match `service.k8s.aws/nlb`

Issue #729, opened 2022-05-17
Cant' deploy after latest (0.20.0) release


kevinlmadison:
**Describe the bug**
Can't install vault with helm after new release.
```
$ helm repo add hashicorp https://helm.releases.hashicorp.com
"hashicorp" already exists with the same configuration, skipping

$ helm install vault hashicorp/vault -n admin                
Error: parse error at (vault/templates/_helpers.tpl:38): unclosed action
```

**To Reproduce**
See above

**Expected behavior**
Vault should install.
**Environment**
* Kubernetes version: 
RKE2 - k8s 1.22.6

tvoran:
Hi @kevinlmadison, I'm not able to duplicate the error you're getting. Could you tell me what version of helm you're using? Also don't forget to run `helm repo update` to pick up the latest vault-helm release.

kevinlmadison:
@tvoran thank you for your reply!
```
$ helm version
version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"clean", GoVersion:"go1.14.9"}
```
after running `helm repo update` the error is still the same : (

tvoran:
Ah, I bet it's a helm version issue. Our acceptance tests here are running in github's `ubuntu-latest` environment, which has helm 3.8.2: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#package-management

Can you try updating your version of helm?

tvoran:
From some quick testing, it looks like the min version to get past the error is helm 3.6.

kevinlmadison:
righto, I'll test that right now!

kevinlmadison:
@tvoran That solved it, thank you for your help!!

PR #769, opened 2022-08-09
Update README.md

jpenghashi:
The helm requirements should be 3.8+ based on https://github.com/hashicorp/vault-helm/issues/729#issuecomment-1129183042


swenson:
According to that comment thread, 3.6 should be sufficient though?

PR #215, opened 2020-02-27
Add Prometheus Operator support

cablespaghetti:
This adds support for Prometheus Operator resources within the Helm chart:
- **ServiceMonitor** - To make the metrics endpoint automatically discoverable
- **PrometheusRules** - To configure Rules (Alerts) for Vault within the chart

I've also added some examples to the values.yaml for how to get Prometheus metrics working in unauthenticated mode, as it's not obvious and I expect it's a common use case.

I'll write some tests over the next few days but this works in my environments.

PR #772, opened 2022-08-10
Add support for the Prometheus Operator

benashz:
This PR is a continuation of #215, which added support for deploying `ServiceMonitor` and `PrometheusRules` custom resources. The feature requires that the [prometheus-operator](https://github.com/prometheus-operator/prometheus-operator) has been deployed to the same Kubernetes cluster as Vault.


values.yaml
```diff
@@ -1008,3 +1030,90 @@ csi:
...
+      enabled: false
+
+      # Default selector labels to add to the PrometheusRule.
+      # Assumes that the operator has been installed by Helm with a release name of 'prometheus'
+      defaultSelectors:
+         release: prometheus
+         app: kube-prometheus-stack
+
+      # Default selector overrides.
+      overrideSelectors: {}
```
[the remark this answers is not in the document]
benashz: Yeah, I like that idea. We can just document what the defaults are here.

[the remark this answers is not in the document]
benashz: That means there would be no way to remove all of the selectors, I am not sure that is valid use case however.

templates/prometheus-servicemonitor.yaml
```diff
@@ -0,0 +1,37 @@
...
+    matchLabels:
+      app.kubernetes.io/name: {{ template "vault.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+      {{- if eq .mode "ha" }}
+      vault-active: "true"
+      {{- else }}
+      vault-internal: "true"
+      {{- end }}
+  endpoints:
+  - port: http
```
tvoran, line 24: Does this need to match the name of the vault service port? Should it be `{{ include "vault.scheme" . }}` instead? I think it's the same for active and internal:

https://github.com/hashicorp/vault-helm/blob/17bffbcabe0221be0b8e0a02f5a9d8b2b76794f0/templates/server-ha-active-service.yaml#L29-L35

https://github.com/hashicorp/vault-helm/blob/17bffbcabe0221be0b8e0a02f5a9d8b2b76794f0/templates/server-headless-service.yaml#L22-L25

benashz: Good call! Fixed with be4019d937d74d8d2b1be57b23122ecc1aed9614

test/acceptance/server-ha-enterprise-dr.bats
```diff
@@ -25,6 +25,7 @@ load _helpers
   local init=$(kubectl exec -ti "$(name_prefix)-east-0" -- \
     vault operator init -format=json -n 1 -t 1)
 
+  echo $init | jq >&2
```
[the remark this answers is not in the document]
benashz: Removed in bc746feb29e3a2b9ebe9ff9f9d9f9a907019ff0e

templates/prometheus-prometheusrules.yaml
```diff
@@ -0,0 +1,26 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{- $selectors := .Values.telemetry.prometheusOperator.prometheusRules.selectors }}
+    {{- if $selectors }}
+    {{- toYaml $selectors | nindent 4 }}
+    {{- else }}
+    {{/*
+    update the selectors docs in values.yaml whenever these defaults change.
+    */}}
+    app: kube-prometheus-stack
```
tvoran, line 18: It looks like this default label was removed from the ServiceMonitor, so I wonder if it needs to be here too? It seems like the app is defined by the `app.kubernetes.io/name` label above, and from what I can tell the default prometheus install only selects on `release: prometheus`.

```suggestion
```

values.yaml
```diff
@@ -1008,3 +1030,82 @@ csi:
...
+      enabled: false
+
+      # Selector labels to add to the ServiceMonitor.
+      # When empty, defaults to:
+      #  release: prometheus
+      #  app: kube-prometheus-stack
+      selectors: {}
+
+      # Some example rules.
+      rules: {}
```
tvoran, line 1097: I wonder if this should somehow be further divided into a server telemetry section? Just thinking that what's being templated here is specific to Vault server, and if we ever wanted to add a serviceMonitor for the vault-k8s injector for example, we'd need another set of these options.

values.yaml
```diff
@@ -1008,3 +1030,82 @@ csi:
...
+      # Selector labels to add to the ServiceMonitor.
+      # When empty, defaults to:
+      #  release: prometheus
+      selectors: {}
+
+      # Interval at which Prometheus scrapes metrics
+      interval: 30s
+
+      # Timeout at which Prometheus timeouts scrape run
+      scrapeTimeout: 10s
```
tvoran, line 1083:
```suggestion
      # Timeout for Prometheus scrapes
      scrapeTimeout: 10s
```

templates/prometheus-prometheusrules.yaml
benashz: Fixed in 4ee36553b854d76b6b2f69980d9d0e643e2dd9e8

benashz: Ah, yes this was a part of the original PR. I wonder if it might make sense to make the `release` value public. We we can set the default in the template here.

values.yaml
benashz: Good call. What do you think of the following structure?
```
  prometheusOperator:
    serviceMonitor:
      server:
        # Enable deployment of the Vault Server ServiceMonitor CustomResource.
        enabled: false

        # Selector labels to add to the ServiceMonitor.
        # When empty, defaults to:
        #  release: prometheus
        selectors: {}

        # Interval at which Prometheus scrapes metrics
        interval: 30s

        # Timeout for Prometheus scrapes
        scrapeTimeout: 10s

    prometheusRules:
      server:
        # Deploy the PrometheusRule custom resource for AlertManager based alerts.
        # Requires that AlertManager is properly deployed.
        enabled: false

[...]
```

We could then extend it to support other `ServiceMonitor` deployments etc.

benashz: Implemented in d6b9c532b7423921f002fc22a1a6b5ab02071268

tvoran: Yeah that structure should work. We could probably remove either `telemetry` or `prometheusOperator` as well to trim the hierarchy a bit.

templates/prometheus-prometheusrules.yaml
tvoran: > I wonder if it might make sense to make the `release` value public

Leaving it with the default like you have it seems fine.

tvoran (approved):

tvoran:
Could also add the changelog entry in this PR as well.

values.yaml
benashz: I like the idea of dropping the latter.

benashz: Fixed in 4886b1469102e0ce906f2f0ce015872f00b1676c

PR #775, opened 2022-08-17
Add server.hostNetwork option

### LINKED ISSUE #643: Support hostNetwork on the server StatefulSet
We'd like to reuse K8s API etcd as a backend for the vault instance running inside the cluster. This is CAPI cluster where etcd pod runs in the host network bound to localhost. This etcd pod doesn't expose any ports. There isn't really way to expose this port cleanly because it's created with upstream controllers. We'd like to run vault server in the host network to make it possible to reach the etcd.

Add `server.hostNetwork` setting.

We were thinking about exposing etcd with a proxy but that may make it complicated with the certificates.

**Additional context**
N/A. I'm happy to provide a PR if this is something you'd accept.

tomhjp:
Fixes #643

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -49,6 +49,10 @@ spec:
       shareProcessNamespace: true
       {{ end }}
       {{- template "server.statefulSet.securityContext.pod" . }}
+      {{- if not .Values.global.openshift }}
+      hostNetwork: {{ .Values.server.hostNetwork }}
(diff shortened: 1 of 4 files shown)


tvoran (approved):

koktlzz:
Why this option is disable for OpenShift?

```
      {{- if not .Values.global.openshift }}
       hostNetwork: {{ .Values.server.hostNetwork }}
       {{- end }}
```

Issue #792, opened 2022-09-28
networking.k8s.io/v1 - pathType set in wrong place


obounaim:
**Describe the bug**

pathType field added in issue [663](https://github.com/hashicorp/vault-helm/issues/633) seems to be in the wrong place. According to ingress specification in networking.k8s.io/v1, pathType field should be set in ingresses.spec.rules.http.paths. Currently it is set in ingresses.spec.rules.http.paths.path.

**To Reproduce**
1. Enable ingress and install chart
2. See error from `terraform apply` or `helm install`
 
│ Error: error running dry run for a diff: error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0]): missing required field "pathType" in io.k8s.api.networking.v1.HTTPIngressPath
│ 

**Expected behavior**
When ingress is enabled, chart installation should not fail

**Environment**
* Kubernetes version: 1.22
  * EKS
* vault-helm version: 0.20.1 (latest version also concerned)
* terraform version: 1.2.9 (Should not make a difference)
Chart values:

```yaml
  ...
  ingress:
    enabled: true 
    annotations: {
      kubernetes.io/ingress.class: nginx
    }
    hosts:
      - host: ${host}

    extraPaths:
      - path: /*
        backend:
          service:
            name: vault 
            port:
              number: 8200 

    tls: []
```

obounaim:
#794

tvoran:
Hi @obounaim, I think that error is because there needs to be a `pathType` in each element of `extraPaths`. Adding that lets the chart install correctly for me:

```yaml
  ...
  ingress:
    enabled: true 
    annotations: {
      kubernetes.io/ingress.class: nginx
    }
    hosts:
      - host: ${host}

    extraPaths:
      - path: /*
        pathType: Prefix
        backend:
          service:
            name: vault 
            port:
              number: 8200 

    tls: []
```

The chart output looks like it matches the spec to me, since `backend`, `path`, and `pathType` are all part of the [HTTPIngressPath object](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#httpingresspath-v1-networking-k8s-io).

Also note that by default the extraPaths may not be necessary, since the chart will setup an ingress rule for `/` like this:

```yaml
    rules:
    - host: <host>
      http:
        paths:
        - backend:
            service:
              name: vault
              port:
                number: 8200
          path: /
          pathType: Prefix
```

PR #794, opened 2022-09-28
Path type issue 792 pr

obounaim:
#792


tvoran:
Thanks for the patch, though I don't think this is necessary. See my comment on #792.

Issue #797, opened 2022-10-12
Make ephemeral storage of Agent sidecar injector configurable.


krishnadas-m:
**Is your feature request related to a problem? Please describe.**
It is observed that the vault agent sidecar gets evicted due to the limit of ephemeral storage. It is currently defaulted to 128Mi and is not configurable via Helm-chart.

**Describe the solution you'd like**
If we can modify the injector-deployment template to template the AGENT_INJECT_EPHEMERAL_LIMIT value from values.yaml

**Describe alternatives you've considered**
We need to add/modify annotation on components to increase the limit which can be easily solved if we add the above mentioned capability on the vault helm-chart.

tvoran:
Hi @krishnadas-m, this can be set via the helm chart currently with the [injector.extraEnvironmentVars](https://developer.hashicorp.com/vault/docs/platform/k8s/helm/configuration#extraenvironmentvars) option.

The next release of vault-k8s will also [stop setting a default ephemeral storage limit](https://github.com/hashicorp/vault-k8s/pull/386).

krishnadas-m:
will pass additional env variables.

PR #798, opened 2022-10-12
made ephemeral storage configurable(#797)

krishnadas-m:


templates/injector-deployment.yaml
```diff
@@ -104,6 +104,10 @@ spec:
               value: "{{ .Values.injector.agentDefaults.memRequest }}"
             - name: AGENT_INJECT_MEM_LIMIT
               value: "{{ .Values.injector.agentDefaults.memLimit }}"
+            - name: AGENT_INJECT_EPHEMERAL_REQUEST
+              value: "{{ .Values.injector.agentDefaults.ephemeralRequest }}"
+            - name: AGENT_INJECT_EPHEMERAL_LIMIT
+              value: "{{ .Values.injector.agentDefaults.ephemeralLimit }}"
```
tomhjp, line 110: I think we'd need to only conditionally set these env vars if the helm values have been specified, and default the values to empty strings. See https://github.com/hashicorp/vault-k8s/pull/386 and its related comments for full details, but TL;DR this setting can affect the whole pod's ephemeral storage instead of just the injected container's if it's not set for the other containers.

krishnadas-m:
@tomhjp Is there anything pending to be done from my side to get this approved. Thanks in advance.

test/unit/injector-deployment.bats
```diff
@@ -834,6 +834,14 @@ EOF
...
       yq -r 'map(select(.name=="AGENT_INJECT_MEM_REQUEST")) | .[] .value' | tee /dev/stderr)
   [ "${value}" = "64Mi" ]
+
+  local value=$(echo $object |
+      yq -r 'map(select(.name=="AGENT_INJECT_EPHEMERAL_REQUEST")) | .[] .value' | tee /dev/stderr)
+  [ "${value}" = "64Mi" ]
+
+  local value=$(echo $object |
+      yq -r 'map(select(.name=="AGENT_INJECT_EPHEMERAL_LIMIT")) | .[] .value' | tee /dev/stderr)
+  [ "${value}" = "128Mi" ]
```
tomhjp, line 844:
```suggestion
  local value=$(echo $object |
      yq -r 'map(select(.name=="AGENT_INJECT_EPHEMERAL_REQUEST")) | length' | tee /dev/stderr)
  [ "${value}" -eq 0 ]

  local value=$(echo $object |
      yq -r 'map(select(.name=="AGENT_INJECT_EPHEMERAL_LIMIT")) | length' | tee /dev/stderr)
  [ "${value}" -eq 0 ]
```

tomhjp (commented):
Sorry for the delay, there's just one unit test failure but otherwise LGTM.

tomhjp:
Oh and please could you add a small changelog entry? e.g. Make an Improvements section and add:

```
injector: Add `ephemeralLimit` and `ephemeralRequest` options for defaulting Agent's ephemeral storage resources [GH-798](https://github.com/hashicorp/vault-helm/pull/798)
```

test/unit/injector-deployment.bats
krishnadas-m: @tomhjp I have removed this from the first test because we have made it an optional component.

tomhjp: I think it's nice to have the assertion that it's not there in a test, because it's important that by default it doesn't get set.

tomhjp: I tested with my suggestion and it fixed the test fwiw

krishnadas-m: yeah, it is working when we don't add value, but if we add value then it fails as the length becomes 1

ainnabilaaa:
is there anything else that needs to be fixed for this PR? We need this feature ASAP, hopefully it can be merged soon on the next update, thanks

tomhjp (approved):
LGTM, thanks!

PR #804, opened 2022-10-26
Fix CircleCI config

tomhjp:
Turns out the docker image I upgraded to previously doesn't exist: https://app.circleci.com/pipelines/github/hashicorp/vault-helm/1395/workflows/2e5b8ff3-3f4f-4276-9cc7-842b30269837/jobs/1776

```shell-session
$ docker run --rm -it docker.mirror.hashicorp.services/golang:1.19.2 /bin/bash
root@e550917e673a:/go# which curl
/usr/bin/curl
root@e550917e673a:/go# go install github.com/mikefarah/yq/v2@latest
go: downloading github.com/mikefarah/yq/v2 v2.4.1
go: downloading github.com/mikefarah/yq v2.4.0+incompatible
go: downloading github.com/pkg/errors v0.8.1
go: downloading github.com/mikefarah/yaml/v2 v2.4.0
go: downloading github.com/spf13/cobra v0.0.5
go: downloading gopkg.in/imdario/mergo.v0 v0.3.7
go: downloading gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473
go: downloading github.com/spf13/pflag v1.0.3
root@e550917e673a:/go# which yq
/go/bin/yq
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,13 +5,13 @@ orbs:
 jobs:
   update-helm-charts-index:
     docker:
-      - image: docker.mirror.hashicorp.services/circleci/golang:1.19.2
+      - image: docker.mirror.hashicorp.services/golang:1.19.2
     steps:
       - checkout
       - run:
           name: verify Chart version matches tag version
           command: |
-            GO111MODULE=on go get github.com/mikefarah/yq/v2
+            go install github.com/mikefarah/yq/v2@latest
             git_tag=$(echo "${CIRCLE_TAG#v}")
             chart_tag=$(yq r Chart.yaml version)
             if [ "${git_tag}" != "${chart_tag}" ]; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,9 +10,12 @@ jobs:
       - checkout
       - run:
           name: verify Chart version matches tag version
+          environment:
+            RELEASE_TAG: << pipeline.parameters.release-tag >>
           command: |
             go install github.com/mikefarah/yq/v2@latest
-            git_tag=$(echo "${CIRCLE_TAG#v}")
+            export RELEASE_TAG=${RELEASE_TAG:-$CIRCLE_TAG}
+            git_tag=$(echo "${RELEASE_TAG#v}")
             chart_tag=$(yq r Chart.yaml version)
             if [ "${git_tag}" != "${chart_tag}" ]; then
               echo "chart version (${chart_tag}) did not match git version (${git_tag})"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,13 +5,13 @@ orbs:
 jobs:
   update-helm-charts-index:
     docker:
-      - image: docker.mirror.hashicorp.services/circleci/golang:1.19.2
+      - image: docker.mirror.hashicorp.services/golang:1.19.2
```
tvoran, line 8: Ah, yeah I think the circle images are now under `cimg`. So this might work too:

```suggestion
      - image: docker.mirror.hashicorp.services/cimg/go:1.19.2
```

https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/

tvoran (approved):

.circleci/config.yml
tomhjp: Oh awesome, TIL, thanks!

tomhjp:
Because we rely on a CircleCI secret to trigger the release, it's kind of awkward to release the helm chart now that I broke the CI at the tag's commit. I've added a version of the pipeline that can be triggered manually so that I can trigger the release once this is merged. LMK if you think the approach seems sound, or if we should solve it a different way.

tomhjp:
Also here's an example where I manually triggered it with the curl command commented out, and instead just told it to echo the version being released: https://app.circleci.com/pipelines/github/hashicorp/vault-helm/1404/workflows/5458c273-b5d3-4448-9d52-35bf967643ac/jobs/1779

swenson (approved):
LGTM

.circleci/config.yml
```diff
@@ -5,32 +5,43 @@ orbs:
...
           name: update helm-charts index
+          environment:
+            RELEASE_TAG: << pipeline.parameters.release-tag >>
           command: |
             curl --show-error --silent --fail --user "${CIRCLE_TOKEN}:" \
                 -X POST \
                 -H 'Content-Type: application/json' \
                 -H 'Accept: application/json' \
-                -d "{\"branch\": \"main\",\"parameters\":{\"SOURCE_REPO\": \"${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}\",\"SOURCE_TAG\": \"${CIRCLE_TAG}\"}}" \
+                -d "{\"branch\": \"main\",\"parameters\":{\"SOURCE_REPO\": \"${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}\",\"SOURCE_TAG\": \"${RELEASE_TAG}\"}}" \
```
tvoran, line 33: Will `RELEASE_TAG` be set when this is not manually triggered? I wonder if we need to stash `TAG` in `$BASH_ENV` up above.

tvoran (commented):
I'd also be fine with just bumping the version again, but I get that's a little messy.

.circleci/config.yml
tomhjp: Oops, good catch. I repeated the same `:-$CIRCLE_TAG` defaulting in this step instead, which should also work

tvoran (approved):

PR #806, opened 2022-10-31
Add extraLabels for Vault server serviceAccount

AleksandrTitov:


tvoran:
Thanks, this looks pretty straightforward. Would you mind adding a test for this in `test/unit/server-serviceaccount.bats`? Something similar to the one for csi: https://github.com/hashicorp/vault-helm/blob/46e6fb5ad19f847ec70d137e1f609b50504cb6a3/test/unit/csi-serviceaccount.bats#L61-L72

AleksandrTitov:
Hi @tvoran , thanks, for your comment. I added an additional change with unit test for the Vault Server service account extra labels value.

tvoran (approved):

PR #811, opened 2022-11-10
Support selectively disabling active/standby services

tomhjp:
- Selectively disable the active and/or standby services
- Selectively disable the service discovery role binding for the Vault server service account

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,11 @@
 ## Unreleased
 
-Features:
+Changes:
+* server: No longer apply Vault service discovery role and role binding when `server.serviceAccount.create` is set to false [GH-811](https://github.com/hashicorp/vault-helm/pull/811)
+
+Improvements:
 * server: Add `extraLabels` for Vault server serviceAccount [GH-806](https://github.com/hashicorp/vault-helm/pull/806)
+* server: Add `server.service.active.enabled` and `server.service.standby.enabled` options to selectively disable additional services [GH-811](https://github.com/hashicorp/vault-helm/pull/811)
 
 ## 0.22.1 (October 26th, 2022)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,8 @@
 ## Unreleased
 
-Improvements:
+Features:
 * server: Add `extraLabels` for Vault server serviceAccount [GH-806](https://github.com/hashicorp/vault-helm/pull/806)
 * server: Add `server.service.active.enabled` and `server.service.standby.enabled` options to selectively disable additional services [GH-811](https://github.com/hashicorp/vault-helm/pull/811)
-
 * server: Add `server.serviceAccount.serviceDiscovery.enabled` option to selectively disable a Vault service discovery role and role binding [GH-811](https://github.com/hashicorp/vault-helm/pull/811)
 
 ## 0.22.1 (October 26th, 2022)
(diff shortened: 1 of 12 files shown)


templates/server-discovery-role.yaml
```diff
@@ -1,7 +1,8 @@
 {{ template "vault.mode" . }}
-{{- if ne .mode "external" }}
```
tvoran, line 2: I think we may want to leave the external checks in place for the discovery role and binding? IIRC the external mode was setup to only require setting `injector.externalVaultAddr` to disable all the server things in the chart (or at least most of them): https://developer.hashicorp.com/vault/docs/platform/k8s/helm/run#external-mode

templates/server-discovery-role.yaml
```diff
@@ -1,7 +1,8 @@
...
-{{- if .serverEnabled -}}
+{{ template "vault.serverServiceAccountEnabled" . }}
 {{- if eq .mode "ha" }}
+{{- if .serverEnabled -}}
+{{- if .serverServiceAccountEnabled -}}
```
tvoran, line 5: I think disabling the discovery role and binding if the server isn't enabled makes sense, but I think we may want to still create them even if the chart isn't creating the service account. Just thinking that the service account that the user supplies via `server.serviceAccount.name` will still need these capabilities and it's convenient if the chart just adds them for the user.

values.yaml
```diff
@@ -596,6 +596,10 @@ server:
   # Enables a headless service to be used by the Vault Statefulset
   service:
     enabled: true
+    active:
+      enabled: true
+    standby:
+      enabled: true
```
tvoran, line 602: Alternatively I guess we could pop out the active and standby service options into their own sections, though that would be a lot of duplication.

templates/server-discovery-role.yaml
tomhjp: My thinking on this one was it's strictly a subset of the `{{- if eq .mode "ha" }}` check below, so unnecessary. But happy to put it back in if it might be wrong/risky.

templates/server-discovery-role.yaml
tomhjp: Yeah that's fair, I'll add a net-new config option to disable it instead

values.yaml
tomhjp: Yeah if there are no objections, I would prefer to take a very small incremental step on this, as it already feels very granular to individually turn the services on/off. Perhaps it will make sense to split it out if we start getting a lot of feature requests to change more about each individual service in the future though.

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 ## Unreleased
 
-Features:
+Improvements:
```
tvoran, line 3: I think new chart options have been considered "features" in the past for this chart? Though looking at the changelog it doesn't seem very consistent in that respect 🤷.

values.schema.json
```diff
@@ -895,8 +911,19 @@
                         "create": {
                             "type": "boolean"
                         },
+                        "extraLabels": {
+                            "type": "object"
+                        },
```
tvoran, line 916: Thanks for catching this!

tvoran (approved):

CHANGELOG.md
tomhjp: Yeah I think you're right though that features is the more common designation for the helm chart. I was thinking along the lines of the `vault` repo guidelines. I'll change this back as it's a separate discussion from this PR.

PR #813, opened 2022-11-17
server: Allow disabling the instance selector for services

tomhjp:
Allows users to optionally loosen the selectors for services a little to select other deployments of Vault as well. If the new option is disabled, the service will essentially match all Vault pods within the same namespace.

I would have preferred to use a [set selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) but unfortunately the `Service` schema doesn't support that.


swenson (approved):
LGTM

tvoran (approved):

PR #823, opened 2022-12-23
Default prometheusRules.rules should be an empty list, not map

vitaliyf:
Support for prometheus-operator was added in https://github.com/hashicorp/vault-helm/pull/772 and a default empty set of rules was defined as an empty map `{}`. However, as evidenced by the commented out rule examples below that very same values.yaml, this is expected to be a list, so `rules:` value should be set to an empty list `[]`.

Because of this Helm issues a warning if you actually define rules and pass them with a --values <filename>:

```
coalesce.go:220: warning: cannot overwrite table with non table for vault.serverTelemetry.prometheusRules.rules (map[])
```

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1125,7 +1125,7 @@ serverTelemetry:
       selectors: {}
 
       # Some example rules.
-      rules: {}
+      rules: []
       #  - alert: vault-HighResponseTime
       #    annotations:
       #      message: The response time of Vault is over 500ms on average over the last 5 minutes.


tvoran:
Hi @vitaliyf, good spot! It looks like you're right, since PrometheusRule.spec.groups.rules is [defined as a list in the operator](https://github.com/prometheus-operator/prometheus-operator/blob/e4e27052f57040f073c6c1e4aedaecaaec77d170/pkg/apis/monitoring/v1/types.go#L1849).

We should update the tests in `test/unit/prometheus-prometheusrules.bats`, and also [update the json schema](https://github.com/hashicorp/vault-helm/blob/main/test/README.md#generating-the-values-json-schema) to ensure that only an array is accepted for `vault.serverTelemetry.prometheusRules.rules`.

tvoran:
Superseded by #886.

PR #828, opened 2023-01-11
Amending comment for the vault service account creation helper

dtrts:
Spotted a small inconsistency with a comment in the `_helpers.tpl` file.


tvoran (approved):
Good spot, thanks!

Issue #832, opened 2023-01-24
Support for setting ipFamilyPolicy on service


UXabre:
**Is your feature request related to a problem? Please describe.**
Deployment works great in a dual-stacked set-up, however, there is no way for me to set the ipFamilyPolicy to "preferDualStack"

**Describe the solution you'd like**
Being able to pass this variable would be great, but to be on the safe side, I'd keep it set to singlestack (to keep working as currently expected).

**Describe alternatives you've considered**
The alternative is me manually adjusting the service to include this.

UXabre:
Just thinking out loud here... should I also include the ipFamilies option to be complete? This way a singlestack ipv6 could be created as well

UXabre:
Meanwhile I've added this ipFamilies thingamajig to the PR; if somebody could review this?

PR #833, opened 2023-01-24
Add support for dual stack clusters

UXabre:
Adds support for dual stack clusters

resolves: https://github.com/hashicorp/vault-helm/issues/832

### DIFF
--- a/templates/ui-service.yaml
+++ b/templates/ui-service.yaml
@@ -20,6 +20,14 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
   {{- template "vault.ui.annotations" . }}
 spec:
+  {{- if (semverCompare ">= 1.23-0" .Capabilities.KubeVersion.Version) }}
+  {{- if .Values.ui.serviceIPFamilyPolicy }}
+  ipFamilyPolicy: {{ .Values.ui.serviceIPFamilyPolicy }}
+  {{- end }}
+  {{- if .Values.ui.serviceIPFamilies }}
+  ipFamilies: {{ .Values.ui.serviceIPFamilies | toYaml | nindent 2 }}
+  {{- end }}
+  {{- end }}
   selector:
     app.kubernetes.io/name: {{ include "vault.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
--- a/test/unit/server-service.bats
+++ b/test/unit/server-service.bats
@@ -467,3 +467,45 @@ load _helpers
       yq -r '.spec.selector["app.kubernetes.io/instance"]' | tee /dev/stderr)
   [ "${actual}" = "null" ]
 }
+
+@test "server/Service: Assert ipFamilyPolicy set" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-service.yaml \
+      --set 'server.service.ipFamilyPolicy=PreferDualStack' \
+      . | tee /dev/stderr |
+      yq -r '.spec.ipFamilyPolicy' | tee /dev/stderr)
+  [ "${actual}" = "PreferDualStack" ]
+}
+
+@test "server/Service: Assert ipFamilies set" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-service.yaml \
+      --set 'server.service.ipFamilies={IPv4,IPv6}' \
+      . | tee /dev/stderr |
+      yq '.spec.ipFamilies' -c | tee /dev/stderr)
+  [ "${actual}" = '["IPv4","IPv6"]' ]
+}
+
+@test "server/Service: Assert ipFamilyPolicy is not set if version below 1.23" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-service.yaml \
+      --kube-version 1.22.0 \
+      --set 'server.service.ipFamilyPolicy=PreferDualStack' \
+      . | tee /dev/stderr |
+      yq -r '.spec.ipFamilyPolicy' | tee /dev/stderr)
+  [ "${actual}" = "null" ]
+}
+
+@test "server/Service: Assert ipFamilies is not set if version below 1.23" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-service.yaml \
+      --kube-version 1.22.0 \
+      --set 'server.service.ipFamilies={IPv4,IPv6}' \
+      . | tee /dev/stderr |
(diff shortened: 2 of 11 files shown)


M0NsTeRRR:
Looking for this feature too, as I see on other PR @UXabre maybe you can ask for a review of hashicorp team to push it further ?

UXabre:
I've updated this PR to resolve the conflict, if somebody could review this? :-)

hsimon-hashicorp:
> @hsimon-hashicorp Maybe you can help me with this PR? As I see nobody has reviewed it yet and perhaps you are my last hope :-) (sorry for spamming)

Not spam at all, this is my job. :) I've pinged our engineers to see if one can help with a review. Thanks!

belyaev-dev:
joining waitlist :)

tvoran (commented):
Thanks, left a couple thoughts. And should we set these parameters in `templates/server-headless-service.yaml` as well?

test/unit/server-service.bats
```diff
@@ -467,3 +467,23 @@ load _helpers
...
+
+@test "server/Service: Assert ipFamilyPolicy set" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-service.yaml \
+      --set 'server.service.ipFamilyPolicy=PreferDualStack' \
+      . | tee /dev/stderr |
+      yq -r '.spec.ipFamilyPolicy' | tee /dev/stderr)
+  [ "${actual}" = "PreferDualStack" ]
+}
```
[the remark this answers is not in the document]
UXabre: Added these tests as requested

values.yaml
```diff
@@ -676,6 +676,23 @@ server:
...
+    #     SingleStack: Single-stack service. The control plane allocates a cluster IP for the Service, using the first configured service cluster IP range.
+    #     PreferDualStack: Allocates IPv4 and IPv6 cluster IPs for the Service.
+    #     RequireDualStack: Allocates Service .spec.ClusterIPs from both IPv4 and IPv6 address ranges.
+    # ipFamilyPolicy: SingleStack
+    # 
+    # Sets the families that should be supported and the order in which they should be applied to ClusterIP as well
+    #
+    # ipFamilies:
+    # - IPv4
+
```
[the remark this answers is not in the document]
UXabre: Changed the defaults to these proposed

UXabre:
> Thanks, left a couple thoughts. And should we set these parameters in `templates/server-headless-service.yaml` as well?

I've also added this to headless :-)

UXabre:
All changes have been applied and I've rebased with main just to be on the safe side :-)

templates/ui-service.yaml
```diff
@@ -20,6 +20,15 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
   {{- template "vault.ui.annotations" . }}
 spec:
+
```
tvoran, line 23:
```suggestion
```

values.yaml
```diff
@@ -676,6 +676,22 @@ server:
     # or NodePort.
     #type: ClusterIP
 
+    # The IP family and IP families options are to set the behaviour in a dual-stack environment
+    # Omitting these values will let the service fall back to whatever the CNI dictates the defaults
+    # should be
```
tvoran, line 681:
```suggestion
    # The IP family and IP families options are to set the behaviour in a dual-stack environment.
    # Omitting these values will let the service fall back to whatever the CNI dictates the defaults
    # should be.
```

values.yaml
```diff
@@ -676,6 +676,22 @@ server:
...
+    #
+    # Configures the service's supported IP family policy, can be either:
+    #     SingleStack: Single-stack service. The control plane allocates a cluster IP for the Service, using the first configured service cluster IP range.
+    #     PreferDualStack: Allocates IPv4 and IPv6 cluster IPs for the Service.
+    #     RequireDualStack: Allocates Service .spec.ClusterIPs from both IPv4 and IPv6 address ranges.
+    ipFamilyPolicy: ""
+    # 
+    # Sets the families that should be supported and the order in which they should be applied to ClusterIP as well
+    #
+    ipFamilies: []
```
tvoran, line 693:
```suggestion
    # Sets the families that should be supported and the order in which they should be applied to ClusterIP as well.
    # Can be IPv4 and/or IPv6
    ipFamilies: []
```

values.yaml
```diff
@@ -979,6 +995,21 @@ ui:
...
+  # Sets the families that should be supported and the order in which they should be applied to ClusterIP as well
+  # Can be IPv4 and/or IPv6
+  serviceIpFamilies: []
+
+  # Configures the service's supported IP family, can be either:
+  #     SingleStack: Single-stack service. The control plane allocates a cluster IP for the Service, using the first configured service cluster IP range.
+  #     PreferDualStack: Allocates IPv4 and IPv6 cluster IPs for the Service.
+  #     RequireDualStack: Allocates Service .spec.ClusterIPs from both IPv4 and IPv6 address ranges.
+  # Only supported for kubernetes versions >=1.23.0
+  serviceIpFamilyPolicy: ""
```
tvoran, line 1011: Minor things, but I'd suggest making the order of the parameters and the doc strings match what's in `server.service`, and also uppercase the "IP" in "serviceIP...".

```suggestion
  # The IP family and IP families options are to set the behaviour in a dual-stack environment.
  # Omitting these values will let the service fall back to whatever the CNI dictates the defaults
  # should be.
  #
  # Only supported for kubernetes versions >=1.23.0
  #
  # Configures the service's supported IP family, can be either:
  #     SingleStack: Single-stack service. The control plane allocates a cluster IP for the Service, using the first configured service cluster IP range.
  #     PreferDualStack: Allocates IPv4 and IPv6 cluster IPs for the Service.
  #     RequireDualStack: Allocates Service .spec.ClusterIPs from both IPv4 and IPv6 address ranges.
  serviceIPFamilyPolicy: ""
  #
  # Sets the families that should be supported and the order in which they should be applied to ClusterIP as well.
  # Can be IPv4 and/or IPv6
  serviceIPFamilies: []
```

templates/ui-service.yaml
UXabre: Removed empty line

values.yaml
UXabre: Added punctuation

values.yaml
UXabre: Yes, I agree, it should be consistent! I've made the changes as requested

values.yaml
UXabre: Added this

tvoran (approved):
Thanks!

PR #834, opened 2023-01-26
Add servicemonitor relabelings

raphaelzoellner:
Adds support for specifying metricRelabelings and relabelings fields of `ServiceMonitor.spec.endpoints`. 
https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.Endpoint

This is useful when labels of the metrics collide with external labels of the Prometheus Server (e.g. `cluster`) and allows one to replace the colliding label name with a new label name.


sunggun-yu:
Hi, any news on this one?

PR #841, opened 2023-02-06
Add extraPorts property

mitashkistz:
Issue:
We are using the upstream helm chart and we added a new listener for monitoring, running on custom port.
The current chart do not allows us to add this custom port to the statefulset from the values file.
Resolution:
These changes allows us to add custom ports to values file and continue using the upstream helm chart at the same time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+Features:
+* server: New `extraPorts` option for adding ports to the Vault server statefulset [GH-841](https://github.com/hashicorp/vault-helm/pull/841)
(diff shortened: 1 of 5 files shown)


tvoran (commented):
Thanks! Could you add a couple tests for this in `test/unit/server-statefulset.bats`? And [update the schema](https://github.com/hashicorp/vault-helm/blob/742ddb1c52934e657ed7d4832fcf80324806adfb/test/README.md#L39) in values.schema.json? Otherwise I think it looks fine.

mitashkistz:
Hi, it is all done.

PR #844, opened 2023-02-11
Adding SPDX copyright headers

tvoran:
Basically the same as #838, but using [template style comments](https://helm.sh/docs/chart_best_practices/templates/#comments-yaml-comments-vs-template-comments) in the templates so that helm doesn't render them and err on yaml that should be empty.


kschoche (approved):
Looks great thanks!

swenson (approved):
LGTM

PR #846, opened 2023-02-15
Fix typo in telemetry example

jordanfelle:
The example for the telemetry block has a comma at the end of the list, which is not valid HCL and causes vault to crashloop if used as is.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -747,7 +747,7 @@ server:
 
       # Example configuration for enabling Prometheus metrics in your config.
       #telemetry {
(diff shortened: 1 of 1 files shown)


tvoran:
Hi @jordanfelle, would you mind changing `test/acceptance/server-test/telemetry.yaml` while you're at it? It isn't causing Vault to crashloop from what I can tell, but I agree it's not quite correct HCL.

tvoran (approved):
Thanks!

PR #848, opened 2023-02-24
fix: remove k8s 1.16 from acceptance testing

thyton:
As `kind` no longer supports creating a k8s 1.16 cluster, we remove 1.16 from the versions tested in `.github/workflows/acceptance.yaml`, and update vault-helm's minimum support k8s version to 1.20 in README and `Chart.yaml`

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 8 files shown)


tvoran (approved):
Nice! It's great to see the all tests green again 🙂 

It would be good to add an entry to the Unreleased section of the Changelog about this, just to increase awareness. Something like:

```markdown
Changes:
* Earliest Kubernetes version tested is now 1.20
```

thyton:
Thank you. I will add the entry

PR #852, opened 2023-03-15
feat: make injector livenessProbe and readinessProbe configurable and add configurable startupProbe

thyton:
* startupProbe default is based on https://github.com/hashicorp/vault-k8s/issues/388#issuecomment-1274154311
* add injector deployment tests to test default and custom livenessProbe, readinessProbe, and startupProbe

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -93,6 +93,42 @@ injector:
       exitOnRetryFailure: true
       staticSecretRenderInterval: ""
 
+  # Used to enable a livenessProbe for the pods
+  livenessProbe:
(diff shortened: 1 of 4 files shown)


values.yaml
```diff
@@ -93,6 +93,43 @@ injector:
       exitOnRetryFailure: true
       staticSecretRenderInterval: ""
 
+  # Used to enable a livenessProbe for the pods
```
tomhjp, line 96: Very picky nit, as the liveness probe is always enabled.

```suggestion
  # Used to configure a livenessProbe for the pods
```

tomhjp (approved):
LGTM!

PR #855, opened 2023-03-21
Fix yaml upon empty volumeMounts or volumes

thoraage:
Today it create illegal yaml with undefined volumeMounts and volumes yielding these messages from kubeval: WARN - stdin contains an invalid Pod (vault.helm-vault-server-test) - spec.containers.0.volumeMounts: Invalid type. Expected: array, given: null WARN - stdin contains an invalid Pod (vault.helm-vault-server-test) - spec.volumes: Invalid type. Expected: array, given: null

Instead; insert empty lists.


templates/tests/server-test.yaml
```diff
@@ -46,10 +46,14 @@ spec:
       volumeMounts:
         {{- if .Values.server.volumeMounts }}
```
arve0, line 47: Could also move the if statement one line up, to make it include the key, in example:

```
     {{- if .Values.server.volumeMounts }}
      volumeMounts:
          {{- toYaml .Values.server.volumeMounts | nindent 8}}
     {{- end }}
```

tvoran (commented):
I think @arve0's suggestion makes sense, and it would also be good to add a test to `test/unit/server-test.bats` to check that the rendering for empty volumes is correct.

templates/tests/server-test.yaml
thoraage: Changed it now (better late than never)

tvoran (commented):
Thanks! I think the only thing missing here is a test in `test/unit/server-test.bats` to verify that `volumeMounts` and `volumes` are not rendered when `server.volumeMounts` and `server.volumes` are not set.

thoraage:
Added some tests.... a long time after

tvoran (approved):
Thanks!

PR #861, opened 2023-03-27
Convert hashicorp/vault-helm to GitHub Actions

hc-github-team-es-release-engineering:
Pipeline migrated from [CircleCI](https://app.circleci.com/pipelines/github/hashicorp/vault-helm) :tada:

## Manual steps

Perform the following steps to complete the migration:

### hashicorp/vault-helm/update-helm-charts-index
- [x] Ensure environment variable is updated: `SLACK_WEBHOOK`
- [x] Ensure environment variable is updated: `VAULT_LICENSE_CI`


dekimsey:
This fixes [RELENG-400](https://hashicorp.atlassian.net/browse/RELENG-400)

tomhjp:
@dekimsey Do lmk when this is ready for review and I'll be happy to help land this.

.github/actions/slack-status/action.yml
```diff
@@ -0,0 +1,37 @@
+name: "Send Slack Status Notification"
```
tomhjp, line 1: Could we move this to the workflows directory? I'd prefer to keep it as un-special as possible so all the tooling we're building around github actions doesn't need to be smart to handle this pattern

.github/actions/slack-status/action.yml
```diff
@@ -0,0 +1,37 @@
...
+  block-color:
+    description: HEX color code, use `ed5c5c` for failure and `1cbf43` for success status
+    required: false
+    default: 1cbf43
+
+runs:
+  using: "composite"
+  steps:
+    - name: Send Slack Notification
+      uses: slackapi/slack-github-action@v1.23.0
```
tomhjp, line 23: Should this be pinned to a SHA?

.github/workflows/update-helm-charts-index.yml
```diff
@@ -0,0 +1,53 @@
...
+          esac
+      - name: Send Slack Notification
+        # note: this will skip Slack notification if the job is skipped or cancelled.
+        if: ${{ (success() || failure()) && env.SLACK_MESSAGE != '' }}
+        uses: ./.github/actions/slack-status
+        with:
+          block-color: ${{ env.SLACK_BLOCK_COLOR }}
+          message: ${{ env.SLACK_MESSAGE }}
+          run-url: ${{ env.RUN_URL }}
+          slack-webhook-url: ${{ secrets.slack_webhook_url }}
```
tomhjp, line 53: I named the secret `SLACK_WEBHOOK` as per the PR description, but looks like this is expecting a different name?

.github/workflows/update-helm-charts-index.yml
```diff
@@ -0,0 +1,53 @@
...
+          if [ "${git_tag}" != "${chart_tag}" ]; then
+            echo "chart version (${chart_tag}) did not match git version (${git_tag})"
+            exit 1
+          fi
+        env:
+          RELEASE_TAG: "${{ inputs.release-tag }}"
+      - name: update helm-charts index
+        id: update
+        run: |-
+          export GITHUB_TOKEN="${{secrets.helm_charts_github_token }}"
```
tomhjp, line 26: I don't think this secret exists as a repository or org secret. Is it an org secret that needs to be shared with this repo?

tomhjp (commented):
Generally LGTM, just a few comments

.github/workflows/update-helm-charts-index.yml
dekimsey: I think you (as the owners) need to create a PAT with your team's service account and add it to the repository's secrets. It needs to have `actions:write` to the `hashicorp/helm-charts` repository. To my knowledge, we don't use org-wide ones (since that'd be shared secrets and all that jazz).

.github/workflows/update-helm-charts-index.yml
dekimsey: Yeah, that's my mistake. I'll correct the capitalization. They aren't POSIX vars anymore, so that's just a habit of mine.

.github/actions/slack-status/action.yml
dekimsey: So if I'm understanding this correctly, it's a composite action. Which differs from a reusable workflow. If we put it in the workflows, then it appears as it's own entity in your Actions tab in GH and would look like something you one can trigger. By making it an action, it basically acts like a single step in the update-helm-charts-index workflow.

If it's a concern, I think we'd probably be better off refactoring it out to a separate repository instead since we do appear to have a few copies floating around at this point.

.github/actions/slack-status/action.yml
dekimsey: I think security is supposed to have a bot that'll come through and do this, but I can certainly fix it now!

.github/actions/slack-status/action.yml
tomhjp: Thanks for looking into it! It would be nice to have it in a common repo somewhere, but I'm ok with leaving that for another PR.

.github/workflows/update-helm-charts-index.yml
```diff
@@ -0,0 +1,56 @@
...
+              ;;
+          esac
+      - name: Send Slack Notification
+        # note: this will skip Slack notification if the job is skipped or cancelled.
+        if: ${{ (success() || failure()) && env.SLACK_MESSAGE != '' }}
+        uses: ./.github/actions/slack-status
+        with:
+          block-color: ${{ env.SLACK_BLOCK_COLOR }}
+          message: ${{ env.SLACK_MESSAGE }}
+          run-url: ${{ env.RUN_URL }}
```
emilymianeil, line 55: I believe this needs to be set somewhere ( accessible via Github context 😄 )

emilymianeil (changes requested):
Just left a comment about the `RUN_URL` needing to be set 😄

.github/actions/slack-status/action.yml
```diff
@@ -0,0 +1,37 @@
...
+        SLACK_WEBHOOK_URL: ${{ inputs.slack-webhook-url }}
+        SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+      with:
+        payload: '
+          {
+            "attachments": [
+              {
+                "mrkdwn_in": ["text"],
+                "color": "#${{ inputs.block-color }}",
+                "text": "${{ inputs.message }}\n<${{ input.run-url }}|View Workflow>"
```
emilymianeil, line 34: Ran into this as well - I believe it should be `inputs.run-url`

.github/workflows/update-helm-charts-index.yml
```diff
@@ -0,0 +1,56 @@
+name: hashicorp/vault-helm/update-helm-charts-index
+on:
+  push:
+    tags:
+      - 'v[0-9]+\.[0-9]+\.[0-9]+\.*'
```
alvin-huang, line 5:
```suggestion
      - 'v[0-9]+\.[0-9]+\.[0-9]+*'
```
This tag shouldn't have the trailing dot on the last patch version since that makes v#.#.# incompatible

.github/workflows/update-helm-charts-index.yml
```diff
@@ -0,0 +1,56 @@
+name: hashicorp/vault-helm/update-helm-charts-index
```
alvin-huang, line 1: We've decided to only keep the last part of these even though the gh-actions-importer adds the full name
```suggestion
name: update-helm-charts-index
```

.github/workflows/update-helm-charts-index.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: |-
+          export TAG=${RELEASE_TAG:-${{ github.ref_name }}}
+          git_tag=$(echo "${TAG#v}")
+          chart_tag=$(yq r Chart.yaml version)
+          if [ "${git_tag}" != "${chart_tag}" ]; then
+            echo "chart version (${chart_tag}) did not match git version (${git_tag})"
+            exit 1
+          fi
+        env:
+          RELEASE_TAG: "${{ inputs.release-tag }}"
```
alvin-huang, line 25: I don't see the input block for this variable

.github/workflows/update-helm-charts-index.yml
```diff
@@ -0,0 +1,56 @@
...
+            --repo hashicorp/helm-charts \
+            --ref main \
+            -f SOURCE_TAG="${{ github.ref_name }}" \
+            -f SOURCE_REPO="${{ github.repository }}"
+      - name: Prepare Slack Notification
+        if: always()
+        run: |
+          case "${{ steps.update.conclusion }}" in
+            failure)
+              echo "SLACK_MESSAGE=:red_circle: The vault-helm charts index update failed." >> "$GITHUB_ENV"
```
alvin-huang, line 40: I think this message (and the one below) is slightly misleading. The vault-helm chart's index update didn't complete but the trigger to kick off the `publish-charts` workflow on `hashicorp/helm-charts` was successful (if `gh workflow run` didn't return nonzero). 

Maybe rewording this to something like 
```
:tada: vault-helm charts index update triggered successfully. View the run at https://github.com/hashicorp/helm-charts/actions/workflows/publish-charts.yml`
```

swenson:
This looks good to me. I'd also be okay if the Slack message is deleted, if it simplifies getting this merged.

Is there a blocker for merging this now?

dekimsey:
Need to confirm the secrets are correct. Specifically, I need to replace the slack webhook one we produce. But I didn't want to do that willy-nilly on your repository :)

.github/workflows/update-helm-charts-index.yml
dekimsey: Everything else looks good, but we do need a github PAT to be defined in the repo's secrets before this can be merged

dekimsey: @swenson added the neccessary secret as `HELM_CHARTS_GITHUB_TOKEN`. I think that resolves this now

dekimsey (approved):

emilymianeil (approved):
🥳 LGTM !

swenson (approved):
LGTM

PR #863, opened 2023-04-01
Updating GHA and default Vault version

tvoran:
Test with [latest kind k8s versions](https://hub.docker.com/r/kindest/node/tags) 1.22-1.26. Remove support for old disruptionbudget and ingress APIs (pre 1.22). 

Pin all actions to SHAs, and use the common jira sync.

Update the default Vault version to v1.13.1. 

Update chart-verifier used in tests to 1.10.1, also add an openshift name annotation to Chart.yaml (one of the required checks).

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -7,19 +7,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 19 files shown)


tomhjp (approved):
LGTM 👍

PR #872, opened 2023-04-12
Add role for creating CSI's HMAC secret key

tomhjp:
https://github.com/hashicorp/vault-csi-provider/pull/198 introduced the need to create a k8s secret for storing an HMAC key. This PR adds the role and role binding with permissions to do so while keeping the least privileges possible.

### DIFF
--- a/test/unit/csi-rolebinding.bats
+++ b/test/unit/csi-rolebinding.bats
@@ -5,7 +5,7 @@ load _helpers
 @test "csi/RoleBinding: disabled by default" {
   cd `chart_dir`
   local actual=$( (helm template \
-      --show-only templates/csi-role.yaml  \
+      --show-only templates/csi-rolebinding.yaml  \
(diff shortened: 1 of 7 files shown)


test/unit/csi-rolebinding.bats
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "csi/RoleBinding: disabled by default" {
+  cd `chart_dir`
+  local actual=$( (helm template \
+      --show-only templates/csi-role.yaml  \
```
tvoran, line 8: Should be rolebinding?

```suggestion
      --show-only templates/csi-rolebinding.yaml  \
```

tvoran (approved):

tvoran:
Looks like "csi/daemonset: extra args can be passed" needs to be updated as well.

PR #881, opened 2023-04-17
ci: Fix yq command syntax

dekimsey:
The original CCI version used an older version of yq. The syntax changed and this was missed when ported.

This addresses the error:

```
Error: 1:1: invalid input text "r"
```

### DIFF
--- a/.github/workflows/update-helm-charts-index.yml
+++ b/.github/workflows/update-helm-charts-index.yml
@@ -16,7 +16,7 @@ jobs:
         run: |-
           export TAG=${{ github.ref_name }}
           git_tag=$(echo "${TAG#v}")
-          chart_tag=$(yq r Chart.yaml version)
+          chart_tag=$(yq '.version' Chart.yaml)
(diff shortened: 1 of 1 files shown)


tomhjp (approved):
Thanks!

.github/workflows/update-helm-charts-index.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         run: |-
           export TAG=${{ github.ref_name }}
           git_tag=$(echo "${TAG#v}")
-          chart_tag=$(yq r Chart.yaml version)
+          chart_tag=$(yq '.version' Chart.yaml)
```
tvoran, line 19: Might want to use the raw string, instead of the default JSON?

```suggestion
          chart_tag=$(yq -r '.version' Chart.yaml)
```

tomhjp: Good call 👍  I've added this while resolving the merge conflict.

tvoran (approved):

PR #886, opened 2023-05-11
Default prometheusRules.rules should be an empty list

rissson:
Revival of https://github.com/hashicorp/vault-helm/pull/823

Support for prometheus-operator was added in https://github.com/hashicorp/vault-helm/pull/772 and a default empty set of rules was defined as an empty map `{}`. However, as evidenced by the commented out rule examples below that very same values.yaml, this is expected to be a list, so `rules:` value should be set to an empty list `[]`.

Because of this Helm issues a warning if you actually define rules and pass them with a --values <filename>:

```
coalesce.go:220: warning: cannot overwrite table with non table for vault.serverTelemetry.prometheusRules.rules (map[])
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+Bugs:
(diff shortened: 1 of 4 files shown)


tvoran (approved):
Thanks!

PR #889, opened 2023-05-15
values.yaml updated to include explicit cluster_name

aphorise:
Setting an [explicit (HCL) `cluster_name`](https://developer.hashicorp.com/vault/docs/configuration#cluster_name) as part of the cluster roll-out / configuration is required without which an undesirable behavior arises where the reported Prometheus label does not correctly contain the cluster name - eg (at present):

```bash
curl -Lks ${VAULT_ADDR}v1/sys/metrics?format=prometheus | grep dr_primary{
  # vault_core_replication_dr_primary{cluster="",host="vault-2"} 0
```

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -3,6 +3,10 @@
 
 # Available parameters and their default values for the Vault chart.
 
+# Set cluster name instead of having it auto-generated.
+# examples may include: "vault1-gke-uswest", "vault2-eks-useast", etc.
+cluster_name: ""
+
 global:
   # enabled is the master enabled switch. Setting this to true or false
   # will enable or disable all the components within this chart by default.
@@ -834,6 +838,7 @@ server:
       # https://www.vaultproject.io/docs/platform/k8s/helm/run#protecting-sensitive-vault-configurations
       config: |
         ui = true
+        cluster_name = {{ .Values.cluster_name | default (printf "%s-k8s" (include "vault.name" .)) | quote }}
 
         listener "tcp" {
           tls_disable = 1


values.yaml
```diff
@@ -3,6 +3,10 @@
 
 # Available parameters and their default values for the Vault chart.
 
+# Set cluster name instead of having it auto-generated.
+# examples may include: "vault1-gke-uswest", "vault2-eks-useast", etc.
+cluster_name: ""
```
jasonodonnell, line 8: This is a Vault specific parameter and should not be at the root level but instead should be in `server.ha` (it's only used for HA).

values.yaml
```diff
@@ -834,6 +838,7 @@ server:
       # https://www.vaultproject.io/docs/platform/k8s/helm/run#protecting-sensitive-vault-configurations
       config: |
         ui = true
+        cluster_name = {{ .Values.cluster_name | default (printf "%s-k8s" (include "vault.name" .)) | quote }}
```
jasonodonnell, line 841: Does this actually work? I've never seen the templates used outside of the K8s yaml helm is generating.

jasonodonnell, line 841: You can't set a default here because if someone has already deployed Vault and it has a generated cluster name, this is going to introduce a new set of metrics.
```suggestion
        cluster_name = {{ .Values.cluster_name | default (printf "%s-k8s" (include "vault.name" .)) | quote }}
```

aphorise: Yes it works - I've tested it (with both explicit value set & nothing set) before making PR

aphorise: `vault.name` is just `vault` right? - so if nothing is set it will always result in the same value no?
If users change it then it may result in what you're describing.

aphorise: Example as per current `helm` documentation.
https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

jasonodonnell: I'm referencing Vault's code that generates the cluster name if it's not set in the config. It [generates a cluster name](https://github.com/hashicorp/vault/blob/main/vault/cluster.go#L170-L188) that looks something like this: `vault-cluster-b0435276`. If you now start setting it in the config with a default anyone who was collecting metrics using the Vault generated name will now have a new set metrics.

aphorise: I added defaults in case for new deployment no value is set. I'm open to any proposal that you may be kind enough to commit.

PR #897, opened 2023-05-21
Add hostAliases to chart

andrewroffey:
Add [hostAliases](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/) to the Vault Helm Chart so that custom entries can be added to /etc/hosts via values.yaml.

Comes in handy for adding host entries to 127.0.0.1 (e.g. to use a 'normal' cert rather than an IP SAN cert) and possibly other uses.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -442,6 +442,12 @@ server:
     #    hosts:
     #      - chart-example.local
 
+  # hostAliases is a list of aliases to be added to /etc/hosts. Specified as a YAML list.
+  hostAliases: null
+    # - ip: 127.0.0.1
+    #   hostnames:
+    #     - chart-example.local
+
   # OpenShift only - create a route to expose the service
   # By default the created route will be of type passthrough
   route:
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -442,6 +442,12 @@ server:
     #    hosts:
     #      - chart-example.local
 
+  # hostAliases is a list of aliases to be added to /etc/hosts. Specified as a YAML list.
+  hostAliases: null
```
tvoran, line 446: I think we could default this to an empty list, and that would make the json schema definition a little simpler (no `null` needed).

```suggestion
  hostAliases: []
```

tvoran (commented):
This is looking good, also would you mind adding some tests for this new parameter in test/unit/server-statefulset.bats?

values.yaml
andrewroffey: @tvoran thanks for taking a look! Sorry for the delay, things have been a bit busy for me but I'll add some tests and update the PR with the schema suggestion soon.

tombokombo:
@andrewroffey are you planning to add tests?

andrewroffey:
@tombokombo I have dragged my feet on this one, mostly because we aren't using the Helm chart anymore (for now). I'll add some tests on my repo next week but I won't resign the new CLA because of the license thing.

PR #902, opened 2023-05-30
publishNotReadyAddresses for headless service always true

tomhjp:
The headless service pods should always be discoverable even if they're not ready, as it's used for cluster joining. However, vault, vault-active, and vault-standby services should be able to set `publishNotReadyAddresses` to false to ensure client requests are always routed to a ready pod. Before this PR, all services used the same values.yaml setting, so it wasn't possible to have different values for the different services. Ideally, it would be nice to set the default `publishNotReadyAddresses` value to be false at some point, but I just wanted to keep this PR to the minimum possible (minor) breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Changes:
 * Latest Kubernetes version tested is now 1.27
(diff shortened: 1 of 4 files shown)


kschoche (approved):
LGTM!

tomhjp:
Thanks!

PR #903, opened 2023-05-30
Fix syntax for actionlint workflow

tomhjp:
See https://github.com/hashicorp/vault-helm/actions/runs/5122269586 for an example error.


benashz (approved):
LGTM!

tvoran:
IIRC actionlint doesn't support composite actions (https://github.com/rhysd/actionlint/issues/46). Is there a way to ignore the setup-test-tools action in this repo?

tomhjp:
> IIRC actionlint doesn't support composite actions. Is there a way to ignore the setup-test-tools action in this repo?

Thanks for pointing that issue out. I moved the composite action out into its own actions sub-folder within .github to fix this. The linter only checks `.github/workflows` and the action uses a slightly different syntax to workflows, so it seems reasonable to give it its own folder anyway.

tomhjp:
Thanks!

Issue #915, opened 2023-06-22
New version of helm chart not created since past few months


praveenjindal62:
**Is your feature request related to a problem? Please describe.**
New version of helm chart is not released in past few month. We need some new features in helm chart which are already part of helm chart in main branch. Is it possible to create new release with the latest commit in `main` branch ?

tomhjp:
Thanks for raising this - hopefully I can get a release out today if all goes smoothly with #916.

hsimon-hashicorp:
Looks like https://github.com/hashicorp/vault-helm/releases/tag/v0.25.0 is out - is that sufficient? Thanks! :)

PR #862, opened 2023-03-29
CSI nodeSelector, affinity

Toninh0:
This helps configuring the scheduling of CSI daemonset pods

Closes #842

PR #916, opened 2023-06-22
Prepare for 0.25.0 release

### LINKED ISSUE #913: Impossible to deploy on Kubernetes < v1.22 due to a hard version constraint in Chart.yaml
**Describe the bug**
The README says:

>Kubernetes 1.22+ - This is the earliest version of Kubernetes tested. It is possible that this chart works with earlier versions but it is untested.

However, in fact, it's impossible to run it on, let's say, Kubernetes v1.21 because there's an explicit version check:
https://github.com/hashicorp/vault-helm/blob/main/Chart.yaml#L8

Consequently, I get the following error:
```
dry-run upgrade for comparison failed: chart requires kubeVersion: >= 1.22.0-0 which is incompatible with Kubernetes v1.21.2
```

**To Reproduce**
1. Install an older chart on Kubernetes 1.21
2. Try upgrading to the latest chart version

**Expected behavior**
It's still possible to deploy the Helm chart on older Kubernetes versions but without any guarantees as stated in README.md

tomhjp:
# 0.25.0 (June 22, 2023)

## Changes

* Latest Kubernetes version tested is now 1.27
* server: Headless service ignores `server.service.publishNotReadyAddresses` setting and always sets it as `true` [GH-902](https://github.com/hashicorp/vault-helm/pull/902)
* `vault` updated to 1.14.0 [GH-916](https://github.com/hashicorp/vault-helm/pull/916)
* `vault-csi-provider` updated to 1.4.0 [GH-916](https://github.com/hashicorp/vault-helm/pull/916)

## Improvements

* CSI: Make `nodeSelector` and `affinity` configurable for CSI daemonset's pods [GH-862](https://github.com/hashicorp/vault-helm/pull/862)
* injector: Add `ephemeralLimit` and `ephemeralRequest` as options for configuring Agent's ephemeral storage resources [GH-798](https://github.com/hashicorp/vault-helm/pull/798)
* Minimum kubernetes version for chart reverted to 1.20.0 to allow installation on clusters older than the oldest tested version [GH-916](https://github.com/hashicorp/vault-helm/pull/916)

## Bugs

* server: Set the default for `prometheusRules.rules` to an empty list [GH-886](https://github.com/hashicorp/vault-helm/pull/886)

Closes #913 #915

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,17 @@
 ## Unreleased
 
+## 0.25.0 (June 22, 2023)
+
 Changes:
 * Latest Kubernetes version tested is now 1.27
 * server: Headless service ignores `server.service.publishNotReadyAddresses` setting and always sets it as `true` [GH-902](https://github.com/hashicorp/vault-helm/pull/902)
+* `vault` updated to 1.14.0 [GH-916](https://github.com/hashicorp/vault-helm/pull/916)
+* `vault-csi-provider` updated to 1.4.0 [GH-916](https://github.com/hashicorp/vault-helm/pull/916)
 
-Features:
+Improvements:
 * CSI: Make `nodeSelector` and `affinity` configurable for CSI daemonset's pods [GH-862](https://github.com/hashicorp/vault-helm/pull/862)
+* injector: Add `ephemeralLimit` and `ephemeralRequest` as options for configuring Agent's ephemeral storage resources [GH-798](https://github.com/hashicorp/vault-helm/pull/798)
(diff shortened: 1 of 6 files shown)


tomhjp:
Upgrading CSI to 1.4.0 required a small test assertion update. From the commit message:

> Starting in 1.4.0, the CSI provider caches Vault tokens locally. The main thing
> we want to check is that the Agent cache is being used so that it's doing the
> renewal legwork for any leased secrets, so check for the renewal log message instead
> because CSI won't auth over and over anymore.

benashz (approved):
LGTM!

tomhjp:
The chart-verifier failures are because 1.14.0-ubi hasn't appeared on the RedHat registry yet: https://catalog.redhat.com/software/containers/hashicorp/vault/5fda55bd2937386820429e0c

PR #909, opened 2023-06-11
Allows the release namespace to be overridden

KhizerJaan:
**Problem:**
- Vault can only be deployed in Release namespace.

**Root Cause:**
- namespace is hardcoded to `{{ .Release.Namespace }}`

**Solution:**
- Templatize namespaec in _helpers.tpl

PR #922, opened 2023-07-14
Allows the release namespace to be overridden (#909)

amir-badar:
#909

### DIFF
--- a/templates/injector-network-policy.yaml
+++ b/templates/injector-network-policy.yaml
@@ -10,6 +10,7 @@ apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: {{ template "vault.fullname" . }}-agent-injector
+  namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/name: {{ template "vault.name" . }}-agent-injector
     app.kubernetes.io/instance: {{ .Release.Name }}


hsimon-hashicorp:
Hello @amir-badar - thank you for this submission! Can you let me know, was this a miss from the linked PR? Thanks!

amir-badar:
> 

@hsimon-hashicorp Is there a PR already raised for this issue? I can't see if you have put a reference link for that PR. 

The problem which I have found is when you deploy this helm chart in a custom namespace it throws an error about Network Policy being missing for Namespace and as I have dig down i found that this object is not namespace scoped and so it throwing InvalidSpecError.

dlydiard:
To give more context we are installing this chart with kustomize (helm enabled) + argo, so the namespace is required on the NetworkPolicy CR. The other injector charts seem to include the namespace, but it was missing for `NetworkPolicy`

amir-badar:
@hsimon-hashicorp Do you have an ETA as to when this will be approved?

templates/injector-network-policy.yaml
```diff
@@ -10,6 +10,7 @@ apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: {{ template "vault.fullname" . }}-agent-injector
+  namespace: {{ .Release.Namespace }}
```
tomhjp, line 13: To ensure the namespace can be overridden like the other changes from #909, I think it makes sense to use the template here:

```suggestion
  namespace: {{ include "vault.namespace" . }}
```

tomhjp (commented):
Thanks for raising this, in addition to my suggested edit, please can we get a bats unit test for this to validate? It doesn't look like we have any unit tests for the network policy yet, so you'll need to add a new file in `test/unit`

Issue #917, opened 2023-06-22
Service account enhancements needed for Kunbernetes v1.24


leathem1:
**Is your feature request related to a problem? Please describe.**
With the new version of kubernetes (v1.24) the previous behavior of service accounts automatically getting created with secrets and tokens in that secret that dont expire has changed.  Can you update these vault templates to now create those for us?

**Describe the solution you'd like**
Have this helm chart create the long lived API tokens for Service accounts (https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-long-lived-api-token-for-a-serviceaccount)  since Kubernetes no longer does this for us.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

tvoran:
Added in https://github.com/hashicorp/vault-helm/pull/923

PR #923, opened 2023-07-18
Adding long lived SA token option (#917)

gillcaleb:
Addresses #917

The intent of this PR is to allow for the [creation of a long-lived API token for the server SA](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#create-token).

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: vault
-version: 0.25.1
+version: 0.25.0
 appVersion: 1.14.0
 kubeVersion: ">= 1.20.0-0"
 description: Official HashiCorp Vault Chart
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: vault
-version: 0.25.0
+version: 0.25.1
 appVersion: 1.14.0
 kubeVersion: ">= 1.20.0-0"
 description: Official HashiCorp Vault Chart
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -89,7 +89,7 @@ Compute if the server serviceaccount should have a token created and mounted to 
 {{- $_ := set . "serverServiceAccountSecretCreationEnabled"
   (and
     (eq (.Values.server.serviceAccount.create | toString) "true")
-    (eq (.Values.server.serviceAccount.generateSecret | toString) "true")) -}}
+    (eq (.Values.server.serviceAccount.createSecret | toString) "true")) -}}
 {{- end -}}
 
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -85,11 +85,11 @@ Compute if the server serviceaccount is enabled.
 {{/*
 Compute if the server serviceaccount should have a token created and mounted to the serviceaccount.
 */}}
-{{- define "vault.serverServiceAccountTokenCreationEnabled" -}}
-{{- $_ := set . "serverServiceAccountTokenCreationEnabled"
+{{- define "vault.serverServiceAccountSecretCreationEnabled" -}}
+{{- $_ := set . "serverServiceAccountSecretCreationEnabled"
   (and
     (eq (.Values.server.serviceAccount.create | toString) "true")
-    (eq (.Values.server.serviceAccount.generateToken | toString) "true")) -}}
+    (eq (.Values.server.serviceAccount.generateSecret | toString) "true")) -}}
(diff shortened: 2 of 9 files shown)


gillcaleb:
It's unclear to my why the license/cla check continues to remain in a pending state as I've gone through the signing process and verified its completion on my end. I've tried re-checking it numerous times but to no avail.

stavvy-cgill:
> > It's unclear to my why the license/cla check continues to remain in a pending state as I've gone through the signing process and verified its completion on my end. I've tried re-checking it numerous times but to no avail.
> 
> Usually that happens if your contributions are under a different account, ie, email. For example, I see that @stavvy-cgill is making the comments, but @gillcaleb is the commenter here. Try signing with your @stavvy-cgill account and that should fix it. Thanks!

Thank you!

gillcaleb:
Hi team! Just wanted to check back in to get a sense for when this might get a review. Thanks!

Chart.yaml
```diff
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: vault
-version: 0.25.0
+version: 0.25.1
```
tomhjp, line 6: Please can you revert this? We make the call on what the next version will be at release time, so no need to edit it in the meantime. Thanks!

values.yaml
```diff
@@ -917,6 +917,9 @@ server:
     # The name of the service account to use.
     # If not set and create is true, a name is generated using the fullname template
     name: ""
+    # Controls whether or not a long lived API token should be created for this Service account
+    # serviceAccount.create must be equal to 'true' in order to use this feature. 
+    generateToken: false
```
tomhjp, line 922: I'd like us to be very explicit that this is not recommended:

```suggestion
    # Create a Secret API object to store a non-expiring token for the service account.
    # Prior to v1.24.0, Kubernetes used to generate this secret for each service account by default.
    # Kubernetes now recommends using short-lived tokens from the TokenRequest API or projected volumes instead if possible.
    # For more details, see https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets
    # serviceAccount.create must be equal to 'true' in order to use this feature.
    createSecret: false
```

I've also suggested a slightly different name, but I won't pepper the accompanying suggestions through the rest of the PR to avoid spam.

templates/server-serviceaccount-token.yaml
```diff
@@ -0,0 +1,21 @@
+{{/*
```
tomhjp, line 1: Please can we rename the file to `server-serviceaccount-secret.yaml`? The `metadata.name` field makes sense to stick with -token, but I think we're fairly consistent about using the API object name in the file name.

tomhjp (commented):
Thanks for raising this - I'm ok with adding it but I want to make sure Kubernetes' alternative recommendations are made clear. LGTM once a few naming and documentation comments are addressed. Thanks for including the tests and schema update too!

gillcaleb:
> Thanks for raising this - I'm ok with adding it but I want to make sure Kubernetes' alternative recommendations are made clear. LGTM once a few naming and documentation comments are addressed. Thanks for including the tests and schema update too!

Appreciate the feedback! I went ahead and added the suggested changes.

kaibepperling:
Thanks for opening the PR @gillcaleb! We ran in to the same thing today and were about to open a PR. 
Is anything blocking a release of that change?

templates/_helpers.tpl
```diff
@@ -82,6 +82,17 @@ Compute if the server serviceaccount is enabled.
...
 {{- end -}}
 
+{{/*
+Compute if the server serviceaccount should have a token created and mounted to the serviceaccount.
+*/}}
+{{- define "vault.serverServiceAccountSecretCreationEnabled" -}}
+{{- $_ := set . "serverServiceAccountSecretCreationEnabled"
+  (and
+    (eq (.Values.server.serviceAccount.create | toString) "true")
+    (eq (.Values.server.serviceAccount.generateSecret | toString) "true")) -}}
```
tomhjp, line 92: To match the values.yaml file, here and elsewhere:

```suggestion
    (eq (.Values.server.serviceAccount.createSecret | toString) "true")) -}}
```

tomhjp (commented):
Sorry for the delay revisiting this - and thanks for addressing the comments. The changes look good to me, except we just need a very mechanical replacement of `generateSecret` with `createSecret` in a few places to match up with the latest name in values.yaml.

tomhjp (approved):
LGTM, thanks! I'll wait for the acceptance tests to pass and then merge.

PR #928, opened 2023-07-26
chore: bump vault from 1.14.0 -> 1.14.1

venkatamutyala:
Updating to the latest tag 1.14.1

https://github.com/hashicorp/vault/releases

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -3,8 +3,8 @@
 
 apiVersion: v2
 name: vault
-version: 0.25.0
-appVersion: 1.14.0
+version: 0.25.1
+appVersion: 1.14.1
 kubeVersion: ">= 1.20.0-0"
 description: Official HashiCorp Vault Chart
 home: https://www.vaultproject.io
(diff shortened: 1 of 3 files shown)


EvertonSA:
please merge I really need this fix on my enterprise instance

venkatamutyala:
@jasonodonnell @tvoran @mitchellh @catsby apologies for the spam but this PR has been sitting for a few days now. Any chance we could get this merged in and a release cut/published?

tomhjp:
Thanks for your interest - we have an internally documented release process that encompasses this + a bunch of other steps, so I'm going to close this PR for now. I'll try to get a release done pretty soon, but note that we strongly recommend pinning your own tag when deploying the helm chart to production, so I would not expect a helm chart release to be a blocker for upgrading to 1.14.1: https://developer.hashicorp.com/vault/docs/platform/k8s/helm/configuration#tag-2

Just to reassure as well, there are no compatibility issues we are aware of between the latest helm chart and the latest Vault version - when we do do a release, the only accommodation we'll make for upgrading to 1.14.1 is simply updating the default tag.

PR #930, opened 2023-07-27
fix(test): add securityContext for Helm test

fty4:
When running tests with Helm the same securityContext also should apply.
Therefore this change will add the securityContext from the values file also to the Helm tests.

The change will use the securityContext of the server statefulSet.
This results from the fact that the same image is used in the test.

<hr />

Marco Lecheler [marco.lecheler@mercedes-benz.com](mailto:marco.lecheler@mercedes-benz.com) Mercedes-Benz Tech Innovation GmbH ([ProviderInformation](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md))

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -563,6 +563,44 @@ securityContext for the statefulset vault container
   {{- end }}
 {{- end -}}
 
+{{/*
+securityContext for the test pod template.
+*/}}
+{{- define "server.test.securityContext.pod" -}}
+  {{- if .Values.server.statefulSet.securityContext.pod }}
+  securityContext:
+    {{- $tp := typeOf .Values.server.statefulSet.securityContext.pod }}
+    {{- if eq $tp "string" }}
+      {{- tpl .Values.server.statefulSet.securityContext.pod . | nindent 4 }}
+    {{- else }}
+      {{- toYaml .Values.server.statefulSet.securityContext.pod | nindent 4 }}
+    {{- end }}
+  {{- else if not .Values.global.openshift }}
+  securityContext:
+    runAsNonRoot: true
+    runAsGroup: {{ .Values.server.gid | default 1000 }}
+    runAsUser: {{ .Values.server.uid | default 100 }}
+    fsGroup: {{ .Values.server.gid | default 1000 }}
+  {{- end }}
+{{- end -}}
+
+{{/*
+securityContext for the test vault container
+*/}}
+{{- define "server.test.securityContext.container" -}}
+  {{- if .Values.server.statefulSet.securityContext.container }}
+      securityContext:
+        {{- $tp := typeOf .Values.server.statefulSet.securityContext.container }}
+        {{- if eq $tp "string" }}
+          {{- tpl .Values.server.statefulSet.securityContext.container . | nindent 8 }}
+        {{- else }}
+          {{- toYaml .Values.server.statefulSet.securityContext.container | nindent 8 }}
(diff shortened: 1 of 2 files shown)


templates/_helpers.tpl
```diff
@@ -563,6 +563,44 @@ securityContext for the statefulset vault container
   {{- end }}
 {{- end -}}
 
+{{/*
+securityContext for the test pod template.
+*/}}
+{{- define "server.test.securityContext.pod" -}}
```
KhizerJaan, line 569: Hello @fty4 , why do we need to define new template , as its supposed to use the securityContext of the server statefulSet. why not use "server.statefulSet.securityContext.pod" for server-test pod?

fty4: Hi @KhizerJaan 
I did that because of the indentation of the template.
But now I just added a new commit to use `nindent` to fix the indentation in the template files.

I hope this is the desired solution.

KhizerJaan: @fty4  Thank you , but can you please check indentation , it does not look right for both server-statefulset and and server-test .

fty4: When I run a diff between the templated yaml from main and after the fix I get this change:

```diff
381a382
>       
414a416
>           
557a560,565
>   
>   securityContext:
>     runAsNonRoot: true
>     runAsGroup: 1000
>     runAsUser: 100
>     fsGroup: 1000
585a594,596
>       
>       securityContext:
>         allowPrivilegeEscalation: false
```

This means the indentation should be fine.
The only issue is see here is that two new lines are beeing added...
But this is a issue in the whole chart and I am not sure how to fix this.

I guess you mean the indentation in the helpers file - I was required to remove the indentation here to later use different indentation on the statefulset and the test.
This is because both use different indentation length ([6](https://github.com/mercedes-benz/vault-helm/blob/fix/test-securitycontext/templates/server-statefulset.yaml#L56) and [10](https://github.com/mercedes-benz/vault-helm/blob/fix/test-securitycontext/templates/server-statefulset.yaml#L77) for the statefulset and [2](https://github.com/mercedes-benz/vault-helm/blob/fix/test-securitycontext/templates/tests/server-test.yaml#L18) and [6](https://github.com/mercedes-benz/vault-helm/blob/fix/test-securitycontext/templates/tests/server-test.yaml#L47) for the server-test).

Do you agree, @KhizerJaan?

KhizerJaan: [this line](https://github.com/hashicorp/vault-helm/pull/930/files#diff-87d68c754766af8e2e930e653be7e4b75fa0c8bdb187cb1bec293f265d9159ffR535)
&
[this](https://github.com/hashicorp/vault-helm/pull/930/files#diff-0ef176392f0823ab2ed4f8ab706fd9bbd322143d85c35831a4c775136da7718aR56)
would add to indent 14 and same is with others.

fty4: Actually this does not result in a diff in the templated resources - but for sure it is better to remove them when the indentation is set in another place.

I've added commit 43ca102 for this.

tomhjp (commented):
This looks good to me, and it seems sensible to reuse the server's security context for the same image. Please could we just add some unit tests to have some enforced validation on the output?

fty4:
I updated the requested changes - also I noticed that a unit-test failed (wondering why this test will not be triggered in a PR).
There was an issue with the line break when specifying separate securityContext values.

I hope the tests implemented fit your needs.

PR #931, opened 2023-07-28
feat(Prometheus/ServiceMonitor): add configurable tlsConfig & bearerTokenFile for authentication

eyenx:
This implements a configurable `tlsConfig` for the `serviceMonitor` used to grab metrics from the Vault API.

Additionally it adds the possibility to set a `bearerTokenFile` for authentication purposes.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: vault
-version: 0.25.0
+version: 0.26.0
 appVersion: 1.14.0
 kubeVersion: ">= 1.20.0-0"
 description: Official HashiCorp Vault Chart
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: vault
-version: 0.28.0
+version: 0.27.0
 appVersion: 1.15.2
 kubeVersion: ">= 1.20.0-0"
 description: Official HashiCorp Vault Chart
--- a/test/unit/prometheus-servicemonitor.bats
+++ b/test/unit/prometheus-servicemonitor.bats
@@ -123,3 +123,45 @@ load _helpers
   [ "$(echo "$output" | yq -r '.spec.endpoints | length')" = "1" ]
   [ "$(echo "$output" | yq -r '.spec.endpoints[0].port')" = "https" ]
 }
+
+@test "prometheus/ServiceMonitor-server: tlsConfig default" {
+  cd `chart_dir`
+  local output=$( (helm template \
+      --show-only templates/prometheus-servicemonitor.yaml \
+      --set 'serverTelemetry.serviceMonitor.enabled=true' \
+      . ) | tee /dev/stderr)
+
+  [ "$(echo "$output" | yq -r '.spec.endpoints[0].tlsConfig.insecureSkipVerify')" = "true" ]
+}
+
+@test "prometheus/ServiceMonitor-server: tlsConfig override" {
+  cd `chart_dir`
+  local output=$( (helm template \
+      --show-only templates/prometheus-servicemonitor.yaml \
+      --set 'serverTelemetry.serviceMonitor.tlsConfig.ca=ca.crt' \
+      --set 'serverTelemetry.serviceMonitor.enabled=true' \
+      . ) | tee /dev/stderr)
+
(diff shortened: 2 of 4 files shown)


eyenx:
@tvoran @swenson @tomhjp 

Anyone able to review this? It's kinda sad this is laying around without reaction for more than 3 months.

dtrts:
I would also like to see this feature added to the chart. 

If possible the chance to configure `bearerTokenSecret` would also be welcomed.

eyenx:
> I would also like to see this feature added to the chart.
> 
> If possible the chance to configure `bearerTokenSecret` would also be welcomed.

Would be nice I know, but HashiCorp does not care about contributions from outsiders. This will be laying around for-ever or they will implement it by themselves.

Chart.yaml
```diff
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: vault
-version: 0.27.0
+version: 0.28.0
```
tomhjp, line 6: We update this separately to individual feature PRs

```suggestion
version: 0.27.0
```

test/unit/prometheus-servicemonitor.bats
```diff
@@ -123,3 +123,45 @@ load _helpers
...
+}
+
+@test "prometheus/ServiceMonitor-server: bearerTokenFile default" {
+  cd `chart_dir`
+  local output=$( (helm template \
+      --show-only templates/prometheus-servicemonitor.yaml \
+      --set 'serverTelemetry.serviceMonitor.enabled=true' \
+      . ) | tee /dev/stderr)
+
+  [ "$(echo "$output" | yq -r '.spec.endpoints[0] | has("bearerToken")')" = "false" ]
```
tomhjp, line 155:
```suggestion
  [ "$(echo "$output" | yq -r '.spec.endpoints[0] | has("bearerTokenFile")')" = "false" ]
```

values.yaml
```diff
@@ -1277,6 +1277,12 @@ serverTelemetry:
     # Timeout for Prometheus scrapes
     scrapeTimeout: 10s
 
+    # tlsConfig used for connecting to the Vault API
+    tlsConfig: {}
+
+    # bearerTokenfile used for authentication to the Vault metrics API
+    bearerTokenFile: ""
```
tomhjp, line 1284: There is a comment up above that needs a bit of updating:

>  ```
> # Enable support for the Prometheus Operator. Currently, this chart does not support
>  # authenticating to Vault's metrics endpoint, so the following `telemetry{}` must be included
>  # in the `listener "tcp"{}` stanza
>  #  telemetry {
>  #    unauthenticated_metrics_access = "true"
>  #  }
> ```

Maybe something along the lines of:

>  ```
> # Enable support for the Prometheus Operator. If bearerTokenFile is not set for authenticating
>  # to Vault's metrics endpoint, the following Vault server `telemetry{}` config must be included
>  # in the `listener "tcp"{}` stanza
>  #  telemetry {
>  #    unauthenticated_metrics_access = "true"
>  #  }
> ```

tomhjp (commented):
Thanks for the PR - I appreciate it's been stuck without feedback for a long time, but if you're still interested I'm happy to help you get this landed. I've triggered the test suite and have a few minor comments but this looks pretty close to merging.

eyenx:
Thank you @tomhjp will look at it today

values.yaml
eyenx: 011f82c

eyenx: wait my lazy-vim did a lot of changes, gonna revert them.

eyenx: There we go 71d6325

eyenx:
@tomhjp ready for a review :)

tomhjp:
Hey @eyenx, thanks for your patience. I wanted to try writing a proper integration test that exercised these options for real, I've pushed some WIP in https://github.com/hashicorp/vault-helm/commit/d9ee3dcdf548f6b7e60e714e62c7119c903f0fc8.

One thing I learnt while I was looking deeper is that the `bearerTokenFile` option is deprecated, with users directed to `authorization` instead: https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.Endpoint. It does seem like a safer and improved pattern, WDYT about supporting `authorization` instead of `bearerTokenFile` in this PR?

tomhjp:
Merged in #1025 - thanks again!

PR #935, opened 2023-08-01
docs: Update outdated `vaultproject.io/docs/` links

tanmay-pnaik:
Update outdated https://www.vaultproject.io/docs/ links that are now redirects to https://developer.hashicorp.com/vault/docs


tvoran (approved):
This looks great, thanks so much!

Issue #918, opened 2023-06-29
Ability to scale replicas to zero


brettcurtis:
We run multiple environments and would like to be able to simply set the replicas to "0" to reduce cost on testing environments when we are not actively using them.

Do you guys feel it's resonable to be able to do this in the chart of should I look into ways do do it outside the chart?

Thanks!

therealjsie:
I've created a PR to change this behavior here: https://github.com/hashicorp/vault-helm/pull/943

kschoche:
Fixed by #943

PR #943, opened 2023-08-14
Allow scale to zero

therealjsie:


therealjsie:
Fixes #918

templates/_helpers.tpl
```diff
@@ -156,7 +156,11 @@ Set's the replica count based on the different modes configured by user
   {{ if eq .mode "standalone" }}
     {{- default 1 -}}
   {{ else if eq .mode "ha" }}
-    {{- .Values.server.ha.replicas | default 3 -}}
+    {{- if kindIs "invalid" .Values.server.ha.replicas -}}
```
tomhjp, line 159: TIL about `kindIs`. I couldn't find any documentation for "invalid" though. Do you have any references you could share? I'm wondering if it might be better to switch the logic though, something like:

```
{{- if kindIs "int" .Values.server.ha.replicas -}}
  {{- .Values.server.ha.replicas -}}
{{- else -}}
  {{- 3 -}}
{{- end -}}
```

Although I haven't thought very deeply about it yet. Do you think that would also work?

test/unit/server-ha-statefulset.bats
```diff
@@ -157,6 +157,28 @@ load _helpers
...
+      yq -r '.spec.replicas' | tee /dev/stderr)
+  [ "${actual}" = "0" ]
+}
+
+@test "server/ha-StatefulSet: custom replicas" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-statefulset.yaml  \
+      --set 'server.ha.enabled=true' \
+      --set 'server.ha.replicas=null' \
```
tomhjp, line 176: Nice tests! I'm 99% sure it will work, but just out of paranoia could we also add a test case where we don't set `server.ha.replicas` at all?

tomhjp (commented):
Thanks for opening this - it seems like an elegant solution while maintaining backwards compatibility. Just one question mark for me around whether we can safely rely on "invalid" as documented behaviour?

templates/_helpers.tpl
therealjsie: The source code here just uses Go's `reflect` lib: https://github.com/Masterminds/sprig/blob/master/reflect.go 

I believe both ways will work, but I like your approach better as it seems more intuitive. I'll adapt my PR tomorrow.

test/unit/server-ha-statefulset.bats
therealjsie: I'm not sure how we'd go about that, as `server.ha.replicas` is always set through the default `values.yaml` file. There is already a test case in the `server-ha-statefulset.bats` file for _no explicit value_, so I'm not sure what else to cover with the unit test. We have the following options now:

- No explicit value -> 3
- Arbitrary, non zero value -> 10
- Zero -> 0
- `null` -> 3

What other case would you add?

tomhjp: Oh my bad, I forgot it's set in the `values.yaml` file. Weird that the templating code previously used ` | default 3` in that case 🤔 but feel free to ignore this comment!

templates/_helpers.tpl
therealjsie: Based on [this](https://github.com/helm/helm/issues/12084) and some manual fiddling, I changed it to check for `float64`.

tomhjp: Thanks for doing the research, that looks good to me! 👍

tomhjp: On the tests, are you referring to the vault-helm-test image? We don't actually host that internally, we just build it on demand and it's documented here: https://github.com/hashicorp/vault-helm/blob/main/CONTRIBUTING.md#test-using-docker-container. But that is pretty easy to miss, so apologies if that caused confusion.

tomhjp (approved):
Thanks for contributing this! LGTM. I've triggered the CI tests and I'll merge once they pass (which I'm certain they will as the unit tests have already passed).

therealjsie:
@tomhjp Thanks for merging! Out of curiosity, is there a regular release cadence, or when will a new release of the helm chart be triggered?

swenson:
@therealjsie we typically aim to release every month

xdoncalvox:
After this change my helm release which uses server.ha.replicas set to 2, defaulted to 3. I am wondering if this might be caused by to the value type "int64" specified in the if statement, does not matter what numeric number I set ha to, the operator seems to ignore the value and default to 3.

My values.yaml:
`
global:
  enabled: true
  tlsDisabled: true
  
injector:
  enabled: true
        
server:
  ha:
    enabled: true
    replicas: 2
`

PR #955, opened 2023-09-14
feat: add hostAliases for statefulset

eufebius:
Add hostAliases to the helm chart to allow adding custom entries to /etc/hosts

Useful when configuring LDAP sign method.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -61,6 +61,10 @@ spec:
         {{ template "vault.volumes" . }}
         - name: home
           emptyDir: {}
+      {{- if .Values.server.hostAliases }}
+      hostAliases:
+        {{ toYaml .Values.server.hostAliases | nindent 8}}
+      {{- end }}          
       {{- if .Values.server.extraInitContainers }}
       initContainers:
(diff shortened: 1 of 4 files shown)


templates/server-statefulset.yaml
```diff
@@ -61,6 +61,10 @@ spec:
         {{ template "vault.volumes" . }}
         - name: home
           emptyDir: {}
+      {{- if .Values.server.hostAliases }}
+      hostAliases:
+        {{ toYaml .Values.server.hostAliases | nindent 8}}
+      {{- end }}
```
tvoran, line 67: Just some extra whitespace here:

```suggestion
      {{- end }}
```

values.yaml
```diff
@@ -444,7 +444,11 @@ server:
     #  - secretName: chart-example-tls
     #    hosts:
     #      - chart-example.local
-
+  # hostAliases is a list of aliases to be added to /etc/hosts. Specified as a YAML list.
+  hostAliases: []
+  # - ip: 127.0.0.1
+  #   hostnames:
+  #     - chart-example.local
```
tvoran, line 451:
```suggestion
  #     - chart-example.local

```

tvoran (approved):
Thanks for picking this up!

PR #966, opened 2023-10-24
openshift/server: readiness probe passes when server uninitialized

### LINKED ISSUE #954: Chart verification failing due to vault being uninitialized/sealed
**Describe the bug**

This is related to the Red Hat chart certification process (https://github.com/openshift-helm-charts/charts), I help maintain the certification workflows. 

A recent feature went into [chart-verifier](https://github.com/redhat-certification/chart-verifier/commit/ab27c64b4603dfa597e6ee40926c47522c3a0aa5) allowing it to monitor the readiness of statefulsets and daemonsets. Historically, it has only checked for Deployment readiness.

With this change, the vault chart fails to complete execution because the vault server stateful set is uninitialized and sealed by default.

Looking at the server test you provide in-chart, I see that you just check to see that you can poll the API and get the seal status, but unfortunately that never runs because we always wait to execute the test until the workloads are ready.

By default the `server.readinessProbe.path` value from values.yaml is commented, meaning a shell exec is executed against the server pod to determine readiness. That's what's failing.

https://github.com/hashicorp/vault-helm/blob/main/templates/server-statefulset.yaml#L155-L169

Your documentation indicates that the values.yaml can be updated to contain a path that returns HTTP 204 if sealed/uninitialized:

https://developer.hashicorp.com/vault/docs/platform/k8s/helm/run#probes

If this was set in the values.yaml file for your chart, it seems like this should work for certification. Would this be something you would consider setting as a part of the default values.yaml?

CC-ing the owners listed in the certified chart entry:

@tvoran
@jasonodonnell
@tomhjp
@calvn
@swenson

**To Reproduce**
1. Install chart-verifier's development version https://github.com/redhat-certification/chart-verifier/releases/tag/0.1.0 (we have not cut a release that contains this. Our nightly workflows caught this bug
2. Run it against  the current chart you have
3. Observe the "permanent hang", (or more specifically, a timeout at 30 minutes. If you want, set the timeout value to a shorter value to see the failure)

**Expected behavior**
The statefulset to become ready so that we can continue with executing chart-testing.

**Environment**
OpenShift (any recent version should work)

Chart values:
The default included in your chart repo.

**Additional context**

I do see a values.openshift.yaml in your repo. The problem with using that overlayed over your default values.yaml is that our certification workflows don't take that into account, unless Helm automatically reads in all values.*.yaml files (I don't think it does). When you folks submit your chart to us, you're submit just the chart, and our automation runs chart-verifier in a fixed way on your behalf to generate a result and check for a passing status. Adding the path values that is tolerant to uninitialized states by default in just the values.yaml simplifies this greatly.

tvoran:
Changes the default server readiness probe (just for openshift) to pass when the server is uninitialized, in order to pass the latest version of the chart-verifier test (see #954 for details).

Also updates the chart-verifier used in our tests to 1.13.0 (latest).

Testing against an OpenShift cluster:

```shell-session
$ go install github.com/redhat-certification/chart-verifier@1.13.0

$ chart-verifier verify ~/projects/vault-helm-review \
  --chart-values ~/projects/vault-helm-review/values.yaml \
  --chart-values ~/projects/vault-helm-review/values.openshift.yaml \
  --timeout 30s -e chart-testing
...
results:
    - check: v1.0/chart-testing
      type: Mandatory
      outcome: PASS
      reason: Chart tests have passed
```

Fixes #954

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -10,14 +10,14 @@ jobs:
   chart-verifier:
     runs-on: ubuntu-latest
     env:
-      CHART_VERIFIER_VERSION: '1.10.1'
+      CHART_VERIFIER_VERSION: '1.13.0'
     steps:
       - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
       - name: Setup test tools
         uses: ./.github/actions/setup-test-tools
       - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
         with:
-          go-version: '1.19.2'
+          go-version: '1.21.3'
       - run: go install "github.com/redhat-certification/chart-verifier@${CHART_VERIFIER_VERSION}"
       - run: bats --tap --timing ./test/chart
 permissions:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changes:
 * Default `vault` version updated to 1.15.0
 * Default `vault-k8s` version updated to 1.3.0
 * Tested with Kubernetes versions 1.24-1.28
+* server: OpenShift default readiness probe returns 204 when uninitialized
 
 Features:
 * server: Add support for dual stack clusters [GH-833](https://github.com/hashicorp/vault-helm/pull/833)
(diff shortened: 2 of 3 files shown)


swenson (approved):
LGTM

tvoran:
Thanks!

PR #969, opened 2023-10-30
Add labels for PVC template

alantang888:
Some volume snapshot automation tools use label selector on PVC.
Unit test all are ok.


tvoran (approved):
Looks great, thanks!

PR #971, opened 2023-11-02
support exec in server liveness probe

thyton:
This PR supports exec in the server liveness probe. If provided, exec is preferred to httpGet as the server liveness probe handler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased
 
-IMPROVEMENTS:
+Improvements:
 
 * Support exec in the server liveness probe [GH-971](https://github.com/hashicorp/vault-helm/pull/971)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+IMPROVEMENTS:
+
+* Support exec in the server liveness probe [GH-971](https://github.com/hashicorp/vault-helm/pull/971)
+
 ## 0.26.1 (October 30, 2023)
 
 Bugs:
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+IMPROVEMENTS:
```
tvoran, line 3: We can probably lowercase this to follow the rest of the file.

```suggestion
Improvements:
```

values.yaml
```diff
@@ -544,6 +544,11 @@ server:
     successThreshold: 1
     # Number of seconds after which the probe times out.
     timeoutSeconds: 3
+    # Used to define a liveness command
+    execCommand: []
+    # - /bin/sh
+    # - -c
+    # - /vault/userconfig/mylivenessscript/run.sh
```
tvoran, line 551: I'd suggest moving this up by `path`, and mention the preference order in the comment.

tvoran (commented):
This is looking great!

It's also good to include a description in a PR like this, even if it's short.

values.yaml
thyton: Thank you. Will do!

values.yaml
```diff
@@ -534,6 +534,11 @@ server:
     path: "/v1/sys/health?standbyok=true"
```
tvoran, line 534: It would be great to add a comment for `path` too, now that we also have `exec`.

tvoran (commented):
Could you also add an entry for `execCommand` in values.schema.json? We don't have a tool to help generate it anymore, so it needs to be done by hand.

values.yaml
thyton: Will update it. Thank you!

tvoran (approved):
Just a couple thoughts on wording. This tested great! 👍

thyton:
Thanks a lot for your feedback!

PR #1001, opened 2024-03-07
Add annotation on config change

### LINKED ISSUE #748: RFE: Ability to restart vault on configmap change

When I change the configuration, such as to add "plugin_directory" to the config, it does not "notify" the pods to restart.

A common way to handle that is to use a sha256sum of contents of the configmap as an annotation to the pod. This allows for a simple "notification" that the pods need to restart.

Some sort of post-upgrade hook?

**Additional context**

https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

swenson:
When updating the Vault config (and corresponding) configmap, we now generate a checksum of the config and set it as an annotation on both the configmap
and the Vault StatefulSet pod template.

This allows the deployer to know what pods need to be restarted to pick up the a changed config.

We still recommend using the standard upgrade
[method for Vault on Kubernetes](https://developer.hashicorp.com/vault/tutorials/kubernetes/kubernetes-raft-deployment-guide#upgrading-vault-on-kubernetes), i.e., using the `OnDelete` strategy
for the Vault StatefulSet, so updating the config
and doing a `helm upgrade` should not trigger the
pods to restart, and then deleting pods one
at a time, starting with the standby pods.

With `kubectl` and `jq`, you can check check which pods need to be updated by first getting the value of the current configmap checksum:

```shell
kubectl get pods -o json | jq -r ".items[] | select(.metadata.annotations.\"config/checksum\" != $(kubectl get configmap vault-config -o json | jq '.metadata.annotations."config/checksum"') ) | .metadata.name"
```

Fixes #748.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -459,7 +459,7 @@ Sets extra pod annotations
 */}}
 {{- define "vault.annotations" }}
       annotations:
-  {{- if .Values.server.configAnnotation }}
+  {{- if .Values.server.includeConfigAnnotation }}
         vault.hashicorp.com/config-checksum: {{ include "vault.config" . | sha256sum }}
   {{- end }}
   {{- if .Values.server.annotations }}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -460,7 +460,7 @@ Sets extra pod annotations
 {{- define "vault.annotations" }}
       annotations:
   {{- if .Values.server.configAnnotation }}
-        vaultproject.io/config-checksum: {{ include "vault.config" . | sha256sum }}
+        vault.hashicorp.com/config-checksum: {{ include "vault.config" . | sha256sum }}
   {{- end }}
   {{- if .Values.server.annotations }}
         {{- $tp := typeOf .Values.server.annotations }}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -459,7 +459,9 @@ Sets extra pod annotations
 */}}
 {{- define "vault.annotations" }}
       annotations:
-        config/checksum: {{ include "vault.config" . | sha256sum }}
+  {{- if .Values.server.configAnnotation }}
+        vaultproject.io/config-checksum: {{ include "vault.config" . | sha256sum }}
+  {{- end }}
   {{- if .Values.server.annotations }}
         {{- $tp := typeOf .Values.server.annotations }}
         {{- if eq $tp "string" }}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -457,9 +457,10 @@ Sets the injector deployment update strategy
 {{/*
 Sets extra pod annotations
 */}}
-{{- define "vault.annotations" -}}
-  {{- if .Values.server.annotations }}
+{{- define "vault.annotations" }}
       annotations:
+        config/checksum: {{ include "vault.config" . | sha256sum }}
+  {{- if .Values.server.annotations }}
         {{- $tp := typeOf .Values.server.annotations }}
         {{- if eq $tp "string" }}
           {{- tpl .Values.server.annotations . | nindent 8 }}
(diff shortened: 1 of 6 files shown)


templates/_helpers.tpl
```diff
@@ -457,9 +457,10 @@ Sets the injector deployment update strategy
 {{/*
 Sets extra pod annotations
 */}}
-{{- define "vault.annotations" -}}
-  {{- if .Values.server.annotations }}
+{{- define "vault.annotations" }}
       annotations:
+        config/checksum: {{ include "vault.config" . | sha256sum }}
```
benashz, line 462: I think we probably want add a configurable that controls inclusion/exclusion of this annotation, since changing a configuration could cause an unexpected restart of Vault. Users may currently have their own strategy for dealing Vault config updates.

benashz (commented):
I like the approach. I think we definitely want make including the dynamic annotation optional. Also we will probably want to mention something about setting up the the `strategy` to control how best to handle the update whenever the configuration changes.

In addition, we should also extend the bats tests to cover the new features being added here.

benashz:
We may also want to prefix the annotation something like `vault.` or something like that, to avoid a collision `values...annotations`

swenson:
Updated. I also fixed the `Test.dockerfile` to work again (the way we were using `pip` to install `yq` no longer worked).

templates/_helpers.tpl
```diff
@@ -457,9 +457,12 @@ Sets the injector deployment update strategy
 {{/*
 Sets extra pod annotations
 */}}
-{{- define "vault.annotations" -}}
-  {{- if .Values.server.annotations }}
+{{- define "vault.annotations" }}
       annotations:
+  {{- if .Values.server.configAnnotation }}
+        vaultproject.io/config-checksum: {{ include "vault.config" . | sha256sum }}
```
tomhjp, line 463: For a recent k8s auth plugin [feature](https://github.com/hashicorp/vault-plugin-auth-kubernetes/commit/daeaf8df3497b60b1d0804b5c0d396c462e46b2e), we went with `vault.hashicorp.com/` as the annotation prefix, and we should probably match that here. I believe @thyton checked a few other examples for consistency before landing on that, cc'ed to keep me honest :)

```suggestion
        vault.hashicorp.com/config-checksum: {{ include "vault.config" . | sha256sum }}
```

tomhjp (approved):
LGTM 👍 just a nit on the annotation prefix

test/unit/server-statefulset.bats
```diff
@@ -1681,6 +1681,25 @@ load _helpers
...
+}
+
+@test "server/standalone-StatefulSet: config checksum annotation can be enabled" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-statefulset.yaml \
+      --set 'server.configAnnotation=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.metadata.annotations["vault.hashicorp.com/config-checksum"] == null' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
```
benashz, line 1700: Should we also test the value here? Maybe assert something about its length, or expect a specific hash?

values.yaml
```diff
@@ -668,6 +668,13 @@ server:
   # of the annotations to apply to the server pods
   annotations: {}
 
+  # Add an annotation to the server configmap and the statefulset pods,
+  # vaultproject.io/config-checksum, that is a hash of the Vault configuration.
+  # This can be used together with an OnDelete deployment strategy to help
+  # identify which pods still need to be deleted during a deployment to pick up
+  # any configuration changes.
+  configAnnotation: false
```
benashz, line 676: The parameter makes me think it takes a value for the annotation. Maybe rename it to something like `includeConfigAnnotation` ?

benashz (commented):
Thanks for making the changes. I have a couple more, minor suggestions to take into consideration.

test/unit/server-statefulset.bats
swenson: I did that initially, but realized that the value might be something noisy and annoying to fix every time we change the config file.

benashz: We are doing those checks as part of the other tests for this kind of feature: https://github.com/hashicorp/vault-helm/blob/04074311f71fd3a9b39f983553b693363f4a4f22/test/unit/prometheus-servicemonitor.bats#L84-L85. 

I think it is worth it to catch any potential regression where we could the expected defaults.

benashz (approved):
I still think it is worth updating the tests to ensure the setting this annotation does not affect the defaults. Other than LGTM!

swenson:
Thanks!

PR #1003, opened 2024-03-11
Test against k8s 1.29

benashz:
Drop k8s 1.24


tvoran (approved):
Should probably use the latest [kind version](https://github.com/hashicorp/vault-helm/pull/1003/files#diff-2b0338db9a045bfa4a108bcd46dd89430617fc02c2047855f20a6e6af0499815R19) if we can too.

PR #1032, opened 2024-06-21
Update vault default image to 1.17.0 in values.yaml for vault-helm chart

patelroshan21:
This PR updates the default image version for `vault` to `1.17.0` in the values.yaml file of the vault-helm chart. This ensures that users will deploy the latest stable version of the vault-agent by default.


tvoran:
Thanks but this has been addressed in the v0.28.1 release (#1036).

PR #1043, opened 2024-07-18
Updated the ConfigMap Annotation Hook

aakash21:
The ConfigMap Annotation has been incorrectly mentioned in the values.yaml file, as https://github.com/hashicorp/vault-helm/blob/main/templates/_helpers.tpl#L462 states that it is expecting the boolean value of `includeConfigAnnotation`
```
  {{- if .Values.server.includeConfigAnnotation }}
        vault.hashicorp.com/config-checksum: {{ include "vault.config" . | sha256sum }}
  {{- end }}
```
Instead in `values.yaml` the hook has been specified as `configAnnotation: false` which is not being used, For Reference: https://github.com/hashicorp/vault-helm/blob/main/values.yaml#L676

After updating this value to true, the checksum annotation could be created in the Stateful Set.
```
annotations:
  vault.hashicorp.com/config-checksum: 3961301403b088XXXXXXXXXXX
```

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -673,7 +673,7 @@ server:
   # This can be used together with an OnDelete deployment strategy to help
   # identify which pods still need to be deleted during a deployment to pick up
   # any configuration changes.
-  configAnnotation: false
+  includeConfigAnnotation: false
 
   # Enables a headless service to be used by the Vault Statefulset
   service:
--- a/values.yaml
+++ b/values.yaml
@@ -671,8 +671,8 @@ server:
   # Add an annotation to the server configmap and the statefulset pods,
   # vaultproject.io/config-checksum, that is a hash of the Vault configuration.
   # This can be used together with an OnDelete deployment strategy to help
-  # identify which pods still need to be deleted during a deployment to pick up
-  # any configuration changes.
+  # identify which pods still need to be deleted during a deployment or with
+  # RollingUpdate on an auto-unsealed vault to pick up any configuration changes.
   includeConfigAnnotation: false
 
   # Enables a headless service to be used by the Vault Statefulset
--- a/values.schema.json
+++ b/values.schema.json
@@ -664,7 +664,7 @@
                         }
                     }
                 },
-                "configAnnotation": {
+                "includeConfigAnnotation": {
                     "type": "boolean"
                 },
                 "dataStorage": {


tvoran:
Good catch, thanks for opening this! Would you mind updating that option in [values.schema.json](https://github.com/hashicorp/vault-helm/blob/636f8da6a614e596fb0af9cb119e3a5816f8ef5a/values.schema.json#L667) as well?

We could also add a bats test to check the default value in [test/unit/server-statefulset.bats](https://github.com/hashicorp/vault-helm/blob/636f8da6a614e596fb0af9cb119e3a5816f8ef5a/test/unit/server-statefulset.bats#L1684); I think we can get the default value from `helm show values .` and parse the output with `yq`. If that's something you'd like to add, let us know! Otherwise we can add it in a separate PR.

aakash21:
@tvoran i have modified the schema.json but in the test/unit/server-statefulset.bats the test case for this hook is already present https://github.com/hashicorp/vault-helm/blob/main/test/unit/server-statefulset.bats#L1702
```@test "server/standalone-StatefulSet: config checksum annotation can be enabled" {
  cd `chart_dir`
      --show-only templates/server-statefulset.yaml \
      --set 'server.includeConfigAnnotation=true' \
      . | tee /dev/stderr |
      yq '.spec.template.metadata.annotations["vault.hashicorp.com/config-checksum"] == null' | tee /dev/stderr)
  [ "${actual}" = "false" ]
}```

File: values.yaml [hunk 1]
```diff
@@ -673,7 +673,7 @@ server:
   # This can be used together with an OnDelete deployment strategy to help
   # identify which pods still need to be deleted during a deployment to pick up
   # any configuration changes.
```
miminar, line 675: Thanks for this contribution. I've just bumped into this as well.

Just a nitpick, because I find the comment misleading as it suggests that it's useful just in combination with OnDelete strategy. I actually find it more useful in combination with the other strategy.

```suggestion
  # This can be used together with an OnDelete deployment strategy to help
  # identify which pods still need to be deleted during a deployment or with
  # RollingUpdate on an auto-unsealed vault to pick up any configuration changes.
```
Since RollingUpdate and OnDelete are the only possible values of updateStrategy, the comment may very well be  removed as well.

aakash21: @miminar Makes sense 👍

tvoran: I think we mention OnDelete specifically since that's the recommended strategy to ensure standby's are updated before the active primary: https://developer.hashicorp.com/vault/tutorials/kubernetes/kubernetes-raft-deployment-guide#upgrading-vault-on-kubernetes

(See also the description on the original PR: https://github.com/hashicorp/vault-helm/pull/1001)

tvoran:
Looks like the parameter in values.yaml was updated as part of #1049, so I'm going to close this for now. Thanks!

PR #1048, opened 2024-07-29
Update CSI log tests

Zlaticanin:
This PR involves updating the CSI log tests to use a new `-log-level` flag.

PR on the CSI provider side: https://github.com/hashicorp/vault-csi-provider/pull/295

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ Bugs:
 
 * Properly handle JSON formatted server config [GH-1049](https://github.com/hashicorp/vault-helm/pull/1049)
 
-
 ## 0.28.1 (July 11, 2024)
 
 Changes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+* Default `vault` version updated to 1.17.3
+* Default `vault-csi-provider` version updated to 1.5.0
+
 ## 0.28.1 (July 11, 2024)
 
 Changes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+Changes:
+
 * Default `vault` version updated to 1.17.3
 * Default `vault-csi-provider` version updated to 1.5.0
 
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -59,8 +59,6 @@ spec:
             - --log-level={{ .Values.csi.logLevel }}
             {{- else if .Values.csi.debug }}
             - --log-level=debug
-            {{- else }}
-            - --log-level=info
             {{- end }}
             {{- if .Values.csi.hmacSecretName }}
             - --hmac-secret-name={{ .Values.csi.hmacSecretName }}
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -55,10 +55,12 @@ spec:
           imagePullPolicy: {{ .Values.csi.image.pullPolicy }}
           args:
             - --endpoint=/provider/vault.sock
-            {{- if .Values.csi.debug }}
+            {{- if .Values.csi.logLevel }}
+            - --log-level={{ .Values.csi.logLevel }}
+            {{- else if .Values.csi.debug }}
             - --log-level=debug
             {{- else }}
-            - --log-level={{ .Values.csi.logLevel }}
+            - --log-level=info
             {{- end }}
             {{- if .Values.csi.hmacSecretName }}
             - --hmac-secret-name={{ .Values.csi.hmacSecretName }}
--- a/templates/csi-daemonset.yaml
+++ b/templates/csi-daemonset.yaml
@@ -55,12 +55,10 @@ spec:
           imagePullPolicy: {{ .Values.csi.image.pullPolicy }}
           args:
             - --endpoint=/provider/vault.sock
-            {{- if .Values.csi.logLevel }}
-            - --log-level={{ .Values.csi.logLevel }}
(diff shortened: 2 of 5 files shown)


test/unit/csi-daemonset.bats
```diff
@@ -177,17 +177,40 @@ load _helpers
...
       --set "csi.debug=true" \
       . | tee /dev/stderr |
       yq -r '.spec.template.spec.containers[0].args[1]' | tee /dev/stderr)
-  [ "${actual}" = "--debug=true" ]
+  [ "${actual}" = "--log-level=debug" ]
 }
 
+@test "csi/daemonset: default log-level" {
+  cd `chart_dir`
+  local actual=$(helm template \
```
benashz, line 194: nit: typically you would declare the local variable before setting it
```suggestion
  local actual
  actual=$(helm template \
```

test/unit/csi-daemonset.bats
```diff
@@ -177,17 +177,40 @@ load _helpers
...
+      --show-only templates/csi-daemonset.yaml \
+      --set "csi.enabled=true" \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].args[1]' | tee /dev/stderr)
+  [ "${actual}" = "--log-level=info" ]
+}
+
+@test "csi/daemonset: log-level is configurable" {
+  cd `chart_dir`
+  local actual=$(helm template \
```
benashz, line 204: nit:
```suggestion
  local actual
  actual=$(helm template \
```

test/unit/csi-daemonset.bats
```diff
@@ -177,17 +177,40 @@ load _helpers
       --set "csi.enabled=true" \
       . | tee /dev/stderr |
       yq -r '.spec.template.spec.containers[0].args[1]' | tee /dev/stderr)
-  [ "${actual}" = "--debug=false" ]
+  [ "${actual}" = "--log-level=info" ]
 
   local actual=$(helm template \
```
benashz, line 182: local is already declared above, we can probably drop the extra declaration here.

values.yaml
```diff
@@ -1225,7 +1225,12 @@ csi:
     # Number of seconds after which the probe times out.
     timeoutSeconds: 3
 
-  # Enables debug logging.
+  # Configures the log verbosity of the injector.
+  # Supported log levels include: trace, debug, info, warn, error, and off
```
benashz, line 1229:
```suggestion
  # Configures the log level for the Vault CSI provider.
  # Supported log levels include: trace, debug, info, warn, error, and off
```

benashz (commented):
Add some initial feedback. Overall this PR is looking good. We will need to probably get a CSI provider release with your fixes so that the tests can pass.

For future consideration, we want to have a better integration with the CSI project e.g: test against development CSI provider builds for example. We would need to add some support VAI and CSI version checking like we do here: https://github.com/hashicorp/vault-secrets-operator/blob/main/scripts/check-versions.sh

templates/csi-daemonset.yaml
```diff
@@ -55,7 +55,11 @@ spec:
           imagePullPolicy: {{ .Values.csi.image.pullPolicy }}
           args:
             - --endpoint=/provider/vault.sock
-            - --debug={{ .Values.csi.debug }}
+            {{- if .Values.csi.debug }}
+            - --log-level=debug
+            {{- else }}
+            - --log-level={{ .Values.csi.logLevel }}
+            {{- end }}
```
tvoran, line 62: I wonder if we should flip the precedence here?

benashz: @tvoran  I believe we want to honor `.csi.debug` if it is set, no? If not, then we would be changing the user's current debug setup.

templates/csi-daemonset.yaml
```diff
@@ -55,7 +55,13 @@ spec:
           imagePullPolicy: {{ .Values.csi.image.pullPolicy }}
           args:
             - --endpoint=/provider/vault.sock
-            - --debug={{ .Values.csi.debug }}
+            {{- if .Values.csi.logLevel }}
+            - --log-level={{ .Values.csi.logLevel }}
+            {{- else if .Values.csi.debug }}
```
benashz, line 60: Since `.csi.debug` existed prior to this change, we will probably want to honor the previous setting of `csi.debug`, otherwise we would be introducing a behavioral change that the user is not expecting.

benashz (commented):
Looking good!

CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
 ## Unreleased
 
+* Default `vault` version updated to 1.17.3
```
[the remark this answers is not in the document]
Zlaticanin: I was following the example [here](https://github.com/hashicorp/vault-helm/blob/main/CHANGELOG.md?plain=1#L24-L29)

templates/csi-daemonset.yaml
Zlaticanin: Very true! Good catch! This would be a breaking change, I reverted it. Thank you 🙏

CHANGELOG.md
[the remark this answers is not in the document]
benashz: Ah, I think I misunderstood the intention. These are the defaults in the values.

benashz (approved):
LGTM!

templates/csi-daemonset.yaml
tvoran: @benashz Yeah I wasn't suggesting removing `.csi.debug`, just to prefer `.csi.logLevel` if both `.csi.debug` and `.csi.logLevel` happen to be set. But I was forgetting that's hard to do with helm since logLevel has a default non-empty value, so 👍

PR #1051, opened 2024-08-20
Update values.yaml

HTHDOTCOM:


values.yaml
```diff
@@ -68,7 +68,7 @@ injector:
   # image sets the repo and tag of the vault-k8s image to use for the injector.
   image:
     repository: "hashicorp/vault-k8s"
-    tag: "1.4.2"
+    tag: "1.16.1"
```
tvoran, line 71: The latest release of vault-k8s is currently v1.4.2: https://github.com/hashicorp/vault-k8s/releases

PR #1074, opened 2024-11-18
ci: latest five k8s versions

tvoran:
Use the latest five k8s versions to match the other k8s projects, and updating the kind version and go for tests.


hashiblaum (approved):

tvoran:
Thanks!

PR #1076, opened 2024-11-20
Update the chart version for release v0.29.1

tvoran:
And adjust the k8s versions in the Makefile and README to match #1074


tvoran:
Thanks!

PR #1091, opened 2025-02-11
Add ability to disable k8s service links

der-eismann:
Hey everyone, we'd love to add this feature to the Helm chart. By default Kubernetes will inject all services via environment variables to a pod. While this normally is no big issue, Vault prints all environment names of variables on startup, causing a huge block in the logs that makes it hard to read.

By setting `enableServiceLinks` to false, this behavior can be stopped.


zewelor:
That would be greatThanks for opening this PR. This would be useful for us as well.

We run Vault in clusters with many Services, and the default Kubernetes service-link environment variables make the Vault startup logs very noisy. Having `server.enableServiceLinks=false` configurable in the chart would let us opt out without patching the rendered StatefulSet.

Since the default remains `true`, this looks backwards-compatible.

PR #1096, opened 2025-03-10
Update values.yaml - Move resource documentation to correct section in values.yaml

ahmadfikrin:
This PR moves the comment about resource requests and limits to the correct section in `values.yaml` . Previously, the documentation was misplaced under the `image` section, but it should be next to the `resources`  key to improve clarity. No functional changes were made—only the comment position was adjusted.


tvoran (approved):
Good catch, thanks!

PR #1099, opened 2025-03-19
server: support setting custom preStop commands

### LINKED ISSUE #1089: Add support for Prestop hook to execute custom commands
Currently there is no way to provide custom commands that will run during pre stop hook, there is an option to specify such commands for post start hook. In some cases users may want to run commands like leader step down for graceful Vault restart, which is not currently possible.

A helm chart configuration variable to specify custom pre stop commands to be executed.

There is no alternative to this feature.

**Additional context**
Link to helm chart template where configuration option need to be added: https://github.com/hashicorp/vault-helm/blob/b3daaad683f48a9e26a095c84e0a83c85060a249/templates/server-statefulset.yaml#L208

tvoran:
Adds a `server.preStop` option to set custom preStop commands for the Vault server pods.

Fixes #1089

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Changes:
 * Default `vault-k8s` version updated to 1.6.2
 * Tested with Kubernetes versions 1.28-1.32
 
+Features:
+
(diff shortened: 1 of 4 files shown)


benashz (approved):
LGTM!

tvoran:
Thanks!

PR #1106, opened 2025-04-08
support custom exec command in server readiness probe

bitnik:
This PR supports custom exec command in server readiness probe. If `server.readinessProbe.execCommand` is set, then it overwrites the default exec command in server readiness probe.

We need to be able to do this since we enforce mTLS by setting `tls_require_and_verify_client_cert = "true"` and so have to provide client certificates in readiness and liveness probes.

Note: I executed tests with `docker run -it --rm -v "${PWD}:/test" hashicorpdev/vault-helm-test:latest bats /test/test/unit/server-statefulset.bats -f "server/standalone-StatefulSet: readiness"`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Features:
 
 * server: Add Vault Enterprise redundancy zones support (requires Kubernetes 1.35+)
 
+Improvements:
+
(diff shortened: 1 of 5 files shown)


bitnik:
Hello, when can we have this merged?

heatherezell:
Hi, apologies for the delay - I'll get this in front of our teams in the next weekly meeting. Thanks!

PR #1131, opened 2025-08-22
feat: add metricRelabelings to serviceMonitor

adberger:
Vault emits metrics with the `cluster` label set.
The `cluster` label is used in many environments for multi-cloud purposes (dashboards).
Therefore it would be nice, if this label could be relabeled before ingesting into prometheus.


tvoran (approved):
Thanks! We'll get this out in the next release later this month.

(test failures are unrelated to this PR)

PR #1142, opened 2025-10-13
feat: Add possibility to use httproute resource

### LINKED ISSUE #1044: Deployment options for Gateway API
Kubernetes is moving toward adopting the new [Gateway API](https://gateway-api.sigs.k8s.io). While existing deployments are likely to continue using Ingress for some time, we are advised to avoid new deployments using Ingress, and to use Gateways instead.

Implement parallel gateway API-based objects under a switch, for example: `server.gateway.enabled` in helm values.

Disabling ingress and exposing the vault server using objects managed as a separate kustomization.

Kaswob:
Add a possibilty to use Gateway API HTTPRoute
Add unit tests
Update schema
Update default values

Resolves https://github.com/hashicorp/vault-helm/issues/1044

## PCI review checklist

### DIFF
--- a/test/unit/server-httproute.bats
+++ b/test/unit/server-httproute.bats
@@ -1,0 +1,364 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "server/httproute: disabled by default" {
+  cd `chart_dir`
+  local actual=$( (helm template \
+      --show-only templates/server-httproute.yaml  \
+      . || echo "---") | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "server/httproute: namespace" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-httproute.yaml  \
+      --set 'server.httproute.enabled=true' \
+      --namespace foo \
+      . | tee /dev/stderr |
+      yq -r '.metadata.namespace' | tee /dev/stderr)
+  [ "${actual}" = "foo" ]
+  local actual=$(helm template \
+      --show-only templates/server-httproute.yaml  \
+      --set 'server.httproute.enabled=true' \
+      --set 'global.namespace=bar' \
+      --namespace foo \
+      . | tee /dev/stderr |
+      yq -r '.metadata.namespace' | tee /dev/stderr)
+  [ "${actual}" = "bar" ]
+}
+
+@test "server/httproute: disable by injector.externalVaultAddr" {
+  cd `chart_dir`
+  local actual=$( (helm template \
+      --show-only templates/server-httproute.yaml  \
+      --set 'server.httproute.enabled=true' \
+      --set 'injector.externalVaultAddr=http://vault-outside' \
+      . || echo "---") | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "server/httproute: checking host entry gets added and path is /" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-httproute.yaml \
+      --set 'server.httproute.enabled=true' \
+      --set 'server.httproute.hostnames[0]=test.com' \
+      --set 'server.httproute.matches.path[0].type=PathPrefix' \
+      --set 'server.httproute.matches.path[0].value=/' \
+      . | tee /dev/stderr |
+      yq  -r '.spec.hostnames[0]' | tee /dev/stderr)
+  [ "${actual}" = 'test.com' ]
+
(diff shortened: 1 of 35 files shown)


timetofaceit:
@tvoran sorry for ping, but maybe you can have a look at this? Would be great to start migrating officially.

MissakaI:
Hi,

Sorry again for the ping. Still waiting for the official migrations.

As for the K8s docs [https://kubernetes.io/docs/concepts/services-networking/ingress/](https://kubernetes.io/docs/concepts/services-networking/ingress/):

> Note:
>
> The Kubernetes project recommends using [Gateway](https://gateway-api.sigs.k8s.io/) instead of [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/). The Ingress API has been frozen.
> 
> This means that:
> 
> - The Ingress API is generally available, and is subject to the [stability guarantees](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api) for generally available APIs. The Kubernetes project has no plans to remove Ingress from Kubernetes.
> 
> - The Ingress API is no longer being developed, and will have no further changes or updates made to it.

And official ingress-nginx

> 
>
> ## Retiring
>
> [What You Need to Know about Ingress NGINX Retirement](https://www.kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/):
>
> * Best-effort maintenance will continue until March 2026.
> * Afterward, there will be no further releases, no bugfixes, and no updates to resolve any security vulnerabilities that may be discovered.
> * Existing deployments of Ingress NGINX will not be broken.
> * Existing project artifacts such as Helm charts and container images will remain available.

Capability to do the migration so from the official chart is much appreciated.

johnjosephg:
Hi @Kaswob, could you please merge the latest changes from main into this branch, since the acceptance tests have now been updated.

Kaswob:
@johnjosephg sure thing, done.

test/unit/server-httproute.bats
```diff
@@ -0,0 +1,364 @@
+#!/usr/bin/env bats
+
```
johnjosephg, line 10: Could we have some documentation for all of the tests written?

values.schema.json
```diff
@@ -857,6 +857,44 @@
...
+                        "hostnames": {
+                            "type": "array"
+                        },
+                        "labels": {
+                            "type": "object"
+                        },
+                        "matches": {
+                            "type": [
+                                "array",
+                                "object"
```
johnjosephg: matches must be an array, it cannot be both, as defined in values.yaml file. 
The corresponding changes must be made in the test file as well.

test/unit/server-httproute.bats
```diff
@@ -0,0 +1,364 @@
...
+  [ "${actual}" = "false" ]
+}
+
+@test "server/httproute: checking host entry gets added and path is /" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      --show-only templates/server-httproute.yaml \
+      --set 'server.httproute.enabled=true' \
+      --set 'server.httproute.hostnames[0]=test.com' \
+      --set 'server.httproute.matches.path[0].type=PathPrefix' \
```
johnjosephg: instead of having this as an object, where path is the array, we should have matches as the array as shown.
```
--set 'server.httproute.matches[0].path.type=PathPrefix'
```
Please make the changes wherever applicable.

johnjosephg (changes requested):

test/unit/server-httproute.bats
Kaswob: Thanks for the review. Sure, docs added, tests and schema fixed. Also did make parentRefs required - it has to be specified so the HTTPRoute can attach to a Gateway.

johnjosephg (changes requested):
Could you please also add http route to an acceptance test, just to make sure that this new resource works with the existing vault setup?

Kaswob:
> Could you please also add http route to an acceptance test, just to make sure that this new resource works with the existing vault setup?

Added a set of simple acceptance tests. Tests verify if HTTPRoute is created and proper fields are set. There is also a test to confirm default chart behaviour so HTTPRoute is disabled by default. Only requirements is to have Gateway API CRDs on the cluster.

johnjosephg (dismissed):
LGTM

vijayavelsekar:
@Kaswob Please add an entry in the CHANGELOG.md file for this change at the top.
```
## Unreleased

Features:

* Add support for Kubernetes Gateway API HTTPRoute resource [#1142](https://github.com/hashicorp/vault-helm/pull/1142)

```

Kaswob:
@vijayavelsekar rebased and added info to CHANGELOG. @johnjosephg I think you have to approve again, sorry about that. Thanks!

vijayavelsekar (approved):

PR #1143, opened 2025-10-13
feat: Add extraObjects for deploying additional Kubernetes objects

### LINKED ISSUE #1146: Feature: Add extraObjects Parameter
Currently, teams must manually apply ConfigMaps, Secrets, and other supporting resources outside of the Helm release, which:

- Increases deployment complexity
- Reduces reproducibility
- Creates potential synchronization issues between resources

I propose introducing an `extraObjects` parameter in Helm charts that allows:
- Inline definition of additional Kubernetes resources
- Templated resource creation
- Management through Helm's native resource tracking

**Additional context**
related PR https://github.com/hashicorp/vault-helm/pull/1143

giuliocalzo:
## Summary

Adds `extraObjects` for deploying additional Kubernetes objects as part of the chart's release: objects the chart does not model, such as a cert-manager `Certificate` issuing the Secret referenced by `server.volumes`. Today these need a separate `kubectl apply`, which splits them from the release that depends on them.

Resolves #1146.

## Behaviour

- Accepts a list or a map. Entries are YAML maps or strings, each rendered with `tpl` against the chart's context, so they can use helpers such as `vault.fullname` and `vault.namespace`.
- The map form lets a per-environment values file add or override a single object, since Helm merges maps key by key but replaces lists wholesale.
- An entry that renders to nothing is skipped, so it can be wrapped in a conditional tied to a component being enabled.
- An entry that is neither a map nor a string fails rendering with an error naming its index or key.
- Defaults to `null`, so charts that do not set `extraObjects` render exactly as before.

## Example

```yaml
extraObjects:
  appConfig:
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: '{{ include "vault.fullname" . }}-app-config'
    data:
      key: value
  csiCertificate: |
    {{- if .Values.csi.enabled }}
    apiVersion: cert-manager.io/v1
    kind: Certificate
    metadata:
      name: {{ include "vault.fullname" . }}-csi-tls
    spec:
      secretName: vault-csi-tls
      issuerRef:
        name: internal-issuer
        kind: ClusterIssuer
      dnsNames:
        - vault-csi.example.com
    {{- end }}
```

## Changes

`templates/extra-objects.yaml`, docs in `values.yaml`, a `values.schema.json` entry, a CHANGELOG entry, and 12 unit tests in `test/unit/extra-objects.bats`. Full unit suite passes locally. Rebased onto `main` at v0.34.1.

The extensions over the original template in this PR came from @khateeb15 in #1206, credited as co-author on the commit.

## PCI review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+Features:
+
+* Add `extraObjects` for deploying additional Kubernetes objects as part of the chart's release. Accepts a list or a map, whose entries are either YAML maps or templated strings, and renders each entry with `tpl` against the chart's context [#1143](https://github.com/hashicorp/vault-helm/pull/1143)
+
 ## 0.34.1 (August 13, 2026)
 
 Changes:
--- a/templates/extra-objects.yaml
+++ b/templates/extra-objects.yaml
@@ -1,0 +1,38 @@
+{{/*
+Copyright IBM Corp. 2018, 2026
+SPDX-License-Identifier: MPL-2.0
+*/}}
+
+{{- /*
+extraObjects is accepted as a list or as a map. Both are normalized to a list
+of entries keeping the list index or the map key, so that an entry of an
+unsupported type can be named in the error. An entry that renders to nothing is
+skipped rather than emitting an empty document, so that an entry may be wrapped
+in a conditional tying it to a component being enabled.
+*/ -}}
+{{- $entries := list -}}
+{{- $extraObjects := .Values.extraObjects | default list -}}
+{{- if kindIs "map" $extraObjects -}}
+{{- range $key, $object := $extraObjects -}}
+{{- $entries = append $entries (dict "id" $key "object" $object) -}}
+{{- end -}}
+{{- else -}}
+{{- range $index, $object := $extraObjects -}}
+{{- $entries = append $entries (dict "id" $index "object" $object) -}}
+{{- end -}}
+{{- end -}}
+{{- range $entry := $entries }}
+{{- $object := $entry.object }}
+{{- $rendered := "" }}
+{{- if kindIs "string" $object }}
+{{- $rendered = tpl $object $ }}
+{{- else if kindIs "map" $object }}
+{{- $rendered = tpl (toYaml $object) $ }}
+{{- else }}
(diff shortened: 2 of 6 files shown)


aqeelat:
Hi @vijayavelsekar @tvoran

Can you please take a look? I desperately need it to manage vault related stuff in argocd.

khateeb15:
Hi @giuliocalzo, thanks for opening this. We needed the same thing and have built out a fuller version of it, currently running in our test environments and working well, so rather than sit on it I have opened PR #1206 with it, and credited you as co-author since it keeps your `extraObjects` name, your string-entry form, and the overall approach from this PR.

What it adds on top of the four-line template here:

- Entries can be YAML maps, not only strings. `tpl` fails on a map, so as written this PR only accepts strings.
- `extraObjects` can be a map of named entries as well as a list. Helm replaces lists wholesale when merging values files but merges maps key by key, so the map form lets a per-environment overlay add or override one object without restating the others.
- An entry that renders to nothing is skipped instead of emitting a stray `---` document, so an entry can be wrapped in a conditional and tied to a component being enabled.
- An entry that is neither a map nor a string fails rendering with an error naming its index or key, rather than producing invalid YAML further down.
- A `values.schema.json` entry, 11 unit tests, a CHANGELOG entry, and the `Copyright IBM Corp.` header that the compliance workflow expects on every template.

No wish to step on this PR: I am equally happy to close mine and send the additions as a patch onto your branch if you would prefer to carry it, or for the maintainers to simply take whichever is easier to review. Whatever gets the feature landed.

@tvoran @vijayavelsekar this one has been open since October 2025 with a few people asking for it on #1146. Would either version be reviewable for an upcoming release?

giuliocalzo:
HI @khateeb15 I'm happy to add all missing fields in my PR, I will add on Monday as first thing

giuliocalzo:
@khateeb15 thanks, that is a better shape than what I had. I have folded all five points into this PR and credited you as co-author on the commit: map entries as well as strings, the map form alongside the list, empty entries skipped, a named error for unsupported types, plus the `values.schema.json` entry, the unit tests, the CHANGELOG entry and the license header. The template is renamed to `extra-objects.yaml` to match the values key, and the branch is rebased onto `main` at v0.34.1. Full unit suite passes locally.

PR #1166, opened 2026-01-08
CI: test against k8s 1.35

Zlaticanin:


tvoran (approved):
Could you also update the changelog here too? Something like "Tested with Kubernetes versions 1.35-1.31".

PR #896, opened 2023-05-19
Allow additional annotations for standby and active services via config

tekicat:
Revival of https://github.com/hashicorp/vault-helm/pull/546

Currently there is no ability to annotate services for active and standby services, as they inherit their settings from the main server service configuration. This is a problem,  when you need to attach different annotations to the different services. Instead of just allowing for setting of annotations, I opted to allow configuring the services for more flexibility.

Helps to close https://github.com/hashicorp/vault-helm/issues/674

Partially resolves https://github.com/hashicorp/vault-helm/issues/588

Thanks @jamesgoodhouse

Issue #674, opened 2022-01-09
service annotations applied to both vault and vault-internal services


raffaelespazzoli:
**Describe the bug**
the annotations in the service section are applied to both the `vault` and the `vault-internal` services and there is not way to distinguish between the two.
When deploying to OpenShift and the annotation is used to request a certificate, this causes a race condition because two services are trying to modify the same secret (containing the certificate)

**To Reproduce**
1. Install chart

with these options:

```
global:
  openshift: true
  tlsDisable: false
  
  ...
server:
  service: 
    annotations:
      service.beta.openshift.io/serving-cert-secret-name: vault-server-tls

```

try to connect to the vault instance via the vault service. 50% of the times the connection will fails because the cert cannot be validated.

Suggested solution: allow for a way to attribute annotations to individual services.
Also when did this change? As a workaround one could use the old version in which this used to work.

eye0fra:
Hi @jasonodonnell and @tvoran,

Can you have a look?

Thanks a lot

tvoran:
Hi @raffaelespazzoli, do you recall which old version used to work for you this way?

raffaelespazzoli:
no I don't recall the old version, actually people have looked into this and said that the issue was there since the beginning, so probably I was just lucky.

PR #1179, opened 2026-03-28
server: add per-service annotations for main and internal services

mbaldessari:
PR #896 added per-service annotation support for the HA active and standby services (server.service.active.annotations and server.service.standby.annotations), partially addressing issue #674. However, it left out two services that still only received the generic server.service.annotations:

  - The main service (server-service.yaml): the primary client-facing service, rendered in every deployment mode (standalone, dev, and HA).
  - The internal headless service (server-headless-service.yaml): the ClusterIP=None service used for pod DNS and Raft join operations.

Without per-service annotations on these two services, setting server.service.annotations applies to *all* four services. This makes it impossible to, for example, set different annotations on the main service versus the internal headless service which is the problem that was described in issue 674 and which I am hitting as well.

This commit adds:

  - server.service.main.annotations: annotations applied only to the main service (server-service.yaml), merged after (and overriding) the generic server.service.annotations.
  - server.service.internal.annotations: annotations applied only to the internal headless service (server-headless-service.yaml), with the same merge/override semantics.

"main" was chosen over "nonha" because this service is rendered in all deployment modes, not just non-HA.

Closes: #647

## PCI review checklist

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -725,7 +725,11 @@ Sets extra route annotations
 {{- end -}}
 
 {{/*
-Sets extra vault server Service annotations
+Sets extra vault server Service annotations.
+These are applied to all services: main, internal, active, and standby.
+Use the per-service annotation templates (vault.service.main.annotations,
+vault.service.internal.annotations, vault.service.active.annotations,
+vault.service.standby.annotations) to target a specific service.
 */}}
 {{- define "vault.service.annotations" -}}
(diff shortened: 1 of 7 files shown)


jneatherwayTPG:
+1 👀  for this.  My requirement is for different MetalLB related annotations to the main and active services.

PR #1180, opened 2026-04-13
Make the audit and data storage PVCs configurable

gblues:
== DETAILS
The use case here is for migrating from a home-grown Vault kubernetes deployment to using the official Helm chart.

Our home-grown deployment used a different name for the statefulset PVCs ("raft" instead of "data"). While it is certainly possible to do some kubernetes magic to rename the existing PVCs, doing so would require some downtime and risks an extended outage if things go sideways.

It would be much less risky if the chart provided a way to specify the name of the PVC, so that we can point it at our existing data with minimal fuss.

This commit provides this functionality in the following way:

- add a `name` field under `.server.auditStorage` and `.server.dataStorage`
- updates `values.yaml` to set these to a default that matches the hardcoded default names
- updates the templates to look up the name provided in values instead of using the hardcoded names

unit tests are included to validate this functionality works as expected.

## PCI review checklist

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -244,24 +244,38 @@ Set's additional environment variables based on the mode.
 {{- end -}}
 
 {{/*
+gets the configured audit storage name
+*/}}
+{{- define "vault.auditStorageName" }}
+{{- .Values.server.auditStorage.name | default "audit" }}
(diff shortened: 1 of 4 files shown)


gblues:
@tvoran are you still maintaining this? please 👀

PR #1185, opened 2026-05-21
fix(server): add IPC_LOCK capability to default container securityContext for Vault 2.0.1+

jadeidev:
- Vault 2.0.1 introduced a breaking change: the cap_ipc_lock file capability is no longer set on the vault binary at image build time, so the container runtime must grant IPC_LOCK instead. 
[https://github.com/hashicorp/vault/blob/main/CHANGELOG.md?plain=1#L8C1-L11C1](https://github.com/hashicorp/vault/blob/main/CHANGELOG.md?plain=1#L8C1-L11C1)
  - This change adds a Linux capability (IPC_LOCK) to the Vault server container. 

## PCI review checklist

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -603,6 +603,9 @@ securityContext for the statefulset vault container
   {{- else if not .Values.global.openshift }}
           securityContext:
             allowPrivilegeEscalation: false
+            capabilities:
+              add:
+                - IPC_LOCK
   {{- end }}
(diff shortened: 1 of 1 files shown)


heatherezell:
This is being tracked internally and has been reported in https://github.com/hashicorp/vault/issues/31919. I'm closing this PR because of a need to handle this with our security operations teams and engineering. Thanks!

PR #1202, opened 2026-08-06
To default helm-vault installation to vault enterprise when license is provided

harshit-nema:
**ISSUE :** 
The chart already wired the licence volume, mount, and VAULT_LICENSE_PATH env var when `server.enterpriseLicense.secretName` was set — but still served the Community Edition image, forcing Enterprise customers to manually override `server.image.repository` and `server.image.tag`.

**SOLUTION :** 
Added `vault.isEnterprise` as a single source of truth for the license check, and two generic helpers — `vault.resolveImageRepository` and `vault.resolveImageTag` — containing the shared resolution logic. Six thin wrapper helpers delegate to these for the server, injector agent, and CSI agent respectively.

Resolution rules:
- **repository**: if unset or `hashicorp/vault` (the default) → auto-select `hashicorp/vault-enterprise` (license) or `hashicorp/vault` (no license). Any other explicit repo is returned verbatim.
- **tag**: if empty → auto-select `AppVersion-ent` (license) or `AppVersion` (no license). Any explicit tag is always returned verbatim — never mutated.

Image render sites in `server-statefulset.yaml`, `injector-deployment.yaml`, and `csi-daemonset.yaml `were updated to use these helpers. Empty tags now fall back to `Chart.AppVersion` instead of `latest`, making CE installs reproducible by default.

`server.enterpriseLicense.secretName` is the only configuration needed — no manual image overrides required for any component.

**IMPLEMENTATION VERIFICATION :**

- helm-vault ent installation 
`helm install vault . --set server.enterpriseLicense.secretName=vault-ent-license`
image

- helm-vault community edition installation
`helm install vault .  `
image

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -125,6 +125,41 @@ Compute if the ui is enabled.
 {{- end -}}
 
 {{/*
+Resolve the Vault server image repository.
+If server.image.repository is explicitly set to a non-default value, use it.
+Otherwise, auto-select hashicorp/vault-enterprise when an enterprise license
+secret is configured, falling back to hashicorp/vault for Community Edition.
+*/}}
+{{- define "vault.imageRepository" -}}
+{{- if and .Values.server.image.repository (ne .Values.server.image.repository "hashicorp/vault") -}}
+  {{- .Values.server.image.repository -}}
+{{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  hashicorp/vault-enterprise
+{{- else -}}
+  {{- .Values.server.image.repository | default "hashicorp/vault" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Resolve the Vault server image tag.
+If server.enterpriseLicense.secretName is set and the tag does not already
+carry the -ent suffix, append it automatically.
+Otherwise return the tag as-is.
+*/}}
+{{- define "vault.imageTag" -}}
+{{- $tag := .Values.server.image.tag | default "latest" -}}
+{{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  {{- if not (hasSuffix "-ent" $tag) -}}
+    {{- printf "%s-ent" $tag -}}
+  {{- else -}}
+    {{- $tag -}}
+  {{- end -}}
+{{- else -}}
+  {{- $tag -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Compute the maximum number of unavailable replicas for the PodDisruptionBudget.
 This defaults to ⌊(n-1)/2⌋ (equivalently, ceil(n/2)-1) where n is the number of
 members of the server cluster.
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -147,7 +147,7 @@ carry the -ent suffix, append it automatically.
 Otherwise return the tag as-is.
 */}}
 {{- define "vault.imageTag" -}}
-{{- $tag := .Values.server.image.tag | default "latest" -}}
+{{- $tag := .Values.server.image.tag | default .Chart.AppVersion -}}
 {{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
   {{- if not (hasSuffix "-ent" $tag) -}}
     {{- printf "%s-ent" $tag -}}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -136,7 +136,7 @@ secret is configured, falling back to hashicorp/vault for Community Edition.
 {{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
   hashicorp/vault-enterprise
 {{- else -}}
-  {{- .Values.server.image.repository | default "hashicorp/vault" -}}
+  hashicorp/vault
 {{- end -}}
 {{- end -}}
 
@@ -174,7 +174,7 @@ This helper mirrors the logic of vault.imageRepository:
 {{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
   hashicorp/vault-enterprise
 {{- else -}}
-  {{- .Values.injector.agentImage.repository | default "hashicorp/vault" -}}
+  hashicorp/vault
 {{- end -}}
 {{- end -}}
 
@@ -213,7 +213,7 @@ vault.agentImageRepository:
 {{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
   hashicorp/vault-enterprise
 {{- else -}}
-  {{- .Values.csi.agent.image.repository | default "hashicorp/vault" -}}
+  hashicorp/vault
 {{- end -}}
 {{- end -}}
 
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -147,7 +147,88 @@ carry the -ent suffix, append it automatically.
 Otherwise return the tag as-is.
 */}}
 {{- define "vault.imageTag" -}}
-{{- $tag := .Values.server.image.tag | default .Chart.AppVersion -}}
+{{- $tag := .Values.server.image.tag | default "latest" -}}
+{{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  {{- if not (hasSuffix "-ent" $tag) -}}
+    {{- printf "%s-ent" $tag -}}
+  {{- else -}}
+    {{- $tag -}}
+  {{- end -}}
+{{- else -}}
+  {{- $tag -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+This helper mirrors the logic of vault.imageRepository:
+- If the user has explicitly overridden injector.agentImage.repository to
+  something other than the default "hashicorp/vault", honour that value as-is
+  (custom registry scenario).
+- If server.enterpriseLicense.secretName + secretKey are both set, auto-select
+  hashicorp/vault-enterprise so the agent edition always matches the server.
+- Otherwise fall back to hashicorp/vault (CE).
+*/}}
+{{- define "vault.agentImageRepository" -}}
+{{- if and .Values.injector.agentImage.repository (ne .Values.injector.agentImage.repository "hashicorp/vault") -}}
+  {{- .Values.injector.agentImage.repository -}}
+{{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  hashicorp/vault-enterprise
+{{- else -}}
+  {{- .Values.injector.agentImage.repository | default "hashicorp/vault" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Companion to vault.agentImageRepository. Vault Enterprise images use a
+"-ent" tag suffix (e.g. 2.0.3-ent). This helper appends "-ent" automatically when the enterprise license
+is configured, matching the behaviour of vault.imageTag for the server.
+If the tag already carries the suffix it is left unchanged (idempotent).
+*/}}
+{{- define "vault.agentImageTag" -}}
+{{- $tag := .Values.injector.agentImage.tag | default "latest" -}}
+{{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  {{- if not (hasSuffix "-ent" $tag) -}}
+    {{- printf "%s-ent" $tag -}}
+  {{- else -}}
+    {{- $tag -}}
+  {{- end -}}
+{{- else -}}
+  {{- $tag -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+This helper mirrors the logic of vault.imageRepository and
+vault.agentImageRepository:
+- If csi.agent.image.repository is explicitly set to a non-default value,
+  honour it as-is (custom registry scenario).
+- If server.enterpriseLicense.secretName + secretKey are both set,
+  auto-select hashicorp/vault-enterprise so the CSI agent edition always
+  matches the server.
+- Otherwise fall back to hashicorp/vault (CE), preserving existing behaviour.
+*/}}
+{{- define "vault.csiAgentImageRepository" -}}
+{{- if and .Values.csi.agent.image.repository (ne .Values.csi.agent.image.repository "hashicorp/vault") -}}
+  {{- .Values.csi.agent.image.repository -}}
+{{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  hashicorp/vault-enterprise
+{{- else -}}
+  {{- .Values.csi.agent.image.repository | default "hashicorp/vault" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Companion to vault.csiAgentImageRepository. Vault Enterprise images on
+DockerHub require the "-ent" tag suffix (e.g. 2.0.3-ent). Fixing the
+repository to hashicorp/vault-enterprise without also fixing the tag would
+result in an ImagePullBackOff because hashicorp/vault-enterprise:2.0.3
+does not exist. This helper appends "-ent" automatically when the enterprise
+license is configured, matching the behaviour of vault.imageTag for the server
+and vault.agentImageTag for the injector sidecar.
+If the tag already carries the "-ent" suffix it is left unchanged (idempotent).
+*/}}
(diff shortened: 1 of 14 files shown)


templates/_helpers.tpl
```diff
@@ -124,6 +124,122 @@ Compute if the ui is enabled.
...
+Otherwise, auto-select hashicorp/vault-enterprise when an enterprise license
+secret is configured, falling back to hashicorp/vault for Community Edition.
+*/}}
+{{- define "vault.imageRepository" -}}
+{{- if and .Values.server.image.repository (ne .Values.server.image.repository "hashicorp/vault") -}}
+  {{- .Values.server.image.repository -}}
+{{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  hashicorp/vault-enterprise
+{{- else -}}
+  {{- .Values.server.image.repository | default "hashicorp/vault" -}}
```
vijayavelsekar: Suggesting below change because at this point of time the repository is either empty or hashicorp/vault.

```suggestion
  hashicorp/vault
```

templates/_helpers.tpl
```diff
@@ -124,6 +124,122 @@ Compute if the ui is enabled.
...
+  hashicorp/vault-enterprise so the agent edition always matches the server.
+- Otherwise fall back to hashicorp/vault (CE).
+*/}}
+{{- define "vault.agentImageRepository" -}}
+{{- if and .Values.injector.agentImage.repository (ne .Values.injector.agentImage.repository "hashicorp/vault") -}}
+  {{- .Values.injector.agentImage.repository -}}
+{{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  hashicorp/vault-enterprise
+{{- else -}}
+  {{- .Values.injector.agentImage.repository | default "hashicorp/vault" -}}
```
vijayavelsekar:
```suggestion
  hashicorp/vault
```

templates/_helpers.tpl
```diff
@@ -124,6 +124,122 @@ Compute if the ui is enabled.
...
+  matches the server.
+- Otherwise fall back to hashicorp/vault (CE), preserving existing behaviour.
+*/}}
+{{- define "vault.csiAgentImageRepository" -}}
+{{- if and .Values.csi.agent.image.repository (ne .Values.csi.agent.image.repository "hashicorp/vault") -}}
+  {{- .Values.csi.agent.image.repository -}}
+{{- else if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  hashicorp/vault-enterprise
+{{- else -}}
+  {{- .Values.csi.agent.image.repository | default "hashicorp/vault" -}}
```
vijayavelsekar:
```suggestion
  hashicorp/vault
```

templates/_helpers.tpl
```diff
@@ -124,6 +124,122 @@ Compute if the ui is enabled.
...
+{{- end -}}
+
+{{/*
+Resolve the Vault server image tag.
+If server.enterpriseLicense.secretName is set and the tag does not already
+carry the -ent suffix, append it automatically.
+Otherwise return the tag as-is.
+*/}}
+{{- define "vault.imageTag" -}}
+{{- $tag := .Values.server.image.tag | default "latest" -}}
```
vijayavelsekar: tag will never be nil as it have a default value from the values.yaml file.
Also having default as `latest` seems contradicting to the default value in the vaules.yaml file.
Or we can have `| default $.Chart.AppVersion` and remove all the `tag` attributes from the values file so we can centrally control the vault version in one place(Chart.yaml)
We can discuss this with the team about this.

```suggestion
{{- $tag := .Values.server.image.tag -}}
```

templates/_helpers.tpl
```diff
@@ -124,6 +124,122 @@ Compute if the ui is enabled.
...
+{{- end -}}
+
+{{/*
+Companion to vault.agentImageRepository. Vault Enterprise images use a
+"-ent" tag suffix (e.g. 2.0.3-ent). This helper appends "-ent" automatically when the enterprise license
+is configured, matching the behaviour of vault.imageTag for the server.
+If the tag already carries the suffix it is left unchanged (idempotent).
+*/}}
+{{- define "vault.agentImageTag" -}}
+{{- $tag := .Values.injector.agentImage.tag | default "latest" -}}
```
vijayavelsekar:
```suggestion
{{- $tag := .Values.injector.agentImage.tag -}}
```

templates/_helpers.tpl
```diff
@@ -124,6 +124,122 @@ Compute if the ui is enabled.
...
+DockerHub require the "-ent" tag suffix (e.g. 2.0.3-ent). Fixing the
+repository to hashicorp/vault-enterprise without also fixing the tag would
+result in an ImagePullBackOff because hashicorp/vault-enterprise:2.0.3
+does not exist. This helper appends "-ent" automatically when the enterprise
+license is configured, matching the behaviour of vault.imageTag for the server
+and vault.agentImageTag for the injector sidecar.
+If the tag already carries the "-ent" suffix it is left unchanged (idempotent).
+*/}}
+{{- define "vault.csiAgentImageTag" -}}
+{{- $tag := .Values.csi.agent.image.tag | default "latest" -}}
```
vijayavelsekar:
```suggestion
{{- $tag := .Values.csi.agent.image.tag -}}
```

vijayavelsekar (changes requested):

values.yaml
```diff
@@ -74,9 +74,11 @@ injector:
   # agentImage sets the repo and tag of the Vault image to use for the Vault Agent
   # containers.  This should be set to the official Vault image.  Vault 1.3.1+ is
   # required.
+  # tag: leave empty to use Chart.AppVersion (central default). Override with
+  # --set injector.agentImage.tag=<version> to pin independently.
   agentImage:
     repository: "hashicorp/vault"
-    tag: "2.0.4"
+    tag: ""
```
vijayavelsekar, line 79: I guess its better to keep the tag as it was before and remove the `default .Chart.AppVersion` parts from _helpers.tpl file.
Also as per our last discussion AppVersion is not the right variable for vault version. It is actually the helm chart version.

harshit-nema: appVersion is vault application version 
version is helm chart version.
so right now, the fix I made is only when customised image is added, values are read from values.yaml, else its centralised to .Chart.AppVersion. Let me know if this looks fine.

harshit-nema: Keeping tag in values.yaml as non-empty as per the last discussion.

templates/_helpers.tpl
```diff
@@ -124,6 +124,126 @@ Compute if the ui is enabled.
...
+so that Chart.yaml is the single source of truth for the Vault version.
+If server.enterpriseLicense.secretName is set and the tag does not already
+carry the -ent suffix, append it automatically.
+Otherwise return the tag as-is.
+*/}}
+{{- define "vault.imageTag" -}}
+{{- $tag := .Values.server.image.tag | default .Chart.AppVersion -}}
+{{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  {{- if not (hasSuffix "-ent" $tag) -}}
+    {{- printf "%s-ent" $tag -}}
```
siyer-corp: When a user explicitly sets `server.image.tag`, this helper still appends `-ent` whenever a license is configured. For example, `server.image.tag=custom `renders as `custom-ent`, changing the existing custom-tag behavior and potentially breaking private-registry images. Please preserve explicitly supplied tags unchanged and only append -ent to the automatically selected default tag.

templates/_helpers.tpl
```diff
@@ -124,6 +124,126 @@ Compute if the ui is enabled.
...
+"-ent" tag suffix (e.g. 2.0.3-ent). This helper appends "-ent" automatically when the enterprise license
+is configured, matching the behaviour of vault.imageTag for the server.
+If injector.agentImage.tag is not set, falls back to $.Chart.AppVersion.
+If the tag already carries the suffix it is left unchanged (idempotent).
+*/}}
+{{- define "vault.agentImageTag" -}}
+{{- $tag := .Values.injector.agentImage.tag | default .Chart.AppVersion -}}
+{{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  {{- if not (hasSuffix "-ent" $tag) -}}
+    {{- printf "%s-ent" $tag -}}
```
siyer-corp: Same issue as line 155. Please preserve explicit custom tags exactly as supplied.

templates/_helpers.tpl
```diff
@@ -124,6 +124,126 @@ Compute if the ui is enabled.
...
+license is configured, matching the behaviour of vault.imageTag for the server
+and vault.agentImageTag for the injector sidecar.
+If csi.agent.image.tag is not set, falls back to $.Chart.AppVersion.
+If the tag already carries the "-ent" suffix it is left unchanged (idempotent).
+*/}}
+{{- define "vault.csiAgentImageTag" -}}
+{{- $tag := .Values.csi.agent.image.tag | default .Chart.AppVersion -}}
+{{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
+  {{- if not (hasSuffix "-ent" $tag) -}}
+    {{- printf "%s-ent" $tag -}}
```
siyer-corp: Same issue as line 155. Please preserve explicit custom tags exactly as supplied.

templates/_helpers.tpl
```diff
@@ -124,6 +124,126 @@ Compute if the ui is enabled.
...
 {{- end -}}
 
+{{/*
+Resolve the Vault server image repository.
+If server.image.repository is explicitly set to a non-default value, use it.
+Otherwise, auto-select hashicorp/vault-enterprise when an enterprise license
+secret is configured, falling back to hashicorp/vault for Community Edition.
+*/}}
+{{- define "vault.imageRepository" -}}
+{{- if and .Values.server.image.repository (ne .Values.server.image.repository "hashicorp/vault") -}}
```
siyer-corp: Please add regression tests covering an Enterprise license together with both custom repository and custom tag values. The expected result should preserve both values exactly, for example `mycorp/vault:custom`, rather than producing `mycorp/vault:custom-ent`. Equivalent coverage is needed for the injector and CSI agent helpers.

templates/_helpers.tpl
```diff
@@ -124,6 +124,126 @@ Compute if the ui is enabled.
...
+Resolve the Vault server image tag.
+If server.image.tag is set, use it; otherwise fall back to $.Chart.AppVersion
+so that Chart.yaml is the single source of truth for the Vault version.
+If server.enterpriseLicense.secretName is set and the tag does not already
+carry the -ent suffix, append it automatically.
+Otherwise return the tag as-is.
+*/}}
+{{- define "vault.imageTag" -}}
+{{- $tag := .Values.server.image.tag | default .Chart.AppVersion -}}
+{{- if and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey -}}
```
siyer-corp: This check is repeated many times. Could we extract this into another template and re-use that?

templates/_helpers.tpl
harshit-nema: Should i have a fallback also in place, where if provided `tag` is same as `Chart.AppVersion` and license is provided, then -ent is appended, but if provided version is different than `Chart.AppVersion`, ent will not be appended ?

values.yaml
harshit-nema: As per the updated implementation, values.yaml will have empty tag for server and agents in csi and agent-injector. Value will be fetched from Charts.appVersion.

values.yaml
```diff
@@ -415,9 +419,15 @@ server:
...
+  # image sets the repository and tag of the Vault server image.
+  # These fields only need to be overridden to pin a custom registry or
+  # a non-standard image name.
+  # When server.enterpriseLicense.secretName is set, the chart automatically
+  # selects hashicorp/vault-enterprise and appends the -ent tag suffix —
+  # no override of these fields is required for Enterprise installations.
   image:
     repository: "hashicorp/vault"
-    tag: "2.0.4"
+    tag: ""
```
vijayavelsekar: @harshit-nema & @siyer-corp 
**New Proposal:**
We need to discuss this if this is what the Customers wants.
New Proposal requires only the `values.yaml` has to be changed. Specifically only the `repository` and `tag` needs to be updated as `hashicorp/vault-enterprise` and `2.1.0-ent` respectively.

**Current Flow:**
```
1. Without repository tag goes to `hashicorp/vault` (Community Edition)
2. tag: with -ent fails
3. tag: without -ent pulls CE
```

**New Proposed Flow:**
```
1. Without repository tag goes to `hashicorp/vault-enterprise`
2. tag: with -ent pulls Enterprise
3. tag: without -ent fails
```

.github/workflows/update-helm-charts-index.yml
```diff
@@ -31,10 +31,3 @@ jobs:
             --ref main \
             -f SOURCE_TAG="${{ github.ref_name }}" \
             -f SOURCE_REPO="${{ github.repository }}"
-      - uses: hashicorp/actions-slack-status@v2
```
vijayavelsekar: These changes are already there in the main. Why is this showing up in this PR ?

values.yaml
```diff
@@ -76,7 +76,7 @@ injector:
   # required.
   agentImage:
     repository: "hashicorp/vault"
-    tag: "2.0.4"
+    tag: ""
```
vijayavelsekar: would it be more cleaner to remove the `tag: ""` entirely ?

.github/workflows/update-helm-charts-index.yml
harshit-nema: rebased the branch. This change is gone now.

templates/_helpers.tpl
```diff
@@ -124,6 +124,144 @@ Compute if the ui is enabled.
...
+{{/*
+Companion to vault.agentImageRepository. Vault Enterprise images use a
+"-ent" tag suffix (e.g. 2.0.3-ent).
+If injector.agentImage.tag is non-empty the user has explicitly supplied a
+tag — return it unchanged, no -ent suffix is appended.
+When injector.agentImage.tag is empty, fall back to Chart.AppVersion; if an
+enterprise license is also configured, append -ent automatically (idempotent
+if already present) so the agent edition matches the server.
+*/}}
+{{- define "vault.agentImageTag" -}}
```
vijayavelsekar: The repository and tag resolution logic across server, injector, and CSI agent is identical. We could keep `_helpers.tpl` DRY by extracting the common logic into two generic helpers (`vault.resolveImageRepository` and `vault.resolveImageTag`) and having the component helpers delegate to them:

```yaml
{{- define "vault.resolveImageRepository" -}}
{{- if and .repo (ne .repo "hashicorp/vault") -}}
  {{- .repo -}}
{{- else if include "vault.isEnterprise" .root -}}
  hashicorp/vault-enterprise
{{- else -}}
  {{- .repo | default "hashicorp/vault" -}}
{{- end -}}
{{- end -}}

{{- define "vault.resolveImageTag" -}}
{{- $tag := .tag | default .root.Chart.AppVersion -}}
{{- if and (not .tag) (include "vault.isEnterprise" .root) -}}
  {{- if not (hasSuffix "-ent" $tag) -}}
    {{- printf "%s-ent" $tag -}}
  {{- else -}}
    {{- $tag -}}
  {{- end -}}
{{- else -}}
  {{- $tag -}}
{{- end -}}
{{- end -}}

{{- define "vault.imageRepository" -}}
{{- include "vault.resolveImageRepository" (dict "repo" .Values.server.image.repository "root" .) -}}
{{- end -}}

{{- define "vault.imageTag" -}}
{{- include "vault.resolveImageTag" (dict "tag" .Values.server.image.tag "root" .) -}}
{{- end -}}

File: test/acceptance/setup_suite.bash [hunk 17]
```diff
@@ -23,18 +29,27 @@ setup_suite() {
...
+    # For Enterprise the tags already carry the -ent suffix (appended above).
+    # derive hashicorp/vault-enterprise automatically from the license secret.
     CHART_VALUES+=(--set injector.agentImage.tag="${INJECTOR_AGENT_VERSION}")
-    CHART_VALUES+=(--set injector.agentImage.repository="${VAULT_REPOSITORY}")
     CHART_VALUES+=(--set server.image.tag="${SERVER_VAULT_VERSION}")
-    CHART_VALUES+=(--set server.image.repository="${VAULT_REPOSITORY}")
     CHART_VALUES+=(--set csi.agent.image.tag="${CSI_AGENT_VERSION}")
-    CHART_VALUES+=(--set csi.agent.image.repository="${VAULT_REPOSITORY}")
+
+    if [ "${ENT_TESTS}" != "true" ]; then
```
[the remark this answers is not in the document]
harshit-nema: Code is updated.

PR #1206, opened 2026-08-22
Add extraObjects for additional Kubernetes objects in the release

### LINKED ISSUE #1146: Feature: Add extraObjects Parameter

khateeb15:
## Summary

Adds an `extraObjects` value for deploying additional Kubernetes objects as part of the chart's release, so objects the chart does not model share the release's lifecycle instead of being applied separately.

Closes #1146. Builds on the approach proposed in #1143 by @giuliocalzo, who is credited as co-author. See "Relationship to #1143" below.

## Problem

Objects the chart does not model, such as a cert-manager `Certificate` issuing the Secret referenced by `server.volumes`, or a `NetworkPolicy` scoped to the chart's workloads, currently need a separate `kubectl apply` or a separate wrapper chart. That splits them from the release that depends on them: they are not tracked by Helm, not removed on `helm uninstall`, and can drift out of step with the resources the chart generates.

## Solution

A new top-level `extraObjects` value, rendered by `templates/extra-objects.yaml`.

**Entries may be a list or a map.** Helm replaces lists wholesale when merging values files but merges maps key by key, so the map form lets a layered values file (a base plus a per-environment overlay) add or override a single object without restating the others.

**Each entry may be a YAML map or a string.** The map form keeps values-file syntax highlighting and schema tooling working; the string form is convenient for manifests copied verbatim.

**Every entry is rendered with `tpl` against the chart's context,** so entries can call the chart's own helpers (`vault.fullname`, `vault.name`, `vault.namespace`) and stay in step with the names and labels the chart generates.

**An entry that renders to nothing is skipped** rather than emitting an empty document. This lets an entry be wrapped in a conditional and tied to a component being enabled, so a single values file can serve releases running any combination of components.

**An entry that is neither a map nor a string fails rendering** with an error naming its list index or map key, rather than producing invalid YAML further down.

`extraObjects` is deliberately not gated on `global.enabled` or on any component being enabled.

## Examples

As a list of maps:

```yaml
extraObjects:
  - apiVersion: cert-manager.io/v1
    kind: Certificate
    metadata:
      name: '{{ include "vault.fullname" . }}-tls'
      namespace: '{{ include "vault.namespace" . }}'
    spec:
      secretName: vault-tls
      issuerRef:
        name: internal-issuer
        kind: ClusterIssuer
      dnsNames:
        - vault.example.com
```

As a map of templated strings, where the second entry renders to nothing unless the CSI provider is enabled:

```yaml
extraObjects:
  serverCertificate: |
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: {{ include "vault.fullname" . }}-server-tls
      namespace: {{ include "vault.namespace" . }}
  csiCertificate: |
    {{- if .Values.csi.enabled }}
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: {{ include "vault.fullname" . }}-csi-tls
      namespace: {{ include "vault.namespace" . }}
    {{- end }}
```

## Relationship to #1143

#1143 proposed the same feature with a minimal four-line template. This PR keeps that idea and its `extraObjects` name and string-entry form, and adds map entries, the map-of-named-entries form, conditional skipping, type validation, a `values.schema.json` entry, unit tests, and a CHANGELOG entry.

I am happy for the maintainers to take whichever they prefer, or for this to be applied onto #1143's branch instead if @giuliocalzo would rather carry it there. I have commented on #1143 to say the same.

## Backwards compatibility

Additive only. The default is `null`, so nothing renders unless the value is set. `null` rather than `[]` so that neither the list nor the map form conflicts with the default's type when values files are merged.

## Testing

`test/unit/extra-objects.bats` adds 11 cases covering: not rendered by default; map entries; string entries; multiple entries in a list; the map-of-entries form; templating against the chart context; templating inside the map form; rendering when `global.enabled=false`; skipping an entry that renders to nothing; two component-scoped objects in one release; and the two failure modes, in both the list and map forms.

```
bats test/unit/extra-objects.bats
```

`helm lint` passes.

## PCI review checklist

  The change is additive and defaults to `null`. Reverting the pull request is sufficient.

  No change to the chart's own security controls. `extraObjects` renders operator-supplied manifests already under the operator's control, with the same privileges as any other manifest they apply to the cluster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+Features:
+
+* Add `extraObjects` for deploying additional Kubernetes objects as part of the chart's release. Accepts a list or a map, whose entries are either YAML maps or templated strings, and renders each entry with `tpl` against the chart's context [#1206](https://github.com/hashicorp/vault-helm/pull/1206)
+
 ## 0.34.1 (August 13, 2026)
 
 Changes:
--- a/templates/extra-objects.yaml
+++ b/templates/extra-objects.yaml
@@ -1,0 +1,38 @@
+{{/*
+Copyright IBM Corp. 2018, 2025
+SPDX-License-Identifier: MPL-2.0
+*/}}
+
+{{- /*
+extraObjects accepts either a list or a map of objects. Both are normalized to
+a list of entries that keeps the list index or the map key, so an entry of an
+unsupported type can be named in the error. Entries that render to nothing are
+skipped rather than emitting an empty document, so an entry may be wrapped in a
+conditional to tie it to a component being enabled.
+*/ -}}
+{{- $entries := list -}}
+{{- $extraObjects := .Values.extraObjects | default (list) -}}
+{{- if kindIs "map" $extraObjects -}}
+{{- range $key, $object := $extraObjects -}}
+{{- $entries = append $entries (dict "id" $key "object" $object) -}}
+{{- end -}}
+{{- else -}}
+{{- range $index, $object := $extraObjects -}}
+{{- $entries = append $entries (dict "id" $index "object" $object) -}}
+{{- end -}}
+{{- end -}}
+{{- range $entry := $entries }}
+{{- $object := $entry.object }}
+{{- $rendered := "" }}
+{{- if kindIs "string" $object }}
+{{- $rendered = tpl $object $ }}
+{{- else if kindIs "map" $object }}
+{{- $rendered = tpl (toYaml $object) $ }}
+{{- else }}
+{{- fail (printf "extraObjects[%v]: expected a map or a string, got %s" $entry.id (kindOf $object)) }}
+{{- end }}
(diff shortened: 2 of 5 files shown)


giuliocalzo:
Thanks @khateeb15. I have folded your additions into #1143 and credited you as co-author on the commit, so the two PRs now cover the same ground.

igor-nikiforov:
@vijayavelsekar could you please review this PR? Thanks!

PR #1211, opened 2026-09-16
Add global.extraLabels for chart-wide common labels

### LINKED ISSUE #911: server.extraLabels missing on statefulset
**Describe the bug**
Setting server.extraLabels only applies labels to pods and not to the statefulset itself.

**To Reproduce**
1. Install chart with server.extraLabels set
2. Look at statefulset

**Expected behavior**
Labels should be applied to the statefulset

**Environment**
* Kubernetes version: EKS 1.24
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.):
  * Other configuration options or runtime services (istio, etc.):
* vault-helm version: 0.24.1

Chart values:

```yaml
# Paste your user-supplied values here (`helm get values <release>`).
# Be sure to scrub any sensitive values!
server:
  extraLabels:
    tags.datadoghq.com/service: vault
```

**Additional context**
We're setting up Datadog and the statefulset should be labelled aswell.

### LINKED ISSUE #992: Ability to have top level label on StatefullSet
Hello ,

It would be great to be able to add custom label on statefullset here : 

https://github.com/hashicorp/vault-helm/blob/main/templates/server-statefulset.yaml#L16

This is implemented in consul chart but can't do this to vault  : 

https://github.com/hashicorp/vault-helm/blob/main/templates/server-statefulset.yaml#L16

Thanks

### LINKED ISSUE #1197: Support chart-wide commonLabels/extraLabels on all resources, and set app.kubernetes.io/version

Many organisations require a standard set of labels on *every* Kubernetes resource, for cost allocation, ownership/asset tracking, and policy enforcement (OPA/Kyverno rules that reject unlabelled resources). Today the chart makes this impossible without post-rendering or forking:

- `extraLabels` is only supported on a handful of resources: `server.extraLabels`, `injector.extraLabels`, `server.serviceAccount.extraLabels`, `csi.daemonSet.extraLabels`, `csi.pod.extraLabels`, and `csi.serviceAccount.extraLabels`.
- Even where supported, coverage is partial: `server.extraLabels` is applied only to the pod template, not to the StatefulSet's own `metadata.labels` (#911, #992).
- All other resources; Services, ConfigMaps, Ingress/Route/HTTPRoute, RBAC objects, the MutatingWebhookConfiguration, PodDisruptionBudgets, NetworkPolicies, Secrets, PSPs, ServiceMonitor/PrometheusRule, accept no custom labels at all.
- The chart never sets `app.kubernetes.io/version`, despite it being one of the [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/), and the standard labels it does set are applied inconsistently across templates (see #1183 for missing `helm.sh/chart` labels).

There is a long tail of open issues and PRs each attacking one slice of this problem (see "Additional context"), but no chart-wide solution.

1. A single chart-wide values entry e.g. `global.extraLabels` (or `commonLabels`) merged into `metadata.labels` of every resource the chart renders, including pod templates. This is a well-established pattern in other popular charts (kube-prometheus-stack `commonLabels`, ingress-nginx `commonLabels`, Bitnami's `commonLabels`, cert-manager `global.commonLabels`).
2. Set `app.kubernetes.io/version: {{ .Chart.AppVersion }}` as part of the standard labels on all resources.

To keep upgrades safe, the common labels should be added to `metadata.labels` and pod template labels only, not to `spec.selector`/`matchLabels`, which are immutable on existing StatefulSets/Deployments.

Existing per-resource `extraLabels` values would keep working as more specific overrides, so the change is fully backwards-compatible.

- **Post-rendering with Kustomize**: works, but pushes complexity onto every consumer of the chart and breaks the plain `helm install/upgrade` workflow.
- **Per-resource `extraLabels` additions**: this is what the existing open PRs do (#505, #680, #1176), but piecemeal coverage has left gaps for years and several of those PRs have gone stale.
- **Maintaining a fork**: our current workaround, but obviously undesirable for everyone.

**Additional context**

Related open issues:
- #911:`server.extraLabels` applies only to pods, not StatefulSet metadata
- #992: ability to set top-level labels on the StatefulSet (near-duplicate of #911)
- #265: standardisation of label naming conventions

Related open PRs:
- #505: adds `server.extraLabels` to StatefulSet metadata (open since ~2021)
- #680: labels for Deployment, StatefulSet, ServiceAccounts, volumeClaimTemplates
  (revival of #478, now stale with conflicts)
- #1176: `extraLabels` for Service resources only
- #1183: fixes missing `helm.sh/chart` labels on some resources

A chart-wide mechanism would resolve #911, #992, and supersede #505, #680, and #1176 in one consistent change rather than several partial ones.

I'm happy to contribute the PR for this (wip)

Siddharth-Kannan:
Adds a chart-wide global.extraLabels value that applies a map of custom labels to metadata.labels of every resource the chart renders, including the server StatefulSet, injector Deployment and CSI DaemonSet pod templates, so the labels land on the pods themselves.

Also completes the standard labels:

Sets the recommended app.kubernetes.io/version (from the chart appVersion, overridable via global.extraLabels) on all resources.
Adds helm.sh/chart to the 41 rendered resources that were missing it.
All label rendering now goes through a single vault.commonLabels helper, replacing the per-template copy-paste that let these labels drift apart in the first place.

Existing per-resource extraLabels keep working and take precedence over global.extraLabels, so the change is backwards compatible. As a side effect this also applies server.extraLabels and injector.extraLabels to the StatefulSet/Deployment metadata.labels, not just their pod templates.

Fixes #1197
Closes #911, closes #992
Supersedes #505, #680, #1176, #1183

Testing
New test/unit/global-extra-labels.bats — 20 tests covering label application across all 45 templates, pod templates, selector and volumeClaimTemplates exclusion, precedence, the failure modes, and label-value validity. Includes a test asserting the suite itself renders every template, so a newly added template cannot silently skip these assertions.
5 existing assertions updated for the added app.kubernetes.io/version label.

## PCI review checklist


templates/_helpers.tpl
```diff
@@ -43,6 +43,72 @@ Allow the release namespace to be overridden
...
+  {{- range $key, $value := $global -}}
+    {{- if has $key $managed -}}
+      {{- fail (printf "global.extraLabels: %q is managed by the chart and cannot be overridden" $key) -}}
+    {{- end -}}
+  {{- end -}}
+  {{- $labels := dict -}}
+  {{- if not .omitChart -}}
+    {{- $_ := set $labels "helm.sh/chart" (include "vault.chart" $ctx) -}}
+  {{- end -}}
+  {{- if $ctx.Chart.AppVersion -}}
```
harshit-nema: app.kubernetes.io/version uses Chart.AppVersion (2.0.4) for every resource, but the injector (vault-k8s:1.7.6) and CSI provider (vault-csi-provider:1.7.4) run different images — verified by rendering both templates.

Suggested fix:
```
{{- $version := .version | default $ctx.Chart.AppVersion -}}
  {{- if $version -}}
    {{- $_ := set $labels "app.kubernetes.io/version" ($version | toString | replace "+" "_" | trunc 63 | trimSuffix "-") -}}
  {{- end -}}
```

Then pass "version" .Values.injector.image.tag / "version" .Values.csi.image.tag / "version" .Values.server.image.tag at their respective call sites, leaving all other call sites to fall back to Chart.AppVersion unchanged.

templates/injector-deployment.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     component: webhook
+    {{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.injector.extraLabels) | nindent 4 }}
```
harshit-nema: Refer comment in _helpers.tpl line90-91
Should we update code to : 
{{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.injector.extraLabels "version" .Values.injector.image.tag) | nindent 4 }}

templates/injector-deployment.yaml
```diff
@@ -30,9 +31,7 @@ spec:
         app.kubernetes.io/name: {{ template "vault.name" . }}-agent-injector
         app.kubernetes.io/instance: {{ .Release.Name }}
         component: webhook
-        {{- if  .Values.injector.extraLabels -}}
-          {{- toYaml .Values.injector.extraLabels | nindent 8 -}}
-        {{- end -}}
+        {{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.injector.extraLabels "omitChart" true) | nindent 8 -}}
```
harshit-nema: Refer comment in _helpers.tpl line90-91
Should we update code to :
`{{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.injector.extraLabels "omitChart" true "version" .Values.injector.image.tag) | nindent 8 -}}`

templates/csi-daemonset.yaml
```diff
@@ -14,9 +14,7 @@ metadata:
     app.kubernetes.io/name: {{ include "vault.name" . }}-csi-provider
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
-    {{- if  .Values.csi.daemonSet.extraLabels -}}
-      {{- toYaml .Values.csi.daemonSet.extraLabels | nindent 4 -}}
-    {{- end -}}
+    {{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.csi.daemonSet.extraLabels) | nindent 4 -}}
```
harshit-nema: Refer to comment in helpers.tpl 90-91
Should we update code to:
`{{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.csi.daemonSet.extraLabels "version" .Values.csi.image.tag) | nindent 4 -}}`

templates/_helpers.tpl
harshit-nema: app.kubernetes.io/version uses Chart.AppVersion (2.0.4) for every resource, but the injector (vault-k8s:1.7.6) and CSI provider (vault-csi-provider:1.7.4) run different images — verified by rendering both templates.

Suggested fix:

{{- $version := .version | default $ctx.Chart.AppVersion -}}
  {{- if $version -}}
    {{- $_ := set $labels "app.kubernetes.io/version" ($version | toString | replace "+" "_" | trunc 63 | trimSuffix "-") -}}
  {{- end -}}
Then pass "version" .Values.injector.image.tag / "version" .Values.csi.image.tag / "version" .Values.server.image.tag at their respective call sites, leaving all other call sites to fall back to Chart.AppVersion unchanged.

harshit-nema: If the above fix is applied, should we add test at : test/unit/global-extra-labels.bats for injector and csi ?

templates/csi-daemonset.yaml
harshit-nema: Refer to comment in helpers.tpl 90-91
Should we update code to:
{{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.csi.daemonSet.extraLabels "version" .Values.csi.image.tag) | nindent 4 -}}

templates/csi-daemonset.yaml
```diff
@@ -34,9 +32,7 @@ spec:
       labels:
         app.kubernetes.io/name: {{ template "vault.name" . }}-csi-provider
         app.kubernetes.io/instance: {{ .Release.Name }}
-        {{- if  .Values.csi.pod.extraLabels -}}
-          {{- toYaml .Values.csi.pod.extraLabels | nindent 8 -}}
-        {{- end -}}
+        {{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.csi.pod.extraLabels "omitChart" true) | nindent 8 -}}
```
harshit-nema: Refer to comment in _helpers.tpl line-90-91
Should we update the code to :
{{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.csi.pod.extraLabels "omitChart" true "version" .Values.csi.image.tag) | nindent 8 -}}

templates/injector-deployment.yaml
harshit-nema: Refer comment in _helpers.tpl line90-91
Should we update code to : 
{{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.injector.extraLabels "version" .Values.injector.image.tag) | nindent 4 }}

templates/injector-deployment.yaml
harshit-nema: Refer comment in _helpers.tpl line90-91
Should we update code to :
{{- include "vault.commonLabels" (dict "ctx" . "extra" .Values.injector.extraLabels "omitChart" true "version" .Values.injector.image.tag) | nindent 8 -}}

